Issue #836, opened 2026-04-30
Dependency resolution audit trail in RegistryEvent


dhyansraj:
## Summary

Record the candidate-filtering decision tree when the registry resolves a dependency, so operators can answer "why did this consumer wire to producer A instead of B?" Reuses the existing `RegistryEvent` table with new event types — no new table, no new query layer.

## Motivation

Today the dependency resolver (`src/core/registry/resolver.go`, `src/core/registry/matcher.go`) is a black box. A consumer wires to one producer over another and there's no record of why. This becomes painful as:

- Mesh size grows — multiple providers per capability is common
- Auto-rewiring kicks in silently when candidates appear/disappear
- The schema-registry feature (#547) adds schema-mode as another filter dimension, multiplying the reasons a candidate might be evicted

An audit trail surfaces:
- Initial wiring decisions (which candidates entered the pipeline, which were rejected at each stage, who won)
- Re-wiring events (and what changed)
- Why specific candidates were rejected — typed reasons, not free-form strings

### Reuse `RegistryEvent` table

The existing `RegistryEvent` entity (`src/core/ent/schema/registryevent.go:12-52`) is the right home:

- Flexible JSON `data` payload — can carry the decision trace without schema changes
- Indexed on `timestamp` + `event_type` — fast `event_type=dependency_resolved AND agent_id=X` queries
- Existing `agent_id` edge + `function_name` column — natural per-consumer / per-dep-slot keying
- `ListRecentEvents()` (`ent_service.go:2121`) and `EventsHistoryResponse` API type already exist — query infrastructure is free

Add new `event_type` enum values via Ent:
- `dependency_resolved` — initial wire, or re-wire to a different producer
- `dependency_unresolved` — no candidate matched; consumer is waiting

### When to emit

**Not on every heartbeat re-resolution** (would flood). Emit only on:

1. Initial wire (consumer's dep first resolved)
2. Re-wire (chosen producer changed since last resolution)
3. (Future) operator-requested replay

Emission is also gated on **interesting decisions**: only record when ≥2 candidates entered the pipeline at any stage, OR when the choice changed. Single-candidate forced choices add no value to the audit and would just be noise.

### Payload schema (in JSON `data` blob)

```json
{
  "consumer": "trip-planner-a3f9",
  "dep_index": 0,
  "spec": {
    "capability": "employee",
    "tags": ["api"],
    "version_constraint": ">=2.0",
    "schema_mode": "compatible"
  },
  "stages": [
[...]

PR #837, opened 2026-05-01
Registry sweep job: purge stale agents and old events

### LINKED ISSUE #835: Registry sweep job: purge stale agents and old events
## Summary

Add a periodic sweep job in the Go registry to purge:
1. Agents that have been unhealthy/expired beyond a configurable retention window (default 1h)
2. Old `RegistryEvent` rows beyond a configurable retention window (default 7d)

Single goroutine, configurable via env vars, runs on a timer + once at registry startup. Works on both SQLite (dev) and Postgres (prod) — both rely on plain `DELETE WHERE timestamp < ?`; neither has native TTL but neither needs it.

## Motivation

**Agent noise in `meshctl list` / `status`.** Users have reported that the count of expired/dead agents grows over time and becomes noise. Today the registry never deletes agent rows after heartbeat failure — they accumulate forever. A simple purge after 1h of unhealthy state removes the noise while still leaving a useful "recently dead" window for debugging.

**Event table unbounded growth.** `RegistryEvent` (`src/core/ent/schema/registryevent.go:12-52`) has no TTL, no sweep, no pruning today. This is already a latent issue. It becomes urgent with the dependency-resolution audit feature (filed separately) which pushes event volume up considerably.

A single sweep job handles both. Configurable, safe defaults, no new operational dependencies.

### Sweep job

- One goroutine launched at registry startup
- Runs on a timer (`MCP_MESH_SWEEP_INTERVAL`, default `5m`)
- Also runs once at startup (catch missed sweeps after downtime)
- On each tick, performs (in order):
  1. Purge stale agents (unhealthy/expired older than agent retention)
  2. Purge old events (timestamp older than event retention)
  3. Apply hard-cap fallback on event table if needed
- Logs what was purged: `sweep: purged 3 agents, 142 events (took 23ms)`

### Configuration (env vars)

| Variable | Default | Description |
|---|---|---|
| `MCP_MESH_SWEEP_INTERVAL` | `5m` | How often the sweep job runs |
| `MCP_MESH_AGENT_RETENTION` | `1h` | Purge unhealthy/expired agents older than this (since last_heartbeat) |
| `MCP_MESH_EVENT_RETENTION` | `7d` | Purge events older than this (since timestamp) |
| `MCP_MESH_EVENT_MAX_ROWS` | `100000` | Hard cap on event table size; oldest beyond this purged regardless of age |

All durations parse via `time.ParseDuration` (so `30m`, `2h`, `48h` etc. work). Set any retention to `0` to disable that purge entirely (useful for tests or operators who want forever-retention).

### Database compatibility

Both SQLite and Postgres handle plain `DELETE WHERE timestamp < ?` fine:
- **Postgres**: autovacuum reclaims space automatically; nothing extra needed.
- **SQLite**: may need an occasional `VACUUM` if delete volume is heavy. Defer until empirically observed; not v1.

Neither DB has native TTL out of the box. Postgres extensions (`pg_partman`, `pg_cron`, TimescaleDB) exist but introduce operational dependencies — out of scope.

### Agent purge specifics

Purge agents where `status IN (unhealthy, expired)` AND `(now - last_heartbeat) > MCP_MESH_AGENT_RETENTION`.

- Agents in `healthy` status are never purged.
- Agents explicitly unregistered (via `unregister` event) are already cleaned up at unregister time — no overlap.
- Cascade-delete the agent's capabilities (FK).
- Existing `dependency_resolution` rows referring to a purged agent become orphan refs — implementer decides between cascade-delete or leaving them as historical (slight preference for cascade since the agent no longer exists).
- After purge: `meshctl list` and `meshctl status` will no longer show the agent. The "expired count" in summary output should be updated to reflect this — likely just remove that count entirely or replace with "recent failures (last 1h)".

### Event purge specifics

Time-based: `DELETE FROM registry_event WHERE timestamp < (now - MCP_MESH_EVENT_RETENTION)`.

Hard cap fallback: if the table still exceeds `MCP_MESH_EVENT_MAX_ROWS` after time-based purge, delete oldest rows until under the cap. Safety net for runaway loops or misconfig.

One global retention window for v1 — not per-event-type. If we ever need per-type windows (e.g., keep `unregister` forever, keep `heartbeat` for 1d), add later.

## Implementation Scope

- [ ] New file `src/core/registry/sweep.go` with `SweepJob` struct + `Start(ctx)` / `Stop()` methods
- [ ] Wire startup in `cmd/mcp-mesh-registry/main.go` (or wherever registry boots)
- [ ] Two Ent `DELETE` queries (agents + events)
- [ ] Env var parsing with defaults; `time.ParseDuration` for durations
- [ ] Structured logs of purge counts and duration
- [ ] Unit tests with a fast tick interval verifying both purges fire
- [ ] Integration test: register an agent, transition to unhealthy, advance time, verify purged from `meshctl list`
- [ ] Update `meshctl list`/`status` summary to remove or rename the "expired count"

## Out of scope (future)

- Per-event-type retention windows
- Postgres partitioning for the events table (revisit if volume grows past ~1M rows)
- TimescaleDB / pg_cron / pg_partman (no new operational deps)
- Showing "last sweep time" in `meshctl status`
- `meshctl sweep --now` operator-triggered manual sweep

## Related

- Precondition for the dependency-resolution audit feature (filed separately) — that feature pushes event volume up and shouldn't ship without retention in place.

dhyansraj:
## Summary

- Adds a periodic sweep goroutine in the Go registry that purges agents in `unhealthy`/`unknown` status whose `updated_at` is older than `MCP_MESH_RETENTION` (default `1h`)
- Single env var knob: `MCP_MESH_RETENTION` (set to `0` to disable the sweep entirely as a forensic escape hatch). Sweep interval (5m) and event hard cap (100k rows) are internal constants — operators get one knob, not four
- Folds in a latent orphan-fix bug: when sweep purges a provider agent, all three resolution tables (`dependency_resolution`, `llm_tool_resolution`, `llm_provider_resolution`) get flipped from `status=available` to `status=unavailable` BEFORE the cascade-set-null FK fires. Without this, consumers would see stale "available" rows pointing at NULL provider FK
- Sweep runs at startup (catches missed sweeps after registry downtime), then every 5 minutes
- Documentation: `MCP_MESH_RETENTION` added to `docs/environment-variables.md`
- meshctl `list` summary text updated: `unhealthy/expired` → `unhealthy` (sweep removes the long tail of expired rows that prompted user feedback about noise)

## Review Notes

Two-round code review caught a BLOCKER and four WARNINGs in the first iteration; all addressed in this PR:

- **BLOCKER (fixed)**: Original orphan-fix only flipped `dependency_resolution` — missed the two sibling tables (`llm_tool_resolution`, `llm_provider_resolution`) with identical schema. `UpdateDependencyStatusOnAgentOffline()` extended to all three.
- **Race-safety (fixed)**: Inside the transaction, the agent is re-queried with the same staleness predicate (`status IN (unhealthy, unknown) AND updated_at < threshold`). If a heartbeat revives the agent between candidate scan and tx open, the re-query returns 0 rows and the tx commits as a no-op. New `TestSweepSkipsAgentRevivedDuringTick` exercises this path.
- **Atomicity (fixed)**: Orphan-fix + delete now run inside a single Ent transaction (`s.entDB.Transaction(ctx, ...)`).
- **Test coverage (fixed)**: `eventMaxRows` is now injectable on `SweepJob` (parallel to the `clock` field). New `TestSweepEventMaxRowsOverCap` sets cap=3, seeds 5 strictly-increasing-timestamp events, and asserts the 2 oldest are deleted in the right order.
- **Boot wiring (simplified)**: Removed redundant `sweepCtx`/`sweepCancel` fields on `Server` — `SweepJob.Stop()` already self-cancels and waits.

### Cascade limitation (accepted)

The existing `Agent.events` edge has `OnDelete: Cascade`, which means purging an agent also deletes its `RegistryEvent` history. For v1 we accept this — the dependency-resolution audit feature (#836) will need to detangle this if forensic event history matters. Captured here so reviewers know it's intentional.

### Verified locally

End-to-end smoke-tested against real agents from `examples/simple/`:
- Healthy agents stay registered; ungraceful kill → marked unhealthy by health monitor → purged on next sweep tick after retention
- Provider purge correctly flips consumer-side `dependency_resolution.status` to `unavailable` before the agent row is deleted
- Sweep handles backlog from previous sessions (33 stale agents purged on first tick after a long downtime)
- `MCP_MESH_RETENTION=0` correctly skips launching the goroutine

Closes #835

## Test plan

- [x] All Go unit tests pass (`go test ./src/core/registry/...`)
- [x] `go build ./...` clean
- [x] Smoke tested with `MCP_MESH_RETENTION=2m` against `examples/simple/` agents
- [x] Verified DB rows actually deleted (sqlite3 query), not just hidden by CLI
- [x] Verified orphan-fix flips consumer-side `dependency_resolutions` row before delete
- [x] Verified `MCP_MESH_RETENTION=0` produces the disabled startup log line and no sweep goroutine
- [x] Verified existing caller of broadened `UpdateDependencyStatusOnAgentOffline` (`TestTopologyChange_AgentOffline`) still passes
- [ ] CI green

## Release Notes

* **New Features**
  * Added automatic registry sweep job with configurable retention period via `MCP_MESH_RETENTION` environment variable (default: 1 hour; disable with "0")
  * Sweep removes stale/unhealthy agents and maintains event log size limits
  * Enhanced `meshctl list --all` to include unhealthy agents with automatic purging

* **Bug Fixes**
  * Registry now properly marks related resolution records as unavailable when agents go offline

* **Documentation**
  * Added `MCP_MESH_RETENTION` environment variable documentation with timing and scope details

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -188,6 +188,29 @@ export HEALTH_CHECK_INTERVAL=10
 export DEFAULT_EVICTION_THRESHOLD=60
 ```
 
+### Registry Sweep / Retention
+
+```bash
+# How long unhealthy/unknown agents are kept in the registry before the
+# sweep job purges them. Go duration string (e.g. "30m", "2h", "48h").
+# Default: 1h. Set to "0" to disable the sweep entirely (forensic mode —
+# keeps all rows). Affects both `meshctl list` (purged agents disappear
+# from output) and the underlying RegistryEvent table (event rows are
+# governed by a separate hardcoded 100,000 rolling cap, not by this
+# variable).
+export MCP_MESH_RETENTION=1h
+```
+
+**Notes:**
+
+- The sweep runs on a hardcoded 5-minute interval, so actual purge can
+  lag retention by up to ~5 minutes.
+- Setting `MCP_MESH_RETENTION` shorter than 5m (e.g. `1m`) will not
+  speed up the purge cadence — it only affects when an agent becomes
+  eligible for purge.
+- Hardcoded internal constants: sweep interval = 5m, event hard cap =
+  100,000 rows.
+
 ### Cache and Performance
 
 ```bash
--- a/src/core/cli/list.go
+++ b/src/core/cli/list.go
@@ -42,7 +42,7 @@ Shows agent status, dependency resolution, uptime, and endpoint information
 in a docker-compose-style format for easy monitoring.
 
 By default, only healthy agents are shown. Use --all to see all agents
-including unhealthy/expired ones.
+including unhealthy ones (purged after MCP_MESH_RETENTION, default 1h).
 
 Examples:
   meshctl list                                    # Show healthy agents only (default)
@@ -72,7 +72,7 @@ Tools listing:
 
 	// New enhanced filtering and display options
 	cmd.Flags().String("since", "", "Show agents active since duration (e.g., 1h, 30m, 24h)")
-	cmd.Flags().Bool("all", false, "Show all agents including unhealthy/expired (default: healthy only)")
+	cmd.Flags().Bool("all", false, "Show all agents including unhealthy (default: healthy only)")
 
 	// Tools listing - use --tools to list all, --tools=<name> for specific tool details
 	cmd.Flags().StringP("tools", "t", "", "List tools: use --tools or -t to list all, --tools=<tool> for details")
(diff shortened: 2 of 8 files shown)


dhyansraj:
## Review findings — verified against the code, two fixed and two skipped

I went through each finding and read the actual code paths instead of trusting the descriptions blindly. Outcomes:

### Fixed in 1e87a658

**1. `LoadSweepConfigFromEnv` accepts negative durations** — REAL.
`time.ParseDuration("-1h")` succeeds with `d=-1h`, gets assigned to `cfg.Retention`, and the `Retention <= 0` check in `Start()` then silently treats it as disabled. A user typing `-1h` (typo for `1h`) gets sweep disabled with zero feedback. Now warns and falls back to default; `0` remains the documented disable mechanism.

**2. `markAgentStaleAttempt` overwrites `UpdatedAt`** — REAL edge case.
`SetStatus(unhealthy).SetUpdatedAt(now)` was bumping `UpdatedAt` to now, defeating the immediate sweep tick (sweep filters `UpdatedAtLT(now-retention)`). After registry restart, agents stale for hours/days survived the immediate sweep and waited another full retention period. Root cause: `UpdatedAt` was being conflated as both "last heartbeat" and "row modification time" — sweep wants the former. Now we only flip `Status`; the original `UpdatedAt` is preserved.

### Verified NOT real, deliberately skipped

**3. Time-based event purge missing** — deliberate design choice.
This was the original implementation; it was explicitly removed when the user said "we are overcomplicating with too many variables." Events are governed solely by the 100k rolling cap. Re-adding time-based event retention would re-introduce the env var we just collapsed. The PR body and #835 issue body both document this as intentional. Re-opening this changes the v1 design contract, not a bug.

**4. Sweep keeps running on startup error** — not real in practice.
`s.engine.Run(addr)` and `s.runWithTLS(addr)` are blocking. They return only on shutdown or startup error. On startup error, `Run()` returns to `main()`, process exits, all goroutines die. Also: `healthMonitor` (line 127) and the admin server goroutine (line 422) have the **identical pattern** with no cleanup-on-error. Fixing only sweep would be inconsistent — defensive cleanup of helper goroutines is a cross-cutting concern, not this PR's scope. Worth a separate issue if desired.

Issue #834, opened 2026-04-29
Python LLM stack: hybrid native vendor SDKs (big 3) + LiteLLM fallback (long tail)


dhyansraj:
## Summary

Move Python's LLM provider layer from LiteLLM-only to a hybrid model: native vendor SDKs (`anthropic`, `openai`, `google-genai` / `google-cloud-aiplatform`) for Anthropic / OpenAI / Gemini (the "big 3" — ~95%+ of usage), keep LiteLLM as the long-tail fallback for everything else (Mistral, Cohere, Bedrock-non-Claude, Groq, Ollama, …). Aligns Python with TypeScript (Vercel AI SDK + per-vendor adapters) and Java (Spring AI per-vendor starters), where this pattern already lives.

Backend variants (Azure OpenAI, Bedrock-Claude, Databricks Anthropic-passthrough, Vertex AI) are handled inside the same per-vendor handler via prefix-based dispatch — the same pattern we already use for `gemini/...` (AI Studio) vs `vertex_ai/...` (GCP IAM).

### Why we're rethinking
- We already maintain per-vendor handlers (`claude_handler.py`, `openai_handler.py`, `gemini_handler.py`) — strong signal that LiteLLM's lowest-common-denominator API is leaky. Each handler has vendor-specific code paths (HINT-first for Claude, response_format quirks for Gemini, parallel_tool_calls for OpenAI).
- TypeScript and Java already use vendor-native adapters via Vercel AI SDK and Spring AI. Python is the only outlier going through a translation layer.
- Vendor features lag in LiteLLM. Concrete examples we can't surface today:
  - **Anthropic**: granular cache usage (`cache_creation_input_tokens` / `cache_read_input_tokens`), Claude 3.7's `thinking` reasoning blocks, citations API, computer-use tool variants
  - **OpenAI**: full structured outputs (`response_format` with `strict: true`), o1/o3-class reasoning tokens (`reasoning.effort`), predicted outputs
  - **Gemini**: Google Search grounding, code execution as a built-in tool, native `responseSchema` (we currently work around limitations)
  - **Streaming nuances**: vendor-specific event types (Anthropic's `message_delta` with chunked usage, OpenAI's tool-call argument streaming) get normalized to a lossy common shape
  - **Errors**: rate-limit headers, retry-after, organization quota info — all flattened by LiteLLM

### Why hybrid (not full replacement)
[...]

Issue #866, opened 2026-05-05
Cross-loop httpx pool binding — apply per-loop fix to anthropic_native + openai_native


dhyansraj:
## Summary

The shared process-global `httpx.AsyncClient` pattern in `anthropic_native.py` and `openai_native.py` is theoretically vulnerable to the same cross-loop binding bug just fixed in `gemini_native.py` (PR #865 follow-up). Currently doesn't surface in their integration tests because the test mix happens not to exercise enough cross-worker-loop churn — but the architectural risk is real.

## Root cause (from PR #865 investigation)

`httpx.AsyncClient` (via httpcore + anyio) lazily constructs sync primitives (Lock, Semaphore, Event for the connection pool) on first I/O. Those primitives bind to the asyncio event loop that issues the first request.

mesh's `shared/tool_executor.py` runs an N-worker thread pool — each worker thread owns its own long-lived asyncio loop — and round-robins tool calls across workers. A process-global `httpx.AsyncClient` binds to the first worker's loop and breaks the moment a call lands on a different worker's loop:

```
RuntimeError: <asyncio.locks.Event object at 0x...> is bound to a different event loop
```

## How Gemini exposed it

`uc10_toolcalls` Gemini tests chain 10+ tool calls per test, repeatedly. That cross-worker churn tripped the boundary every run after PR #865 merged. All 9 uc10 tests with `gemini_provider_py` failed.

## Fix already applied to Gemini (PR #865 follow-up commit)

```python
_CACHED_HTTPX_CLIENT_BY_LOOP: dict[int, httpx.AsyncClient] = {}

def _get_shared_httpx_client() -> httpx.AsyncClient:
    try:
        loop_id = id(asyncio.get_running_loop())
    except RuntimeError:
        # No running loop (sync path) — return uncached client
        return httpx.AsyncClient(...)
    
    with _LOCK:
        client = _CACHED_HTTPX_CLIENT_BY_LOOP.get(loop_id)
        if client is None or client.is_closed:
            client = httpx.AsyncClient(...)
            _CACHED_HTTPX_CLIENT_BY_LOOP[loop_id] = client
        return client
```

## Why anthropic/openai haven't surfaced this yet

Their current integration test mix doesn't exercise the same cross-worker churn pattern as uc10_toolcalls Gemini does. But the architectural vulnerability is identical — same shared-pool-on-first-loop pattern.

## Scope

[...]

Issue #860, opened 2026-05-04
Refactor: move LLM content-block translation upstream (vendor-native shape from emitter)


dhyansraj:
## Summary

Move multimedia / content-block translation from the per-vendor adapter (current state) **upstream** into the mesh content emitters, so each emitter produces the vendor-native shape directly. The adapter then becomes a passthrough — only doing transport, not transformation.

Cleaner architectural separation; same end behavior.

### Current state (post #834 PR 1)

Mesh internally emits OpenAI-shape content blocks regardless of which vendor will receive them:

- `mesh/helpers.py:437-439` — `resolve_resource_links(result, "openai")` (hardcoded)
- `_mcp_mesh/engine/mesh_llm_agent.py:672-674` — `_resolve_media_inputs` always uses `_format_for_openai` regardless of vendor
- Both emit `{"type": "image_url", "image_url": {"url": "data:<mime>;base64,<data>"}}`

The native vendor adapters then translate at the wire boundary:

- `anthropic_native.py:_translate_content_block_to_anthropic` (added in PR for #834) — converts OpenAI shape → Anthropic `{"type": "image", "source": {...}}`
- LiteLLM (used for non-Anthropic vendors today) does the same translation internally

This is the **defensive, lower-risk fix** that unblocked the multimedia regression in PR 1 of #834. It works, but it inverts the natural responsibility:
- The emitter knows which vendor will receive the content (already passes `vendor` through the agentic loop)
- But emits in a generic shape, forcing every vendor adapter to translate

### Already-built infrastructure that we don't use

`mesh/media/resolver.py` already has the right pattern, partially implemented:

\`\`\`python
_VENDOR_FORMATTERS = {
    \"openai\": _format_for_openai,
    \"anthropic\": _format_for_claude,
    \"claude\": _format_for_claude,
    # gemini missing
}
\`\`\`

`_format_for_claude` returns Anthropic-native shape directly. It exists. Nothing calls it.

### Cleaner target state

\`\`\`
@mesh.llm consumer
   ↓ ([typed mesh content], vendor=\"anthropic\")
mesh emitters use _VENDOR_FORMATTERS[vendor] → produce native shape
   ↓ ([{type: \"image\", source: {...base64...}}, ...])
native Anthropic adapter (passthrough)
   ↓
api.anthropic.com
\`\`\`

Same flow but the adapter just packages and ships — no shape transformation. Mirrors how Spring AI and Vercel AI SDK structure their internal flow.

[...]

PR #867, opened 2026-05-05
fix: output_type kwarg collision in MeshLlmAgent (#863)

### LINKED ISSUE #863: Bug: output_type kwarg collision in MeshLlmAgent._build_request_params
## Summary

`MeshLlmAgent._build_request_params` at `src/runtime/python/_mcp_mesh/engine/mesh_llm_agent.py:821` fails with `TypeError: prepare_request() got multiple values for keyword argument 'output_type'` when the LLM agent is configured with a Pydantic `output_type` AND certain decorator/call kwargs include it.

Affects structured-output workflows on both LiteLLM and native dispatch paths — the error fires BEFORE any dispatch decision.

## Discovery

Surfaced during E2E verification of #834 PR 2 (native OpenAI dispatch). Test-agent built a simple consumer:

```python
class CityFact(BaseModel):
    city: str
    fact: str

@mesh.llm(provider={"capability": "llm"}, output_type=CityFact)
def extract(city: str, llm = None) -> CityFact:
    return llm(f"Give one fact about {city}")
```

Calling `extract` produces:

```
File ".../mesh_llm_agent.py:821 in _build_request_params
    return self._provider_handler.prepare_request(
        messages=messages,
        tools=effective_tools if effective_tools else None,
        output_type=self.output_type,
        **call_kwargs,
    )
TypeError: prepare_request() got multiple values for keyword argument 'output_type'
```

## Root cause

`_build_request_params` passes `output_type=self.output_type` AS AN EXPLICIT KWARG, then splats `**call_kwargs`. If `call_kwargs` (built from `_default_model_params` which captures decorator kwargs) ALSO contains `output_type`, you get the collision.

Looking at `mesh_llm_agent.py:821`:
```python
return self._provider_handler.prepare_request(
    messages=messages,
    tools=effective_tools if effective_tools else None,
    output_type=self.output_type,
    **call_kwargs,        # ← may contain output_type
)
```

## Why PR 1 (#862) didn't catch this

PR 1's E2E used the same `output_type=Pydantic` pattern with Anthropic provider and worked. Possible explanations:
1. ClaudeHandler.prepare_request signature absorbs `output_type` differently (need to verify)
2. The `_default_model_params` capture path differs based on decorator pattern (need to verify)
3. The OpenAI E2E used a slightly different consumer setup that exposed the bug

Need to investigate which is the actual trigger to fix correctly.

## Suggested fix

Pop `output_type` from `call_kwargs` before splatting:

```python
call_kwargs = {**effective_kwargs, "output_mode": self.output_mode} if ...
call_kwargs.pop("output_type", None)  # avoid kwarg collision

return self._provider_handler.prepare_request(
    messages=messages,
    tools=...,
    output_type=self.output_type,
    **call_kwargs,
)
```

Or thread `output_type` through `call_kwargs` only and remove the explicit kwarg.

## Reproduction

Use the test-agent's consumer setup at `/tmp/openai-wire-test/openai-consumer/main.py`:
```python
@mesh.llm(provider={"capability": "llm"}, output_type=CityFact)
def extract(city: str, llm = None) -> CityFact:
    return llm(f"Give one fact about {city}")
```

With ANY provider (openai/anthropic) running, call `meshctl call openai-wire-consumer:extract '{"city": "Tokyo"}'`.

## Scope

- Affects: `@mesh.llm(output_type=PydanticModel)` consumer agents
- Both native and LiteLLM dispatch paths fail before dispatch
- Direct mode (`@mesh.llm(model=...)`) likely also affected (same `_build_request_params`)
- Mesh-delegated mode (`@mesh.llm(provider=...)`) confirmed affected
- Severity: structured-output workflows on consumer side blocked

## Out of scope

- Native OpenAI dispatch correctness (verified working for text path in PR 2)
- Direct path retirement (#859)

## Related

- Discovered while validating #862 (PR 1) and PR 2 of #834
- Provider-side structured output (handler-side) works fine — this is consumer-side

dhyansraj:
## Summary

Pop `output_type` from `call_kwargs` before splatting into `prepare_request`, preventing `TypeError: got multiple values for keyword argument 'output_type'` when `output_type` leaks into `_default_model_params` from decorator capture.

Discovered during PR 2 of #834 (#864) E2E testing.

## Root cause

1. `mesh/decorators.py:1840` — `resolved_config.update(kwargs)` merges all caller kwargs (including `output_type=...`) into `resolved_config`
2. `_mcp_mesh/engine/mesh_llm_agent_injector.py:752-769` — builds `default_model_params` filtering by `INTERNAL_CONFIG_KEYS`, which doesn't include `"output_type"` — so it passes through to the agent
3. `mesh_llm_agent.py:_build_request_params` (line ~821) — does `prepare_request(..., output_type=self.output_type, **call_kwargs)`. If `call_kwargs` ALSO contains `output_type` from `_default_model_params`, Python raises `TypeError`.

## Fix

Single-line defensive pop at the consumer site:

```python
# Avoid kwarg collision: prepare_request takes output_type as explicit
# kwarg below; pop from call_kwargs to prevent "got multiple values"
# TypeError if it leaked into _default_model_params (e.g., from decorator
# capture). See #863.
call_kwargs.pop("output_type", None)
```

`self.output_type` (set at agent construction) is authoritative — wins over any leaked-through kwarg.

## Other call sites checked + safe

- `mesh_llm_agent.py:732` `format_system_prompt` — explicit kwarg, no spread
- Helper-routed call sites at `:918`, `:1340`, `:1565` — all go through `_build_request_params`, protected by the single fix
- Mesh-delegate streaming branch `:1373` — builds `model_params` via filter, no spread collision

## Tests

2 regression tests in `test_mesh_llm_agent_proxy.py`:
1. `test_build_request_params_no_kwarg_collision_when_output_type_in_default_params` — direct repro by planting `output_type` in `_default_model_params`; asserts no TypeError + `output_type` arrives exactly once with agent's value
2. `test_build_request_params_call_time_output_type_kwarg_also_dropped` — defense-in-depth: caller-supplied `output_type=...` at call time is also dropped; `self.output_type` wins

Both would fail without the one-line fix (verified independently — the bug reproduces).

## Test results

- 1093 unit tests pass (1091 baseline + 2 new)
- Zero regressions

## Deeper fix (deferred)

The cleanest long-term fix is to add `"output_type"` to `INTERNAL_CONFIG_KEYS` in `mesh_llm_agent_injector.py:752` so `_default_model_params` doesn't carry the redundant value at all. Not done here per minimal-fix scope; defensive pop covers all flows uniformly. Could be a small follow-up if drift cleanliness matters.

Closes #863

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a TypeError that occurred when output type parameters were inadvertently duplicated during request preparation.

* **Tests**
  * Added regression tests to ensure output type parameter handling remains stable and prevents future collisions.

### DIFF
--- a/src/runtime/python/_mcp_mesh/engine/mesh_llm_agent.py
+++ b/src/runtime/python/_mcp_mesh/engine/mesh_llm_agent.py
@@ -813,6 +813,12 @@ class MeshLlmAgent:
             if self.output_mode
             else effective_kwargs
         )
+        # Avoid kwarg collision: prepare_request takes output_type as an
+        # explicit kwarg below; pop it from call_kwargs to prevent
+        # "got multiple values for keyword argument 'output_type'" TypeError
+        # if it leaked into _default_model_params (e.g., via decorator
+        # **kwargs capture in @mesh.llm). See #863.
+        call_kwargs.pop("output_type", None)
         effective_tools = (
             self._enrich_tools_with_endpoints()
             if self._is_mesh_delegated and self._tool_schemas
--- a/src/runtime/python/tests/unit/test_mesh_llm_agent_proxy.py
+++ b/src/runtime/python/tests/unit/test_mesh_llm_agent_proxy.py
@@ -2344,3 +2344,103 @@ class TestMeshDelegationMeta:
 
         assert usage is None
         assert model is None
+
+
+class TestBuildRequestParamsKwargCollision:
+    """Regression tests for #863: output_type kwarg collision in
+    _build_request_params when output_type leaks into _default_model_params
+    (e.g., via @mesh.llm decorator **kwargs capture)."""
+
+    def test_build_request_params_no_kwarg_collision_when_output_type_in_default_params(
+        self,
+    ):
+        """#863: prepare_request was failing with 'got multiple values for
+        keyword argument output_type' when output_type leaked into
+        _default_model_params (decorator **kwargs capture path)."""
+        from _mcp_mesh.engine.mesh_llm_agent import MeshLlmAgent
+
+        agent = MeshLlmAgent(
+            config=make_test_config(
+                provider="claude",
+                model="claude-3-5-sonnet-20241022",
+                api_key="test-key",
+                max_iterations=10,
+            ),
+            filtered_tools=[],
+            output_type=ChatResponse,
+        )
+
+        # Simulate output_type leaking into _default_model_params via the
+        # decorator path (resolved_config.update(kwargs) in @mesh.llm).
+        agent._default_model_params["output_type"] = ChatResponse
+
+        # Capture prepare_request invocation; assert no TypeError raised
+        # and output_type arrives exactly once with the agent's value.
+        captured: dict = {}
+
+        def fake_prepare_request(*, messages, tools, output_type, **kwargs):
+            captured["messages"] = messages
+            captured["tools"] = tools
+            captured["output_type"] = output_type
+            captured["kwargs"] = kwargs
+            return {
+                "messages": messages,
+                "tools": tools,
+                "output_type": output_type,
+                **kwargs,
+            }
+
+        with patch.object(
+            agent._provider_handler,
+            "prepare_request",
+            side_effect=fake_prepare_request,
+        ):
+            params = agent._build_request_params(
+                [{"role": "user", "content": "hi"}]
+            )
+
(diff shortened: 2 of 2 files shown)


dhyansraj:
Combining with #866 and #860 into a single LLM-stack-cleanup PR for batch review.

PR #1212, opened 2026-06-12
fix(python): build SSE route endpoints at decoration time — settle grace holds on streaming paths

### LINKED ISSUE #1206: python: settling-window grace does not hold calls on the mesh.Stream SSE gateway path
## Problem

The #1193 settling-window grace does not hold calls on the `mesh.Stream` SSE gateway path. Observed during the test-suite settle migration (two tests reverted to their sleeps):

- Single-hop: an HTTP request hitting a `@mesh.route` whose handler uses `mesh.Stream` while the dependency is still unresolved returns a gateway 500 instead of the wrapper holding for the settle window.
- Multi-hop passthrough: an intermediate agent calling `McpMeshTool.stream()` proceeds before its dep resolves rather than waiting.

The non-streaming DI wrapper paths (tool, route, claim) hold correctly — the gap is specific to the streaming invocation path, likely related to the known FastAPI streaming-route wrapping traps (the stream wrapper's settle-keys integration may not cover the path the SSE endpoint actually takes).

Repro: `tests/integration/suites/uc18_streaming/tc01_single_hop_sse` and `tc02_multi_hop_passthrough` with their dependency-resolution sleeps removed and `MCP_MESH_SETTLE_TIMEOUT=60` set (the migration's standard transform — currently reverted for these two).

## Acceptance
Both uc18 tests pass with their sleeps removed; the streaming path's settle behavior matches the non-streaming wrappers (hold during settling, today-behavior after).

dhyansraj:
## Summary
Closes #1206 — the last real gap from the settle-grace rollout, and with it the last dependency-resolution sleeps in the entire integration suite.

- **Root cause (traced via live repro; all three working hypotheses disproven)**: a startup-ordering race the settle hold widened into near-certainty. `@mesh.route` registered the *plain* DI wrapper with FastAPI at import; the SSE endpoint was only swapped in by the debounced route-integration step ~1–2s after uvicorn started serving. Pre-settle, a request in that window failed fast on the unresolved dep; post-settle it was *held across the swap*, resolved, and the plain wrapper returned a raw async generator FastAPI can't serialize → 500. The reported multi-hop `.stream()` shape was misattribution — the stream wrapper's hold was verified correct all along.
- **Fix**: streaming routes build their SSE endpoint at decoration time, exactly like non-streaming routes always did — eliminating the window and the post-registration-swap FastAPI traps on the primary path. The endpoint resolves its inner DI wrapper *dynamically* (own-attribute read at call time), and the integration step converges it to the preferred wrapper instance in the double-import (`__main__` + named module) scenario, registering that instance for the heartbeat funnel. The old swap survives only as a loudly-logged fallback (consequence stated) if the build fails.
- **Two latent bugs fixed en route**: the integration step re-wrapped SSE endpoints on a second pipeline run (re-entrancy); and the old path's "CRITICAL FIX" convergence block was **dead code** — it queried the dependency mapping with the wrong key shape, so `affected_functions` was always empty. The extracted helper implements the intended semantics against the real keying (conservative: same source qualname+location, converge away from `__main__` only) — making convergence functional for non-SSE routes for the first time (flagged as an intended semantic activation, not just refactoring).
- **Settle-state addition**: `retire_declared` — an abandoned twin's orphaned declared keys would otherwise pin the eager latch open for the full window; retirement re-checks the latch and wakes parked waiters.
- **Contract preserved**: settle timeout → None → the user's defensive branch (a clean 503 in the canonical shape); no stream-chunk buffering — the hold completes before streaming begins.
- **uc18 tc01/tc02 re-migrated sleep-free** (closing the final #1209 exclusion), with a bounded gateway `/api/health` poll covering the bind-liveness gap the deleted sleep had been incidentally masking (the gateway registers as `api-<hash>`, so named-agent polls never covered it).

## Review Notes
- Independent review verified all six documented streaming-trap risk areas: decoration-time inputs all stamped before the build, the old annotation-strip's three jobs subsumed by the rebuilt signature, the by-reference dep-array chain mutation-only, fallback triggers environmental-only, re-entrancy idempotent, non-streaming routes byte-identical.
- Its warning (SSE branch bypassing multi-instance convergence — permanent heartbeat disconnect in the double-import scenario) drove the dynamic-resolution + convergence + key-retirement round, pinned by a dual-import test that exercises the full chain through a real FastAPI TestClient.

Closes #1206

## Test plan
- [x] Full Python unit suite: 1163 passed (15 new stream-settle tests incl. the pre-integration-500 regression pin and the dual-import convergence chain)
- [x] End-to-end repro replay: previously-500 single-hop → held 5.5s, streamed 9/9 chunks, HTTP 200; multi-hop both-hops-settling → 200
- [x] uc18 tc01/tc02 sleep-free on the cluster: 4 rounds green (plus a post-convergence rerun)
- [x] src-tests image rebuild green

## Summary by CodeRabbit

* **New Features**
  * Streaming routes may build SSE endpoints at decoration time for faster, more reliable startup.
  * Decoration-time validation now fails fast for unsupported stream element types.

* **Bug Fixes**
  * Improved convergence so heartbeat updates target the correct wrapper in dual-import scenarios.
  * Settle logic now retires unresolved dependencies cleanly and avoids lingering waits.
  * Integration tests use health polling and extended settle timeout for robust startup checks.

### DIFF
--- a/src/runtime/python/_mcp_mesh/engine/settle.py
+++ b/src/runtime/python/_mcp_mesh/engine/settle.py
@@ -179,6 +179,37 @@ class SettleState:
                 # nothing left to wake.
                 pass
 
+    def retire_declared(self, dep_key: str) -> None:
+        """Remove a declared key that no update path will ever resolve.
+
+        Used by route-wrapper convergence in the dual-import scenario
+        (``python main.py`` + ``from main import X``): each decoration pass
+        created its own DI wrapper and declared its own func_id-scoped keys,
+        but after convergence only the preferred instance receives heartbeat
+        updates. Leaving the abandoned instance's keys declared would pin
+        the eager latch open for the full window — every graced call would
+        then wait out the budget even after all live dependencies resolved.
+
+        Any waiter already parked on the retired key (a call that collected
+        its pending set just before convergence) is woken so it proceeds
+        immediately instead of dead-waiting a key that can never resolve.
+        """
+        with self._lock:
+            self._declared.discard(dep_key)
+            event = self._events.get(dep_key)
+            async_waiters = list(self._async_waiters.get(dep_key, ()))
+            if self._declared and self._declared <= self._resolved:
+                # Eager latch: the retired key was the last unresolved one.
+                self._settled = True
+        if event is not None:
+            event.set()
+        for loop, async_event in async_waiters:
+            try:
+                loop.call_soon_threadsafe(async_event.set)
+            except RuntimeError:
+                # The waiter's loop already closed (shutdown mid-wait).
+                pass
+
     def is_settled(self) -> bool:
         """Permanent latch check; flips on window expiry or timeout=0."""
         if self._settled:
--- a/src/runtime/python/_mcp_mesh/engine/settle.py
+++ b/src/runtime/python/_mcp_mesh/engine/settle.py
@@ -121,6 +121,12 @@ class SettleState:
         self._lock = threading.Lock()
         self._declared: set[str] = set()
         self._resolved: set[str] = set()
+        # Terminal-but-NOT-resolved keys (see retire_declared): retired by
+        # route-wrapper convergence because no update path will ever resolve
+        # them. Semantically distinct from _resolved — no proxy ever landed —
+        # but both mean "stop waiting": late waits short-circuit on retired
+        # keys exactly like resolved ones.
+        self._retired: set[str] = set()
         self._events: dict[str, threading.Event] = {}
         # Loop-native mirrors for async waiters: dep_key -> [(loop, event)].
         # Set via loop.call_soon_threadsafe from the resolution funnel so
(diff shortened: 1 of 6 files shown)


dhyansraj:
Review-round note for release-note visibility: the stream-annotation fix is a deliberate decoration-time contract tightening — `Stream[bytes]` (or any non-`str` stream type) previously decorated silently and degraded to a non-streaming route (accumulate-and-serialize, usually broken downstream); it now fails at import with a descriptive ValueError naming the supported form, matching the `@mesh.tool` path's existing behavior. Benign non-stream routes and hint-resolution failures are untouched.

Also in this round: retired settle keys are now terminal (a late wait on an abandoned wrapper twin's key short-circuits instead of parking for the full budget — `_retired` kept distinct from `_resolved` since no proxy ever landed), and the re-entrancy test asserts the full second-pass contract.

PR #1248, opened 2026-07-02
fix(llm): guarantee reply envelope carries the answer as a string

### LINKED ISSUE #1247: fix(llm): provider reply envelope can carry empty or non-string content for structured output; consumers degrade silently
## Contract

The LLM provider loop returns every reply to consumers as a `{"role": ..., "content": ...}` envelope. The intended contract is that `content` always carries the answer as a string (JSON string for structured output). Consumer runtimes (Python, TypeScript, Java) parse the declared response type out of `content`.

## Violations

**Provider side (Python):**

1. **Structured-output answer lost when upstream delivers it in a non-text carrier.** The native adapters build `message.content` by joining text blocks/parts only; the `output_config` path in `helpers.py` then emits `content: ""` when the join is empty. Upstream client libraries have moved (and continue to move) structured output between carriers — synthetic tool calls (`tool_use` block / `arguments`), native text blocks, and parsed/structured fields (`message.parsed`, `provider_specific_fields`). Any carrier the text-join doesn't see collapses the envelope to empty content, silently discarding a valid answer. The Gemini adapter has the identical text-join assumption; the OpenAI adapter had no handling for a `parsed`-only response with `content=None`.
2. **No diagnostic on the unrecoverable case.** When the answer genuinely can't be recovered, the envelope goes out with `content: ""` and no warning, so the failure surfaces far downstream as a misleading response-model parse error.

**Consumer side (all three runtimes):**

3. **Silent degradation on malformed envelopes.** Content that is a dict/Map (structured answer leaked unserialized) or an envelope-less bare map resolves to `""`/`undefined`/`None`, and the eventual error ("failed to parse LLM response as <Model>") blames the answer instead of showing the malformed payload. The LLM's actual output is discarded even when it is exactly the declared response shape.

## Fix

**Provider (primary — the contract is minted here):**
- `output_config` path recovers the structured answer from `message.parsed` / `provider_specific_fields` when the text-join is empty; logs a warning naming model + mode + truncated raw message when nothing is recoverable.
- Envelope content is coerced to a string at every construction site (`_coerce_content_to_str`), always valid JSON.
- OpenAI native adapter recovers from `parsed` when `content` is `None` — gated on the message having no tool calls, so content is never fabricated on an ordinary tool-call turn (the same invariant is enforced at the loop layer; intermediate tool-call turns replay with unchanged history).

**Consumers (defense in depth, all three runtimes):**
- Dict/Map-shaped `content` is serialized to a JSON string and parsed normally.
- A bare map with no `content`/`role` keys is treated as the answer itself (excluding maps carrying `error` / `tool_calls` / `_mesh_usage`, which route to the diagnostic path instead so real failures stay visible).
- Empty-content parse failures now include a truncated snippet of the raw provider payload.

## Notes

- Happy path (text content present) is behavior-identical; adapter text-join expressions are unchanged.
- New unit tests pin the string-typed content contract, the recovery priorities, the bare-map exclusions, and — critically — that replayed assistant history on intermediate tool-call turns is unchanged.

dhyansraj:
## Summary
- **Provider (Python, primary fix)**: the `{role, content}` reply envelope now always carries the answer as a string. The `output_config` structured-output path recovers the answer from `message.parsed` / `provider_specific_fields` when the text-block join is empty (upstream client libraries move structured output between text blocks, tool-call arguments, and parsed fields across versions); envelope content is coerced to valid-JSON strings at every construction site; a warning naming model + mode + truncated raw message fires when nothing is recoverable, instead of silently emitting empty content.
- **OpenAI native adapter**: recovers a `parsed`-only response (`content=None`) — gated on the message having no tool calls, so content is never fabricated on an ordinary tool-call turn and replayed assistant history is byte-for-byte unchanged (pinned by tests).
- **Consumers (Python / TypeScript / Java, defense in depth)**: dict/Map-shaped `content` is serialized and parsed normally; a bare envelope-less map is recovered as the answer itself (excluding maps carrying a truthy `error`, `tool_calls`, or `_mesh_usage` — those route to the diagnostic path so real failures stay visible); empty-content parse failures now include a truncated snippet of the raw provider payload instead of a bare "failed to parse as <Model>".

## Review Notes
Two independent zero-context review passes were applied during development:
- Pass 1 found (and the fix removed) an adapter-level fallback that would have leaked tool-call arguments into replayed assistant history on every agentic tool turn; recovery now lives only where the no-tool-calls context is known.
- Pass 2: 0 blockers; both warnings fixed (truthy-`error` bare-map exclusion; hardened `json.dumps` in the Python consumer). Remaining known-minor INFOs, deliberately deferred: TS serializes array-shaped content whole where Java/Python extract the first block's text; TS streaming path's empty-content diagnostic falls back to the last message rather than the raw wire payload; Java uses `isBlank()` where Python/TS check `== ""`.

Closes #1247

## Test plan
- [x] Unit tests across all three runtimes: string-content contract, recovery priorities (`parsed` > `provider_specific_fields`), bare-map recovery + exclusions, empty-content diagnostics, non-serializable-leaf hardening
- [x] Regression pins: happy path (text present) behavior-identical; intermediate tool-call turns replay with unchanged history
- [x] `tsuite run --suite-path tests/src-tests` — 12/12 (all SDKs build, local image rebuilt)
- [x] `tsuite run --suite-path tests/integration --parallel 8` — full suite green (7 transient failures passed on targeted re-run)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery for structured outputs returned in unexpected forms (including nested wrappers and bare answer maps), reducing empty-response parsing failures.
  * Tool-call-only and metadata-only payloads are no longer misclassified as final answers.
  * When final content is blank, thrown parsing errors now include a truncated snippet of the original provider payload for easier troubleshooting.
  * Enhanced multi-runtime content normalization so non-text carriers and structured “parsed” fields are handled consistently.
* **Tests**
  * Added extensive coverage for content recovery and “empty content” diagnostics across agent loops and delegated providers.

### DIFF
--- a/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshLlmAgentProxy.java
+++ b/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshLlmAgentProxy.java
@@ -1430,13 +1430,32 @@ public class MeshLlmAgentProxy implements MeshLlmAgent {
         if (trimmed.startsWith("{") && trimmed.endsWith("}")) {
             try {
                 Map<String, Object> parsed = objectMapper.readValue(trimmed, Map.class);
-                // If it has a "content" field, extract it (this is the LLM's actual response)
-                if (parsed.containsKey("content")) {
+                // Provider envelope wrapper? (carries a content/role marker.) Mirror
+                // the top-level extractContent recovery so a nested envelope — the
+                // MCP content[0].text carrying the serialized {role, content} object
+                // — is handled symmetrically instead of returning the wrapper JSON.
+                if (parsed.containsKey("content") || parsed.containsKey("role")) {
                     Object innerContent = parsed.get("content");
                     if (innerContent instanceof String s) {
                         log.debug("Extracted inner content from LLM provider wrapper");
                         return s;
                     }
+                    if (innerContent instanceof Map<?, ?> innerMap) {
+                        return serializeContent(innerMap,
+                            "structured (Map) content nested in a provider wrapper");
+                    }
+                    // No usable string/map content (null, empty, tool-call turn, or a
+                    // truthy nested error) — return "" so empty-content diagnostics
+                    // fire instead of feeding the wrapper JSON to the schema parser.
+                    return "";
+                }
+                // Not an envelope. A bare payload carrying reserved wire keys or a
+                // truthy "error" is NOT an answer either — surface it via empty-
+                // content diagnostics rather than returning the error/tool JSON.
+                if (parsed.containsKey("tool_calls")
+                        || parsed.containsKey("_mesh_usage")
+                        || isTruthy(parsed.get("error"))) {
+                    return "";
                 }
             } catch (JacksonException e) {
                 // Not valid JSON, return as-is
--- a/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshLlmAgentProxy.java
+++ b/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshLlmAgentProxy.java
@@ -424,6 +424,12 @@ public class MeshLlmAgentProxy implements MeshLlmAgent {
         private long accumulatedOutputTokens = 0;
         private String effectiveModel = null;
 
+        // Raw provider envelope of the final (no-tool-calls) response, captured so
+        // generate(Class) can surface it when content resolves to empty and the
+        // structured parse fails — the honest signal is "empty content", not
+        // "failed to parse as <Model>".
+        private Map<String, Object> lastRawResponse = null;
+
         // --- Messages ---
 
         @Override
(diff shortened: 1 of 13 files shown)


dhyansraj:
Review comments addressed in cce32cb52:

- **Java nested-envelope recovery** (`parseNestedContent`) — valid, fixed. A nested wrapper whose inner `content` was a map fell through the `instanceof String` check and returned the whole wrapper JSON; a nested truthy `error` was likewise returned verbatim as the answer. The nested path now mirrors the top-level `extractContent` guards (map content serialized, truthy `error`/reserved keys route to the empty-content diagnostics). +5 tests.
- **`_raw_response_snippet` → `_dumps_safe`** — valid, fixed. Non-serializable payloads previously degraded the diagnostic to an object repr. +1 test.
- **`_parsed_to_json_str` missing `default=str`** — valid, fixed. A non-JSON leaf (e.g. datetime) silently turned the recovered answer into `None`. +1 test.
- **Consolidating the three safe-JSON helpers** — skipped. The genuinely shared core is a single `json.dumps(x, default=str)` line; the rest differs by design (None-handling, string passthrough, Pydantic branch, failure fallback). No existing shared module fits (`shared/json_fast.py` is orjson with different `default=` semantics), so consolidation would add a new dependency edge from the deliberately-lean adapter layer for negligible dedup. The drift concern is addressed substantively: all three helpers now share identical `default=str` + fallback semantics.

Java module 27/27, Python affected + regression suites all green.

Issue #1252, opened 2026-07-03
design(jobs): claim epochs, poll-liveness, and per-filter event cursors — multi-replica execution integrity


dhyansraj:
## Problem (field-verified on 2.7.0)

A long-running interactive MeshJob (sequential `recvEvent(type=X, timeout)` gates, consumer capability at 2+ replicas) corrupts when one gate legitimately blocks quietly past the lease window:

1. **Quiet waiting earns no liveness.** The lease renews on exactly one operation — an accepted non-terminal `/jobs/batch` delta (`ent_service_jobs.go:340-352`). `recv_event` is an anonymous read (`task_backend.rs:436` "anyone-side") that round-trips the registry every poll yet never touches the lease. A handler blocked in a legitimate wait is indistinguishable from a wedged one; the lease expires and the reclaim sweep re-dispatches the job.
2. **Reclaim does not fence the previous owner.** There is no claim generation/epoch anywhere; the only identity is `owner_instance_id`. After re-dispatch: the old owner's *writes* are rejected cross-instance (but NOT on a same-instance re-claim, where fencing is zero), its *reads* are never checked at all, and the new owner's controller replays the event log from seq 0 while the old handler keeps executing — duplicate live execution of non-idempotent work. Documented as a known non-goal at `ent_service_jobs.go:969-976`; this issue promotes it to a requirement.
3. **The per-controller event cursor is shared across type filters** (`jobs.rs:259-262`; delivery strictly `seq > last_seen_seq`). Consuming a type-A match at seq N permanently skips type-B events at seq < N. Sequential in-order gates never hit this — two out-of-phase owners (defect 2) hit it constantly, silently swallowing events. (Note: the reported "empty filtered page leaps the cursor" mechanism was investigated and refuted — mock and registry agree, both return `next_after = after` on no-match pages. The shared-cursor hazard above is the real defect; the misleading client comment at `jobs.rs:617-620` should be fixed.)

Compounding: missed input → quiet gate → lease expiry → dual owners → phase slip → cursor swallowing → every subsequent round runs to its full deadline, duplicated.

**Interim mitigation for affected deployments:** declare `max_duration` ≈ the job's real ceiling. `leaseWindowFor` derives the lease window from it (default 300s otherwise), and the stale sweep already honors `max(MCP_MESH_JOB_STALE_TIMEOUT, max_duration)` (#1244).

### Phase 1 — claim epochs (fencing)
[...]

PR #1253, opened 2026-07-03
feat(jobs): claim-epoch fencing and poll-liveness (Phases 1+2 of #1252)

dhyansraj:
## Summary
Implements Phases 1+2 of #1252 — execution integrity for MeshJobs under multi-replica consumers and mid-execution reclaim.

- **Claim epochs (fencing)**: every successful claim/re-claim mints a monotonic `claim_epoch` in the same guarded UPDATE that assigns the owner. `/jobs/batch` deltas carry it; any epoch-bearing delta whose `(owner, epoch)` doesn't match the live row — cross-instance, reclaimed `owner=NULL`, or same-instance re-claim — rejects as `claim_superseded`. Epoch-less deltas keep legacy `not_owner` semantics.
- **Supersession abort, frame-exact**: the core translates `claim_superseded` (batch rejection or executor-read 409) into firing *that execution's own* cancel frame — keyed by epoch, never top-of-stack — so a superseded zombie aborts while a healthy same-process re-claim keeps running. `is_cancelled()` (the Java poll surface) reads its own epoch frame too. User cancels (`POST /jobs/{id}/cancel`) still target the live attempt by design.
- **Poll-liveness**: an epoch-valid executor `recv_event` poll extends the lease (capped by `total_deadline` and the `max_duration`-derived stale ceiling — #1244 semantics preserved, cap math structurally shared with the sweep). A handler legitimately blocked in an event gate is no longer reclaimed as wedged; artificial `progress()` keepalives are unnecessary.
- **Cancel responsiveness**: the `recv_event` long-poll races the cancellation token, so user cancels and supersession interrupt blocked gates promptly.
- **Observer purity & skew**: event reads without the paired `instance_id`+`claim_epoch` params are anonymous observer reads, byte-identical to before (A2A unaffected); half-paired params 400. Old SDK + new registry and new SDK + old registry both degrade to today's behavior; epoch 0 is never fabricated. Registry 409s are classified `claim_superseded` only when the body carries the marker — proxy/ingress 409s stay retryable.
- **Surfaces**: `claimEpoch` exposed read-only on `JobContext` in Python/TS/Java; all three SDKs thread the epoch claim→controller via the shared Rust core (new `mesh_job_controller_new_with_epoch` FFI entry point; existing 4-arg symbol untouched).

## Review Notes
Zero-context review applied; all findings fixed — including a HIGH where supersession originally fired the top cancel frame (the healthy re-claim) instead of the zombie in same-instance scenarios, now epoch-targeted with a two-frame test. `MESHJOB_DESIGN.org` gains a fencing & liveness section; the "reclaim is not mutual exclusion" non-goal comment is superseded.

Part of #1252 (Phases 3-5 follow: per-filter cursors, multi-replica integration UC, man pages).

## Test plan
- [x] Registry Go: full package incl. new epoch suite (mint/re-claim, cross-instance + same-instance + reclaimed-NULL supersession, lease extension + caps, observer no-op, half-pair 400, terminal reads)
- [x] Rust core: 461 tests incl. two-frame supersession targeting, 409 marker classification, long-poll cancel race; all binding features check clean
- [x] SDK unit: Python 189, TS 35 affected suites, Java 514+15+98 across modules
- [x] `tsuite run --suite-path tests/src-tests` — 12/12
- [x] Full integration — 511 passed; 9 Java-LLM-consumer failures triaged as cold-JVM timing flakes (diff structurally exonerated; all 9 pass on re-run)

## Summary by CodeRabbit

* **New Features**
  * Added claim-epoch support across job claims, updates, and event reads.
  * Job controllers and runtime contexts now carry an optional claim generation value.
  * New APIs and headers expose this value to Java, Python, TypeScript, C, and native integrations.

* **Bug Fixes**
  * Improved fencing so stale job owners and outdated retries are rejected consistently.
  * Executor reads now distinguish from observer reads and return clearer errors when identity information is incomplete or superseded.
  * Lease handling and cancellation now respect the active claim epoch more accurately.


dhyansraj:
Review comments addressed in babfdb348 — all four verified valid and fixed:

- **TS `readJobHeaders` permissive parse** — real: `parseInt("5.5")`/`parseInt("5abc")` slipped through as 5. Now a strict whole-string `/^\d+$/` match; malformed-input tests added. Python (`int()` raises) and Java (`Long.parseLong` throws) verified already strict — no cross-runtime hole.
- **Python duplicated epoch validation** — extracted `_valid_claim_epoch`, both call sites, byte-identical behavior.
- **Redundant job lookup on executor reads** — split service method: executor path reuses the row existence proven by `AuthorizeExecutorRead` (`listJobEventsCore`), observer path keeps the public wrapper with its `Exist` check. Status-code behavior preserved exactly; executor-path 404 now pinned by its own test.
- **Per-request stale-timeout env parse** — cached once at `EntService` construction (verified no test mutates the env var across executor reads; sweep behavior untouched).

Go registry package, scoped TS (22) and Python (188) suites all green; vet/build clean.

Issue #1193, opened 2026-06-11
runtime: settling-window grace for dependency injection — bounded wait for resolution during agent startup


dhyansraj:
## Problem

Dependency injection resolves asynchronously: an agent's declared dependencies become available when the first full heartbeat cycle completes and `dependency_available` events arrive. A tool call (or background-loop first pass) that fires during this settling window hits a declared-but-not-yet-resolved dependency and fails immediately — even though resolution typically lands moments later. The window is most visible in interactive dev loops (calls arrive seconds after agent start) and during prod rolling restarts.

Today's failure is harmless but noisy, and the only mitigations are app-side retries or sleeps.

## Proposed design — settling-window grace (auto-detected, no per-agent code)

When a call needs a dependency that is **declared but unresolved**, AND the agent is still settling (first full heartbeat cycle incomplete, or uptime below a small threshold), the injected proxy waits — bounded by the remaining settle budget — for the `dependency_available` event before failing. After the window closes, current fail-fast behavior is unchanged.

Properties:
- No new API surface; behavior is adaptive rather than configured (an explicit `await deps_ready()` style hook is deliberately NOT the primary design).
- Bounded wait only during settling → no steady-state blocking, no deadlock exposure for circular topologies.
- Benefits dev loops and prod rolling restarts identically.
- Cross-runtime: Python/TS/Java proxies all consume the same Rust-core event stream, which already carries the needed signal.

## Open design questions
- Settling signal: first-full-heartbeat-completed flag from the core vs uptime threshold vs both.
- Budget default (a few seconds?) and the env knob to tune/disable.
- Whether background-task primitives (cron-style entry points) get the same grace on their first pass (they should).

## Acceptance criteria
- A tool call arriving before first resolution succeeds once the dependency resolves within the budget, with no app-side code.
- Post-settling behavior byte-identical to today.
- All four runtimes consistent; integration test simulating the early-call race.

dhyansraj:
Design settled after validation against the shipped examples in all three runtimes:

- **Wrapper-level wait, never a placeholder**: the grace lives in each runtime's DI wrapper, before the user function runs. Params remain binary — real proxy or None/null — so the defensive `if dep:` idiom in every shipped example is untouched; on timeout the call proceeds exactly as today. No pending state is ever observable.
[...]

PR #1254, opened 2026-07-03
feat(jobs): per-filter event cursors, multi-replica tests, lease docs (closes #1252)

### LINKED ISSUE #1252: design(jobs): claim epochs, poll-liveness, and per-filter event cursors — multi-replica execution integrity

dhyansraj:
## Summary
Completes #1252 (Phases 3-5; Phases 1+2 landed in #1253).

- **Per-filter event cursors (Phase 3)**: `JobController` tracks an independent cursor per canonical type-filter (sorted/deduped, trimmed, empty types dropped to match registry semantics; unfiltered is its own stream). Exactly-once within a stream, documented at-least-once across streams; interleaved gates with different filters can no longer permanently skip each other's events. Per-filter locks replace the global recv lock, so a 60s long-poll on one type doesn't block another — while same-filter calls stay serialized and monotonic. The epoch/supersession logic from #1253 is untouched (verified byte-intact in review). The mock-vs-registry page contract is pinned by a dual-backend scenario table with exact path/query matching (the registry's own behavior is pinned Go-side), and the in-process mock now applies the registry's trim/drop-empties filter rule.
- **Multi-replica integration tests (Phase 4, new `uc33_meshjob_replicas`)**: two 2-replica scenarios from the field report — `tc01` proves a quietly-gating handler polling `recvEvent` through 2× its lease window keeps a single claim (poll-liveness), and `tc02` proves a genuinely-wedged handler is re-claimed with the stale owner fenced (epochs `[1,2]`, exactly one surviving owner, job completes once). Assertions run against the registry job API with the epoch/attempt/transition evidence; timing margins sized for CI load.
- **Harness hygiene**: the nine cold-JVM-sensitive tests (uc10 java-consumer LLM suite, uc06 java tracing, uc20 committee) now poll-until-registered with a 240s deadline instead of sleeping a fixed 25s; test-level budgets raised where the waits stack. These were the recurring registration flakes in full-suite runs — all nine pass under parallel-8 in this PR's gate run.
- **Docs (Phase 5)**: jobs man pages (base + TS/Java variants), `environment.md`, and `docs/concepts/jobs.md` now document the lease window derivation (`max_duration`, 300s default), what renews a lease (progress deltas AND executor `recvEvent` polls), multi-replica claiming and epoch fencing, per-runtime supersession surfaces, per-filter cursor semantics with checkpoint guidance, `claimEpoch` accessors, and the `MCP_MESH_JOB_STALE_TIMEOUT` reap-ceiling vs lease-window distinction. Every claim was verified against the implementation before writing.

## Review Notes
Zero-context review applied (0 HIGH); all findings fixed: contract-test circularity (now matches `after`/`types` per scenario), tc02 lease-margin hardening (2s finish-gate rounds), retention raised to 60s with the 10s sweep tick kept, fail-fast re-claim wait, filter-key input hygiene + honest edge-case docs. Backlog note from testing: `MCP_MESH_AGENT_NAME` seeds only the instance-id prefix — there is no env knob for a shared registered name across replicas.

Closes #1252

## Test plan
- [x] Rust core 468 tests (incl. the shared-cursor defect pin, canonicalization, concurrency, dual-backend contract table)
- [x] `tsuite run --suite-path tests/src-tests` — 12/12
- [x] uc33 scoped — 2/2, plus green in the full run
- [x] Full integration — 507 passed + 15 API-credit-exhaustion failures, all 15 green on re-run after top-up (522/522 accounted); uc21/uc22 jobs suites and all nine harness-fixed tests pass under parallel-8

## Summary by CodeRabbit

* **New Features**
  * Improved `recv_event` to treat each event type filter as its own independent stream with separate cursors, preventing cross-filter skipping.
  * Added multi-replica execution fencing with a per-claim epoch to prevent stale executions from winning.

* **Bug Fixes**
  * Refined lease renewal/reaping behavior for quiet and long-running jobs to rely on accepted non-terminal events and active event polling.
  * Clarified supersession and cancellation propagation so superseded work can’t overwrite newer results.

* **Documentation**
  * Updated job, environment, and JavaScript/Java semantics for event cursors, lease recovery, and fencing.

* **Tests**
  * Added UC33 replica scenarios and fixtures; extended timeouts and unified agent readiness polling.

### DIFF
--- a/docs/concepts/jobs.md
+++ b/docs/concepts/jobs.md
@@ -485,9 +485,42 @@ such jobs run unbounded, subject only to lease recovery.
 
 For a job that runs for hours: set `max_duration` to its real
 per-attempt runtime (this sizes the lease AND floors the stale
-ceiling) and/or emit periodic progress to keep the lease renewed. Set
-`total_deadline` if you want a hard total-runtime bound across all
-attempts, or to opt out of the registry-wide stale ceiling entirely.
+ceiling). A handler sitting in `recv_event` gates renews the lease on
+every poll; one that computes silently for long stretches should emit
+periodic progress to keep the lease renewed. Set `total_deadline` if you
+want a hard total-runtime bound across all attempts, or to opt out of the
+registry-wide stale ceiling entirely.
+
+`MCP_MESH_JOB_STALE_TIMEOUT` is a **reap ceiling, not the lease window** —
+it caps total runtime from submission but does not size or extend the
+per-attempt lease. The lease window is derived per-job from `max_duration`
+(300s default) and is renewed by progress deltas and executor `recv_event`
+polls; tuning the stale timeout does not change how quickly a quiet
+handler's lease expires.
+
+### Multi-replica fencing and poll-liveness
+
+Several replicas can declare the same `task=true` capability. Each job is
+claimed by exactly one replica per attempt (the guarded atomic `UPDATE`
+above, so concurrent claimers race and one wins). The owner's lease is
+renewed by **any accepted non-terminal delta and any executor `recv_event`
+poll from the current claim** — a handler parked in a legitimate
+`recv_event` gate is provably alive, so no artificial `update_progress`
+keepalives are needed. Poll-liveness is capped so it never pushes the lease
+past `total_deadline` or the stale ceiling.
+
+If a lease genuinely expires (a handler wedged, neither progressing nor
+polling), the sweep returns the job to the pool and a peer re-claims it.
+Every claim — including a same-replica re-claim — carries a monotonically
+increasing **claim epoch**. The registry fences the superseded execution:
+its writes are rejected `claim_superseded` and its cancel token fires, so a
+parked `recv_event` breaks out with the cancelled error and the handler
+observes cancellation exactly as it would a user cancel. Handlers should
+honor cancellation promptly; the read-only `claimEpoch` accessor on the job
+context (`job.claim_epoch` in Python, `currentJob()?.claimEpoch` in
+TypeScript, `JobContext.current().claimEpoch` in Java) can be stamped onto
+external side effects so downstream can dedupe a fenced re-execution's
+duplicate write.
 
 The **agent runtime** is a thin client. On the producer side, it
 maintains a claim queue per `task=true` capability (`UPDATE jobs SET
@@ -564,7 +597,7 @@ sequenceDiagram
     Note over R: Append-only log<br/>(per-job seq)
     H->>R: GET /jobs/{id}/events?after=N  (long-poll)
     R-->>H: event dict (or empty after timeout)
-    Note over H: Cursor advances<br/>per-controller
+    Note over H: Cursor advances<br/>per-controller, per-filter
 ```
 
 ### Receiving events inside a handler
@@ -640,9 +673,15 @@ The producer-side recv loop. Filter by `types` to drop noise, set a
     ```
 <!-- markdownlint-enable MD046 -->
 
-The handler's `recv_event` cursor is per-`JobController` instance — a
-fresh controller for the same `job_id` replays from `seq=0`. The
-running handler always sees events in monotonic seq order.
+The handler's `recv_event` cursor is per-`JobController` instance **and
+per-`types` filter**: each distinct filter is an independent stream with
+its own cursor, so interleaving `recv_event(types=["A"])` and
+`recv_event(types=["B"])` never lets one filter's consumption skip the
+other's earlier events. Delivery is exactly-once within a filter stream
+and at-least-once across different filters. A fresh controller for the
+same `job_id` (e.g. a re-claim) replays every filter from `seq=0`; the
+running handler always sees events in monotonic seq order within a
+stream.
 
 ### Posting events from outside the handler
 
--- a/src/core/cli/man/content/environment.md
+++ b/src/core/cli/man/content/environment.md
@@ -531,6 +531,12 @@ export MCP_MESH_SWEEP_INTERVAL=5m
 # set their own total_deadline are fully exempt. Default: off (unset /
 # "0") — jobs without an explicit total_deadline run unbounded, subject
 # only to lease recovery.
+#
+# NOT the lease window. This is a total-runtime REAP CEILING (measured
+# from submission); it does not size or extend the per-attempt lease. The
+# lease window is derived per-job from max_duration (300s default) and is
+# renewed by progress deltas / recvEvent polls — tuning this variable does
+# NOT change how quickly a quiet handler's lease expires.
 export MCP_MESH_JOB_STALE_TIMEOUT=2h
 
 # CORS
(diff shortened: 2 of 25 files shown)


dhyansraj:
Review comments addressed in afc1a270f:

- **`recv_event` per-filter lock vs timeout budget** — valid, fixed. The deadline is now computed before queueing; lock acquisition races the remaining budget (timeout semantics preserved: `Ok(None)`) and this execution's cancel token (a cancelled/superseded execution no longer parks on the lock behind a 60s long-poll). The poll loop's per-iteration budget derives from the same deadline, so lock-wait time is subtracted automatically. Epoch/supersession logic untouched.
- **Concurrency tests don't prove overlap** — valid, reworked. A gated mock backend (`Notify` + park) makes call 1 provably in-flight before call 2 starts; different-filter concurrency and same-filter serialization are now asserted under proven overlap, plus two new tests pinning the fixes above (queued call times out at its own budget; cancel-while-queued returns promptly). Core suite: 470 passed.
- **Registration wait duplication** — valid, extracted. New global routine `wait_for_agents_registered` (parameterized by agent list, liveness-only rule documented in its header) replaces the eleven inlined loops across the nine tests; behavior identical, validated by dry-run + scoped run + a full-suite run in which all nine pass through the routine.
- **Dep-aware readiness ("wait for 1/1 like tc03")** — skipped, with the inverse consistency fix. Per the settle-grace design (#1193), registration waits are liveness-only: dependency resolution is covered at call time by the settle window, and all nine tests pass under parallel-8 on that basis. tc03's loop was actually the outlier carrying a dep-aware grep — it's now normalized to liveness-only rather than spreading the condition.

Full suite on the refactored tree: 520 passed; the 2 failures are pre-existing uc04 structured-output tests in an untouched use case (re-run in progress to confirm).

Issue #1249, opened 2026-07-02
feat(di): required=true — transitive capability availability with perimeter 503


dhyansraj:
## Design (supersedes the original require()/@MeshDependency(required=true) call-time proposal — see edit history)

Soft-fail stays the default: unresolved deps inject null/unavailable proxies, agents always start and register, topology settles organically. This adds the opt-in strictness half as a **topology property instead of a runtime API**.

### The predicate

A dependency edge may be declared `required: true` (default false). Then:

> **capability availability = own agent healthy AND every required dependency's capability available**

computed by the registry, transitively. Optional edges never propagate — strictness flows only where explicitly declared, per edge.

### Why this shape

- **The propagation transport already exists.** The registry already marks capabilities unavailable when their agent goes unhealthy and pushes dependency updates that flip consumers' proxies. This changes only the availability predicate; delivery to consumers is the existing background update channel. Old-SDK consumers benefit with zero changes.
- **Hysteresis is inherited.** Agent health already debounces (3 missed heartbeats), and the only other propagation trigger is discrete topology events — no new flap source.
- **Runtime handling becomes nearly free.** Chain example: A→B (optional), B→C (required). C down ⇒ B unavailable ⇒ A's proxy for B goes null and A null-checks as today (its choice). If A had declared required on B, A goes unavailable too — and nobody holding a proxy to A can call it.

### Perimeter (external callers don't go through proxies)

Mesh-internal calls are gated by proxies; HTTP routes, direct MCP calls, and A2A entry are not. For a `@mesh.route`/`@MeshRoute` declaring a required dep, the framework's own wrapper returns **503** (with the capability reason) when the predicate is false — before user code, no hand-written requireDep. 503 (not 404): monitoring alarms on 5xx, LB health checks eject the instance (traffic-level unregistration), clients see retryable-unavailable rather than permanently-missing. Direct MCP tool calls on an unavailable capability get the equivalent error result. Same predicate, evaluated at the perimeter from proxy state already held locally.

### Correctness requirements

1. **Cycle detection**: a cycle among *required* edges can never converge (both sides unavailable forever). The registry detects required-edge cycles as registrations arrive and rejects the registration closing the cycle with an explicit error naming the loop. Cycles through optional edges remain legal (that's the bootstrapping story).
[...]

PR #1255, opened 2026-07-03
feat(di): required=true — transitive capability availability (Python first)

dhyansraj:
## Summary
First slice of #1249: dependency edges may declare `required: true`; the registry computes capability availability transitively (own agent healthy AND every required edge resolves under **full tag/version/schema matching** via the existing resolver) and propagates through the existing dependency-update channel — consumers' proxies gate mesh-internal calls with zero SDK changes. Routes declaring required deps auto-return 503 (`{"error":"dependency_unavailable","capability":...}`) before user code. Required-edge cycles are rejected loudly at registration AND on heartbeat metadata refresh; evaluation is memoized (O(nodes)) and fail-closed; the check-persist window is serialized. Reason strings name the first broken edge with constraint detail. Also: semantic registry rejections (HTTP 200 + `status:"error"`) now surface as repeating `registration_failed` errors — previously parsed as success, which could also tear down resolved deps on transient DB errors.

Soft-fail unchanged: `required` defaults false everywhere; optional edges never propagate; zero-required-edge meshes pay ~nothing on the hot path.

## Validation
- Live local e2e (registry + editable Python runtime): kill-provider chain propagation in one debounce cycle (~13s) with route 503 **while the intermediate agent was alive**; constraint-mismatch (healthy same-name provider must not satisfy a tag-narrowed required edge); cycle rejection repeating in agent logs; clean bootstrap ordering.
- Two zero-context review rounds; all findings fixed incl. a BLOCKER (heartbeat path bypassing cycle detection).
- Registry Go suite, Rust core 476, Python 1257 unit tests green; src-tests 12/12; full integration 520 passed + 2 pre-existing uc04 LLM flakes green on re-run.

## Follow-ups tracked in #1249 (stays open)
Java/TS declaration syntax, meshctl/UI reason-chain display, integration suite, man pages, jobs claim-gating.

## Summary by CodeRabbit

* **New Features**
  * Added support for marking dependencies as required or optional.
  * Capabilities now report whether they are available, plus a reason when they are not.
  * Routes can now enforce required dependency availability and return a clear 503 response when a required dependency is missing.

* **Bug Fixes**
  * Improved dependency handling so availability checks work across chained and shared dependency graphs.
  * Prevented invalid required-dependency cycles from being accepted.
  * Heartbeat errors are now surfaced more reliably instead of being treated as successful updates.


dhyansraj:
Review comments addressed in the latest commit:

- **`cycleWriteMu` panic safety** — valid, fixed: check + capability-write now run inside `guardedCapabilityWrite` with a deferred unlock; both `RegisterAgent` and `UpdateHeartbeat` call sites converted, post-transaction resolution work outside the scope. Whitespace-ignored logical diff is ~50 lines. New 24-goroutine concurrency test (mixed required/zero-required writes + a cycle-closer) passes 3× under `-race`.
- **Scan-under-lock hot path** — the scan was already gated on incoming required edges; the gap was the lock. Zero-required-edge writes now skip both lock and scan entirely (required-free meshes pay nothing). Scan narrowing beyond the existing in-memory filter deliberately skipped: the specs live in a JSON blob with no portable server-side index, the path is reached only by required-edge-declaring writes, and cache+invalidation isn't justified at registry capability counts — rationale documented on the function.
- **503 perimeter duplication** — valid, fixed: one shared `_route_perimeter_response` closure used by both async and sync wrappers, behavior byte-identical.
- **Streaming bypass silence** — valid, fixed: streaming routes declaring required deps now emit a creation-time warning that the 503 perimeter is bypassed by design; test added.
- **Semantic-rejection backoff** — skipped with evidence: the rejection path already rides the shared exponential backoff (base×2^attempt, 30s cap, equal jitter — heartbeat.rs:297-316), so permanent misconfigurations self-throttle to ~one loud log per 30s while healing without restart when fixed. Now documented with a `DELIBERATE:` comment at the site.

Registry Go suite + `-race` green; Python 1258; Rust 476.

Issue #1274, opened 2026-07-04
fix(meshctl): stop verifies and escalates against the parent PID, not the process group — child JVM lingers unkilled


dhyansraj:
## Contract
`meshctl stop` terminates the agent: it should return only when the process tree is dead, force-killing at the timeout.

## Violation (field-measured on 2.8.1)
For Maven agents the recorded PID is the `mvn spring-boot:run` parent (own pgid via `Setpgid:true`, start_commands.go:279). `KillVerifyAndCleanup`'s graceful branch (lifecycle/kill.go:254-265) sends SIGTERM to the **group** (correct — the child JVM receives it) but then `pollUntilDead(pid)` waits on the **parent only**, and the SIGKILL escalation is gated on parent liveness. Once the parent exits: stop returns "stopped" and removes pidfiles while the child Spring JVM is still bound to the port and heartbeating; a child that slow-walks or ignores SIGTERM is never force-killed (field measurement: port held ~3 minutes = Spring's own shutdown, not meshctl's timeout). The sibling orphan-cleanup branch (kill.go:228-251) already does this correctly with `pollUntilGroupDead` + group escalation — only the graceful branch regresses to parent scope.

## Fix
In the graceful branch, swap both waits to `pollUntilGroupDead` (SIGTERM-wait and post-SIGKILL wait), mirroring the orphan branch — stop then blocks until the whole group is dead and escalates SIGKILL to the group at the timeout. Consistency check the watcher's terminate/restart paths (watcher.go:271-376), which poll the group but share the escalation shape.

## Tests
Unit around the kill lifecycle where feasible; integration: stop a Maven agent whose JVM delays shutdown → stop blocks until port released or SIGKILLs at timeout; no orphaned heartbeating child; registry row deregisters.

## Docs rider (by-design behaviors confirmed in the same field report)
- `--watch` triggers on content writes (fsnotify Write/Create/Rename; mtime-only touch emits Chmod and is filtered) and is a rebuild trigger, NOT a crash supervisor (kill -9'd agents stay down until the next file change — deliberate, watcher.go:190-194). Document both in cli.md/tutorial.md watch sections.
- Document stop's process-group semantics once fixed.

dhyansraj:
[...]

Issue #1273, opened 2026-07-04
fix(di): direct tool invocation can reach a handler with a null required dependency during dependency recovery


dhyansraj:
## Contract

`required=true` (#1249): a handler never observes null for a required dependency — the framework refuses before invocation. Enforced today at the `@MeshRoute` perimeter (503 `dependency_unavailable`) and, since #1268 (v2.8.1), on the MeshJob claim paths (pre-claim skip + claim-invoke release).

## Violation (field-verified on 2.8.1, deterministic repro)

The direct `tools/call` dispatch has no required-dep check. When a required dependency flaps DOWN→UP, the registry-side view flips available immediately, but the CALLEE's injection cache refills only on its own next heartbeat — a remote tool call landing in that window (observed ~600ms) invokes the handler with the injected required dep NULL. The caller receives an NPE-shaped 500 instead of a classifiable topology error. Same two-clock race as #1268, different invocation path; the callee needs no restart to be exposed — any dependency recovery opens the window on every `required=true` tool it hosts.

## Fix

Extend the pre-invoke required-dep check (required flags already threaded into the wrappers by #1268) to the inbound tool dispatch in all three runtimes:
- Unresolved required slot at invocation → do NOT invoke; return a structured `dependency_unavailable` tool error naming the capability — same semantic class as the route perimeter's 503, so callers classify it as retryable topology, not application failure.
- Optional deps keep null-passthrough. Claim/route paths unchanged.
- Java: the check exists as `MeshToolWrapper.firstUnresolvedRequiredDependency()` — apply it on the direct-invoke path (`invokeInternal`/`buildFullArgs` branch). Python: `@mesh.tool` dispatch in the injection wrapper. TS: agent tool dispatch.

Follow-up candidate (not the contract fix): a small bounded pre-refusal wait to absorb sub-second recovery windows entirely.

## Tests
Per-runtime unit: required dep unresolved at direct invoke → structured refusal, handler not invoked; optional dep null → invoked (regression). Integration: DOWN→UP flap of a callee's required dep while a caller invokes through the window → refusal then success after convergence, never a null-dep handler error.

## Rider
Docs: `meshctl job status` reports `working`; align man-page wording that implies `running`.

dhyansraj:
Shipped in #1275 (merged to main; releases in v2.8.2 via #1276).

Issue #1268, opened 2026-07-04
fix(jobs): claim dispatch can invoke a handler with a null required dependency (gate/injection race)


dhyansraj:
## Contract

`meshctl man jobs` (claim-gating, #1258): "A claim worker never pulls a job for a capability whose `required` dependencies are currently unavailable … Claiming it would only run the handler far enough to fail on the missing dependency and burn a retry." Intent of `required=true` (#1249): a handler never observes null for a required dependency.

## Violation (field-verified on 2.8.0, reproducible)

The claim gate and the consumer's local injection run on **two independent clocks**: the registry-side gate (`capabilityUnavailableForClaim`, ent_service_jobs.go:606-642) flips available the moment the provider re-registers, while the consumer's injected proxy slot refills only when the consumer's own heartbeat response delivers the dependency event (`MeshEventProcessor.updateWrapperDep` → `MeshToolWrapper.updateDependency`). A required dep flapping DOWN→UP therefore opens a window where `/jobs/claim` grants a job whose handler args still carry a **null** required proxy:

- `MeshToolWrapper.buildFullArgs:842-848` passes null through ("graceful degradation") — no required check on the claim path.
- The settle-grace wait (`buildFullArgs:831-838`) is gated on the permanent one-way settle latch (`MeshSettleState.java:35`) — a no-op for post-boot flaps.
- The only required-enforcing perimeter is `@MeshRoute`-only (`MeshRouteHandlerInterceptor.java:163-219`).

Result observed: handler NPEs on the null proxy; the dispatcher's exception path posts a terminal `fail()` (correct per design — non-`retryOn` exceptions are developer-intent terminal), so a purely topological race permanently fails the job at `attempt_count 1 of max_retries 3`.

**All three runtimes share the gap**: Python (`dependency_injector.py:1502` — route-only required perimeter; `claim_dispatcher.py:276`), TypeScript, and the shared rust `claim_worker.rs` have no pre-claim local check and no pre-invoke required guard.

## Fix (defense in depth — both layers)

1. **Pre-claim local skip (primary)**: the claim worker skips claiming while any of the tool's `required` dependency slots is locally unresolved — the job stays queued with no owner and **no attempt increment**, exactly the #1258 posture; it self-heals on a later poll. Requires threading per-dep `required` metadata to the claim path (`MeshToolRegistry.DependencyInfo.required` exists; `MeshToolWrapper` currently receives names only and discards the flag).
[...]

PR #1275, opened 2026-07-04
fix(cli,di): group-scoped stop + required-dep guard on direct tool calls (2.8.2)

dhyansraj:
## Summary
The 2.8.2 patch — both fixes field-reported by the first 2.8.1 adopter.

- **#1274 — meshctl stop**: the graceful branch verified and escalated against the parent PID only; a Maven agent's child JVM survived stop, held its port (~3 min measured), and kept heartbeating. Now group-scoped end to end, with a **zombie-aware drain probe** (on Linux an unreaped zombie group answers `kill(-pgid,0)` with success — a container PID 1 that doesn't reap would hang the wait forever; macOS reaps instantly, which masked it locally). Watch reload routes through the same logic (regression caught by tc29 and fixed); `terminateAgent` polls ESRCH-only with a WARN when a group survives SIGKILL. Multi-name stop (`stop a b c`, existed since v1.1.0 but invisible behind the old group-teardown semantics) is now unit-tested and documented, along with stop's group semantics (+ setsid-escape caveat) and the by-design watch behaviors (content-write trigger; watch ≠ crash supervisor).
- **#1273 — required-dep guard on direct tools/call**: the #1268 pre-invoke check now covers the last unguarded dispatch path. Cross-runtime invariant: **direct calls refuse after the settle window** with a structured `{"error":"dependency_unavailable","capability":...}` tool error (byte-identical envelopes); **all job-flavored invocations release the lease** (claim + inbound job-header paths — never terminal, never a stranded row); **local/self-dependency dispatch raises** (a refusal can't masquerade as success to LLM/local callers). Handlers never observe null for `required=true` dependencies on any invocation path. Docs rider: job-status wording aligned to the real `working` vocabulary.

## Review Notes
Zero-context review applied: 1 BLOCKER (Java guard fired before the settle-grace wait — every agent restart would have burst refusals; now guards after settle with an armed-settle test) + 4 warnings (job-header paths on Java/TS needed release-not-refuse semantics; Python claim path routed refusals to terminal fail; Java local dispatch returned refusals as success) — all fixed. The watch-reload regression the gate caught was root-caused to the zombie-group probe gap above.

## Test plan
- [x] Unit: Java 75+, Python full sweep, TS 81 + tsc clean, Go cli/lifecycle suites (darwin + linux-container zombie tests)
- [x] Live darwin validation: real mvn parent/child topology — stop blocks till port release, SIGKILL escalation, multi-name stop
- [x] src-tests 12/12; tc29_watch_mixed green 3× (2 scoped + full run)
- [x] Full integration: 530 passed / 1 known-class LLM flake, green on sequential re-run

## Summary by CodeRabbit

* **New Features**
  * Direct tool calls now clearly refuse when a required dependency is unavailable, with a structured error message.
  * Watch mode behavior is now documented more precisely, including when rebuilds trigger and how restarts work.

* **Bug Fixes**
  * Stop and shutdown flows now wait for the whole process group to exit, improving reliability during graceful shutdown.
  * Jobs with temporarily unavailable dependencies are now re-queued instead of failing immediately.

* **Documentation**
  * Expanded guidance for stopping agents, watch mode, and upgrade procedures.

### DIFF
--- a/src/core/cli/lifecycle/group_zombie_darwin.go
+++ b/src/core/cli/lifecycle/group_zombie_darwin.go
@@ -1,0 +1,49 @@
+//go:build darwin
+
+package lifecycle
+
+import (
+	"os/exec"
+	"strconv"
+	"strings"
+)
+
+// groupAllZombie reports whether process group pgid is non-empty and EVERY
+// member is a zombie (state 'Z'). See the linux build's doc comment for the
+// full rationale (zombies hold no port/FDs; an unreaped orphan zombie in a
+// container must not hang the group-drain waits forever).
+//
+// Returns:
+//   - (true, nil)  at least one member was found and ALL members are zombies
+//   - (false, nil) a live (non-zombie) member exists, or the group is empty
+//   - (false, err) the ps scan itself failed — inconclusive; caller MUST NOT
+//     treat this as "dead"
+//
+// Uses `ps -A -o pgid=,stat=` (ps is part of the Darwin base system). Each line
+// is `<pgid> <stat>`; stat may carry flag suffixes (e.g. "Z+", "Ss"), so we key
+// on the first character.
+func groupAllZombie(pgid int) (bool, error) {
+	if pgid <= 1 {
+		return false, nil
+	}
+	out, err := exec.Command("ps", "-A", "-o", "pgid=,stat=").Output()
+	if err != nil {
+		return false, err
+	}
+	found := false
+	for _, line := range strings.Split(string(out), "\n") {
+		fields := strings.Fields(line)
+		if len(fields) < 2 {
+			continue
+		}
+		pg, err := strconv.Atoi(fields[0])
+		if err != nil || pg != pgid {
+			continue
+		}
+		found = true
+		if fields[1][0] != 'Z' {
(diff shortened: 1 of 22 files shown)


dhyansraj:
Review comments addressed in the follow-up commit — 7 of 8 fixed, 1 adjusted with evidence: the zombie-leader group-death shortcut removed (real-process test pins zombie-leader+live-child → escalation proceeds); stop docs corrected to the conditional refcount behavior; watcher restores currentCmd on kill failure (double-spawn hazard); **TS job-header detection was a real bug** — it read the allowlist-filtered map where the dispatch pair is deliberately absent, so genuine job dispatches would have been refused; now reads raw inbound headers with the test exercising real extraction; Java job-guard release path wrapped (open/releaseLease failures stay retryable, close always runs); python test settle reset added. The dependency_injector minimal-path item: reachable only for zero-injectable-param tools where a null dep is structurally unobservable — mirrored the route perimeter's documented choice (#1249) and added the INACTIVE-guard WARN instead of refusing, keeping cross-perimeter consistency. All per-runtime suites green (Java 40, Python, TS 40 + tsc, Go cli/lifecycle on darwin+linux).

Issue #1116, opened 2026-06-02
TypeScript runtime: consolidate MCP-call/LLM duplication & shared event-loop


dhyansraj:
Duplication and drift in the TS runtime — notably MCP-call setup that is *already diverging* between the buffered and streaming paths. Source: multi-agent code audit (2026-06-01); adversarially verified.

## MCP call / proxy
- **`callMcpTool` / `streamMcpTool` duplicate ~120 lines of request setup** (header allowlist, trace injection, timeout) — `proxy.ts:233-310` & `477-619`. **Already drifting**: stream added a `effectiveTimeout<=0` guard the buffered path lacks. *Fix: extract `buildMcpRequest` + `wireJobCancel`.*
- **`MeshDelegatedProvider` hand-rolls the MCP `tools/call` HTTP/SSE/span path** that `callMcpTool` already provides (and its own `streamComplete` already reuses) — `llm-agent.ts:160-390`/`1153-1307`. Misses retry, max-size, per-job cancel. *Fix: route through `callMcpTool`/`streamMcpTool`; centralize span publishing in `publishProxySpan`.*

## LLM agent
- **`MeshLlmAgent.run()`/`stream()` duplicate ~90 lines of message assembly**; `complete()`/`streamComplete()` duplicate modelParams/request building — `llm-agent.ts:534-643`/`811-917`/`176-216`/`419-451`. *Fix: `buildAgentMessages` + `buildMeshLlmRequest` helpers.*
- **Provider-managed agentic loop hard-codes `maxIterations=10`** while the consumer side is configurable; `maxIterations` is never forwarded in `model_params` — `llm-provider.ts:866-871`, `llm-agent.ts:617-621`. *Fix: read cap from model_params/`MESH_LLM_MAX_ITERATIONS`; surface a truncation marker.* (Relates to the cross-runtime parity issue.)
- **Output schema computed twice per run** via `zodToJsonSchema` — `llm-agent.ts:632-643`/`1087-1098`/`560-565`. *Fix: compute once and cache on the agent (`returnSchema` is immutable).* `[performance, low]`
- **`createCallable` duplicates run-vs-stream context-merge** — `llm-agent.ts:932-994`. *Fix: local `mergeRunContext(options)` closure.*
[...]

Issue #1355, opened 2026-07-21
LLM iteration-cap exhaustion is indistinguishable from a successful response (all runtimes)


dhyansraj:
## Summary

When an LLM agentic loop exhausts its `max_iterations` cap, most code paths emit something **indistinguishable from a successful model response**. A caller cannot tell a truncated loop from a completed one without exact-matching an English string — and on two paths, not even that.

A cap being hit is a failure. It should surface as a typed, catchable error, not as content occupying the same channel and shape as a model answer.

## Current behavior, by path

| # | Location | On exhaustion |
|---|---|---|
| P1 | `python/_mcp_mesh/engine/mesh_llm_agent.py:1141-1148` | **raises** `MaxIterationsError` ✅ |
| P2 | `python/mesh/helpers.py:1870-1878` | returns `{"role": "assistant", "content": "Maximum tool call iterations reached"}` |
| P3 | `python/mesh/helpers.py:2301-2307` | yields the bare string `"Maximum tool call iterations reached"` into the same `AsyncIterator[str]` as real tokens |
| T1 | `typescript/src/llm-agent.ts:794-797` | **throws** `MaxIterationsError` ✅ |
| T2 | `typescript/src/llm-provider.ts:1367-1375` | assigns `{ role: "assistant", content: "Maximum tool call iterations reached" }` |
| T3 | `typescript/src/llm-provider.ts:1008-1014` | Gemini path hands the cap to the AI SDK as `stopWhen: stepCountIs(...)` — **no exhaustion handling at all**; silent truncation, no marker |
| J1 | `java/.../MeshLlmAgentProxy.java:926-927` | logs a warning and returns `extractLastAssistantContent(llmMessages)` — **prior model text, no marker whatsoever** |
| J2 | `java/.../MeshLlmProviderProcessor.java:520-524` | synthetic message, plus `usageMeta` copied from the last response — making it look *more* genuine than the Python one |

Only P1 and T1 do the right thing, and they are on the path a delegating consumer does not take (see below).

## The synthetic message carries no machine-readable marker

- Python buffered (`helpers.py:1874-1878`): plain `{"role", "content"}`. No `stop_reason`, no sentinel, no metadata.
- Python stream (`helpers.py:2307`): a bare `str` in the token stream. The comment at `:2301-2302` states the intent — "Emit a textual indicator so consumers iterating `async for` always see at least one chunk."
[...]

PR #1359, opened 2026-07-21
fix(llm): forward the consumer's max_iterations to the provider-managed loop

### LINKED ISSUE #1356: python: consumer max_iterations is not forwarded to the provider-managed loop (hardcoded 10)
## Summary

In the Python runtime, a consumer's `max_iterations` setting is **inert on the path it actually takes**. The provider-managed loop hardcodes `10` and never reads the consumer's value, while the consumer-side loop that *does* honour it is unreachable in the delegating configuration.

Found while auditing #1355; distinct bug, filed separately.

## Detail

The provider-managed loop is invoked with a literal:

- `python/mesh/helpers.py:2829` — `max_iterations=10` (buffered)
- `python/mesh/helpers.py:3171` — `max_iterations=10` (stream)

The consumer's configured value never reaches the provider. It flows into `llm_config.py:30` and governs only the consumer-side loop at `mesh_llm_agent.py:968`, and the `model_params` construction at `mesh_llm_agent.py:990-1000` does not include it — so it is not on the wire at all.

Combined with the delegation behavior described in #1355 (the provider returns a terminal message with no `tool_calls`, so the consumer's loop exits on iteration 1), the net effect for a `@mesh.llm(provider={...}, filter={...})` consumer is:

- the value they set does nothing
- the effective cap is a hardcoded 10 on the provider side
- there is no error or warning indicating their setting was ignored

## Cross-runtime comparison

- **TypeScript forwards it correctly** — `llm-agent.ts:224` → `llm-provider.ts:702-705` (`resolveMaxIterations`).
- **Java's consumer proxy honours its own value** (`MeshLlmAgentProxy.java:835`), but **`MeshLlmProviderProcessor.java:442` hardcodes `int maxIterations = 10;`** — same defect as Python on the provider side.

So TypeScript is the reference implementation here; Python and Java's provider paths both need to accept and apply the caller's value.

## Suggested fix

Thread `max_iterations` through the delegation request the way TypeScript does, and have the provider-managed loop use the caller's value with the current `10` as the default when absent. Sanitize it on the provider side (positive integer, sane upper bound) since it now crosses a trust boundary.

Fixing this alongside #1355 makes sense — both concern the same loop and the same wire hop, and the exhaustion signal added there gives the caller a way to observe that a cap was hit at all.

dhyansraj:
A consumer's `max_iterations` was **inert on the path it actually takes**. The Python provider-managed loop hardcoded `10` (`helpers.py:2829`, `:3171`) and the consumer never put the value on the wire — so a `@mesh.llm(provider={...})` consumer setting it got no error, no warning, and no effect. Java hardcoded `10` the same way (`MeshLlmProviderProcessor.java:442`) and its consumer never forwarded the key either.

## Contract

Wire key `model_params.max_iterations`, resolved provider-side with the semantics TypeScript already ships (#1116):

```
sanitize(v)     = floor(Number(v)) if finite and > 0 else unset
resolve(param)  = param present ? (sanitize(param) ?? 10)
                                : (sanitize($MESH_LLM_MAX_ITERATIONS) ?? 10)
```

Including the subtlety that a **present-but-invalid** value falls back to 10 rather than to the environment.

The key is **stripped provider-side** so it can never leak into a vendor API request. Two tests cover this: the loop path, and the legacy single-call path where `completion_args.update(model_params_copy)` would otherwise hand it straight to LiteLLM.

## Precedence decision

Forwarding is **explicit-only**, so a provider operator's `MESH_LLM_MAX_ITERATIONS` still applies to consumers that configured nothing.

That required making "unset" representable:
- **Python** — decorator default moves `10` → `None`. `LLMConfig` exposes `effective_max_iterations` (`None` → 10) and `max_iterations_explicit`, so nothing downstream ever sees `None`. `get_config_value` semantics are untouched (other call sites depend on them); a consumer-side env var still counts as explicit, since only "nothing configured anywhere" yields `None`.
- **Java** — new `MeshLlmDefaults.MAX_ITERATIONS_UNSET = 0`, matching the existing `MAX_TOKENS_UNSET` pattern. `@MeshLlm(maxIterations)` defaults to it: source- and binary-compatible, and behavior-preserving locally since `0` already normalized to `10`.

⚠️ **Release note:** `@MeshLlm(maxIterations)`'s annotation default changes from `10` to the unset sentinel. No behavioral change, but it is a visible default.

## Known remaining gap — TypeScript, deliberately not changed here

The issue described TypeScript as the correct reference. **It isn't.** Its guard looks explicit-only:

```ts
if (options?.maxIterations !== undefined) { modelParams.max_iterations = options.maxIterations; }
```

but `llm.ts:234` defaults `config.maxIterations ?? 10`, and `llm-agent.ts:674-677,712` passes the **resolved** value into the very `options` the guard tests — so the guard always fires. A TS consumer that configures nothing still sends `max_iterations: 10`, leaving the provider's env inert for it.

Python and Java now behave correctly; TS keeps the old behavior. The sanitize/resolve semantics are byte-identical across all three, so only the omit-when-unset case differs. Tracked separately rather than widened into this PR.

## Also found

- **Consumer-side env parsing is stricter than the wire path in Python.** `MESH_LLM_MAX_ITERATIONS=3.5` on a consumer is rejected by `NONZERO_RULE` (→ unset), while the same value forwarded on the wire floors to `3`. Pre-existing `get_config_value` behavior, left alone rather than widening shared semantics.
- **Stale docs corrected**: `docs/python/llm/index.md` and `docs/java/annotations.md` both documented the default as `1`; now `10` plus the forwarding note. No `meshctl man` page mentions `max_iterations`, so no parallel man edit.

## Tests

- **Python** (new, 39 tests): sanitize table (zero/negative/NaN/inf/non-numeric/fractional/empty/None/collections), resolve (valid, invalid-explicit→10 *with env set*, absent→env, absent+invalid env→10, absent+no env→10), `LLMConfig` unset/explicit/explicit-10/zero-rejected, consumer forward-vs-omit on **both** buffered and stream paths, decorator resolution, and the two leak tests.
- **Java** (new, 13 tests): provider sanitize/resolve + key-stripping; consumer explicit-forwarded / unset-omitted / unset-still-10-locally / escape-hatch-wins.
- One pre-existing test updated (`test_mesh_llm_decorator.py:123`) — it encoded the old default.

## Verification

Run independently, not just reported:
- Python CI-equivalent (`pytest tests/unit/ _mcp_mesh/`): **0 failures** (2419 passed, 2 pre-existing live-integration skips)
- Java (`mvn -pl mcp-mesh-spring-ai,mcp-mesh-spring-boot-starter -am test`): **BUILD SUCCESS**, exit 0 — spring-ai 207, starter 661, 0 failures/errors
- TypeScript: 83 files, 1179 tests pass (unchanged)

Closes #1356

https://claude.ai/code/session_01G1cLws2dhLUhhVdGHJ5aXQ

## Summary by CodeRabbit

* **New Features**
  * LLM iteration limits can now be left unset, allowing the provider’s configured limit or default to apply.
  * Explicit iteration limits are forwarded to provider-managed agent loops across Python and Java runtimes.
  * Environment-based iteration limits are supported when no explicit limit is configured.
  * Invalid iteration-limit values are safely rejected or replaced with defaults.

* **Documentation**
  * Updated Java and Python documentation to describe unset behavior, provider delegation, and defaults.

* **Tests**
  * Added coverage for forwarding, provider configuration, registration, and cross-runtime behavior.

### DIFF
--- a/docs/java/annotations.md
+++ b/docs/java/annotations.md
@@ -200,7 +200,7 @@ public AnalysisResult analyze(
 | Attribute          | Required | Default | Description                          |
 | ------------------ | -------- | ------- | ------------------------------------ |
 | `providerSelector` | Yes      |         | `@Selector` for mesh LLM discovery   |
-| `maxIterations`    | No       | `1`     | Max agentic loop iterations          |
+| `maxIterations`    | No       | `10`    | Max agentic loop iterations (when set explicitly it is also forwarded to the provider-managed loop) |
 | `systemPrompt`     | No       | `""`    | System prompt or template path       |
 | `contextParam`     | No       | `""`    | Parameter name for template context  |
 | `filter`           | No       |         | `@Selector` for tool filtering       |
--- a/docs/python/llm/index.md
+++ b/docs/python/llm/index.md
@@ -57,7 +57,7 @@ def assist(ctx: AssistContext, llm: mesh.MeshLlmAgent = None) -> AssistResponse:
 | Parameter        | Type | Description                                       |
 | ---------------- | ---- | ------------------------------------------------- |
 | `provider`       | dict | LLM provider selector (capability + tags)         |
-| `max_iterations` | int  | Max agentic loop iterations (default: 1)          |
+| `max_iterations` | int  | Max agentic loop iterations (default: 10; when set explicitly it is also forwarded to the provider-managed loop) |
 | `system_prompt`  | str  | Inline prompt or `file://path` to Jinja2 template |
 | `response_model` | type | Pydantic model the LLM must emit and is validated against (separate from the return type) |
 | `context_param`  | str  | Parameter name receiving context object           |
--- a/src/runtime/java/mcp-mesh-sdk/src/main/java/io/mcpmesh/MeshLlm.java
+++ b/src/runtime/java/mcp-mesh-sdk/src/main/java/io/mcpmesh/MeshLlm.java
@@ -89,8 +89,14 @@ public @interface MeshLlm {
      * Maximum iterations for the agentic loop.
      *
      * <p>Each iteration allows the LLM to make tool calls.
+     *
+     * <p>Unset by default ({@link MeshLlmDefaults#MAX_ITERATIONS_UNSET}), which
+     * means "10, and let the provider decide its own cap". When set explicitly
+     * — here or via {@code MESH_LLM_MAX_ITERATIONS} — the cap is ALSO forwarded
+     * to the provider on the wire ({@code model_params.max_iterations}) so the
+     * provider-managed loop honors it (issue #1356).
      */
-    int maxIterations() default MeshLlmDefaults.MAX_ITERATIONS;
+    int maxIterations() default MeshLlmDefaults.MAX_ITERATIONS_UNSET;
 
     /**
      * System prompt for the LLM.
--- a/src/runtime/java/mcp-mesh-sdk/src/main/java/io/mcpmesh/MeshLlmDefaults.java
+++ b/src/runtime/java/mcp-mesh-sdk/src/main/java/io/mcpmesh/MeshLlmDefaults.java
@@ -22,6 +22,15 @@ public final class MeshLlmDefaults {
     public static final int MAX_ITERATIONS = 10;
 
     /**
+     * Sentinel for an unset {@code maxIterations} (issue #1356). When
+     * {@code maxIterations} equals this value the consumer's own loop uses
+     * {@link #MAX_ITERATIONS} but no {@code max_iterations} is sent on the
+     * wire, so the provider-managed loop applies its own
+     * {@code MESH_LLM_MAX_ITERATIONS} (or the same default of 10).
+     */
+    public static final int MAX_ITERATIONS_UNSET = 0;
+
+    /**
      * Sentinel for an unset {@code maxTokens}. When {@code maxTokens} equals
      * this value, no {@code max_tokens} is sent on the wire and the provider's
      * own default applies.
--- a/src/runtime/java/mcp-mesh-spring-ai/src/main/java/io/mcpmesh/ai/MeshLlmProviderProcessor.java
+++ b/src/runtime/java/mcp-mesh-spring-ai/src/main/java/io/mcpmesh/ai/MeshLlmProviderProcessor.java
@@ -2,6 +2,7 @@ package io.mcpmesh.ai;
 
 import tools.jackson.core.JacksonException;
 import tools.jackson.databind.ObjectMapper;
+import io.mcpmesh.MeshLlmDefaults;
 import io.mcpmesh.MeshLlmProvider;
 import io.mcpmesh.core.MeshObjectMappers;
 import io.mcpmesh.ai.handlers.LlmProviderHandler;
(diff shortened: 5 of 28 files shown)


dhyansraj:
## Integration testing caught a regression in this PR — fixed in 4663379e7

The unit tests were green and the change looked complete. The new `tc46` case, written specifically to exercise "consumer configures nothing", found that **every Python `@mesh.llm` consumer without an explicit cap failed to register**:

```
ERROR Rust heartbeat failed for agent '...': argument 'max_iterations':
      'NoneType' object cannot be interpreted as an integer
```

### Mechanism

Three things had to line up, and this branch changed the first:

1. `decorators.py:3028` **always** emits the `"max_iterations"` key — its value became `None` when nothing configured it.
2. `rust_heartbeat.py:511` read `config.get("max_iterations", 1)` — `dict.get` only substitutes the default when the key is **absent**, so present-with-`None` passed straight through.
3. `core.LlmAgentSpec` types the field as a non-optional `u32` (`spec.rs:283-296`) and rejects `None`.

### Why it was dangerous

- **Silent and unrecoverable.** The agent started, served HTTP, logged two ERROR lines, then never registered and **never retried**, while `meshctl start` exited 0 reporting success.
- **It hit the canonical documented form.** `@mesh.llm(provider={...})` with no cap is exactly what `meshctl scaffold` prints as its usage hint (`llm_subcommand.go:393`), and ~19 shipped example agents use it.
- **Unit tests could not see it** — they covered config, forwarding and resolution, never the registration path.
- **The existing integration suite nearly masked it too**: every pre-existing uc04 Python consumer happens to set `max_iterations` explicitly. Only a test written for the unset case entered the broken path.

### Fix

Register the **effective** cap (10 when unset) — precisely what the spec carried before this branch. Note the old fallback was `1`, not `10`, so a naive `.get(..., 1)` patch would have silently registered a bogus cap. The explicit/unset tri-state stays confined to consumer→provider forwarding; the registration spec has always carried a concrete number and still does.

The regression test drives the real `_build_agent_spec` against the **real Rust `LlmAgentSpec`** — a stubbed spec class would have accepted `None` and proved nothing. Verified failing before the fix, passing after.

Java and TypeScript verified unaffected rather than assumed: Java never sends LLM specs at all (`AgentSpec.LlmAgentSpec` is unreferenced outside its own file), and TypeScript coerces to a concrete number at registration (`llm.ts:234`).

## Test results

**Regression gate — existing cases whose declared caps this PR makes real for the first time.** All pass, so those flows genuinely fit inside their declared caps and no adjustment is needed:

| case | cap was (inert) | now enforced | result |
|---|---|---|---|
| `tc37` gemini3 multiturn tools loop stress | 10 | 8 | ✅ |
| `tc36` gemini3 response_json_schema + tools | 10 | 6 | ✅ |
| `tc19` avatar structured | 10 | 1 | ✅ |

**New coverage** (`3 passed, 0 failed`, against an image rebuilt from this branch):

| case | proves |
|---|---|
| `tc45_max_iterations_explicit_cap_py` | explicit cap honored end-to-end |
| `tc46_max_iterations_provider_env_py` | unset consumer → provider's `MESH_LLM_MAX_ITERATIONS` applies |
| `tc47_max_iterations_cross_runtime_java_py` | Java consumer → Python provider |

`tc46` discriminates against both failure modes: the old behavior (provider hardcodes 10) and a naive "always forward 10" fix both yield 4 tool invocations. Only correct explicit-only forwarding yields 1.

The cases assert on a **tool-invocation counter**, not the provider's `"Maximum tool call iterations reached"` string, so #1355 can restructure that signal without breaking them. Determinism comes from a forced data dependency — each call returns a random token required by the next — so the model cannot batch several calls into one loop iteration and corrupt the count.

Python suite green; Java and TypeScript unchanged and green.

## Follow-up worth its own issue

A spec-build failure is a **configuration** error, not a transient one, yet it is swallowed into a background heartbeat log while the process stays up looking healthy. That deserves a hard startup refusal rather than two ERROR lines and an agent that never joins the mesh.

PR #1367, opened 2026-07-22
fix(llm): structural signal for max_iterations exhaustion + drop dead core loop (Python)

### LINKED ISSUE #1355: LLM iteration-cap exhaustion is indistinguishable from a successful response (all runtimes)

dhyansraj:
## Summary

**Part of #1355** — LLM agentic-loop `max_iterations` exhaustion was indistinguishable from a successful model response. A caller could only tell a truncated loop from a completed one by exact-matching an English string (and on some paths, not even that). This PR lands the **Python reference implementation** plus a **core cleanup**; TypeScript and Java parity follow as Phases 2/3 in separate PRs.

### Phase 0 — delete the dead core `agentic_loop` module
- Removed `core/src/agentic_loop.rs` (an unused discriminated-action state machine) and its PyO3 (`lib.rs`), C-FFI (`ffi.rs`), NAPI (`napi.rs`) exports, the regenerated cbindgen header entries, and the `mcp_mesh_core/__init__.py` re-exports.
- Zero callers outside the core crate — all three runtimes hand-roll their loops, and the structurally-correct-but-unused implementation is exactly what let the runtimes drift. **No behavior change** (−1166 lines).

### Phase 1 — Python: structurally distinguishable exhaustion
Honors the #1247 envelope contract (`content` is always the answer string), so the discriminant is a **namespaced sibling field**, never inside `content`.
- **New `_mcp_mesh/engine/llm_stop_reason.py`** — `_mesh_stop_reason` / `"max_iterations"` vocabulary + stream terminal-frame encode/parse.
- **Provider (`mesh/helpers.py`)** — buffered exhaustion sets `content` to the last real assistant text (or `""`) plus sibling `_mesh_stop_reason: "max_iterations"` (was: the English marker in `content`); streaming emits a reserved terminal control frame instead of yielding the bare string. Field **omitted on success**.
- **Consumer (`mesh_llm_agent.py`)** — captures the field, adds it to the bare-map reserved-key guard, and **raises `MaxIterationsError`** on the mainline (buffered path, primary stream path, and the degraded stream buffered-fallback). This makes the documented `types.py:460` contract true for the provider-managed-loop path, which was previously **unreachable** (the managed loop returns a terminal no-tool-calls message, so the consumer used to exit normally on iteration 1).
- **Exports** — `MaxIterationsError` and `ToolExecutionError` are now importable from `mesh`.

### Streaming mechanism
The provider stream channel is strictly `str`-typed (FastMCP progress notifications carry string chunks), so a structured object can't ride it. The provider emits a **reserved-prefixed sentinel JSON string** (`{"_mesh_stream_end": {"stop_reason": "max_iterations"}}`) as the final chunk; the consumer wrapper intercepts it via a cheap prefix check → `json.loads`, never forwards it, and raises. A real text token can't collide (would have to be that exact JSON object).

## Review notes

- Independent diff review: **0 blockers, 2 warnings, 3 info**.
- **W1 (fixed in this PR):** the degraded stream buffered-fallback path didn't check the exhaustion field and would have produced a silent empty stream; it now raises `MaxIterationsError`, with a new test covering it.
- **W2 (release sequencing — no code fix):** this changes the wire shape a Python provider emits on exhaustion. A Python provider paired with an **old** (pre-#1355) TS/Java consumer from a different release would show empty content (buffered) or forward the raw control frame into the token stream (streaming). Same-version and Python↔Python are fine, and an updated consumer degrades gracefully when the field is absent. **Ship the #1355 phases within the same release** (or land consumer tolerance before/with provider emission) so no cross-version skew hits users. Worth a release-note callout when this ships.
- Infos folded/accepted: `parse_stream_end` now debug-logs an unrecognized `stop_reason` instead of silently dropping it; `iteration_count` on the new path is the cap value (the consumer doesn't see the provider's internal counter — matches the design); `_mesh_usage` omission on exhaustion is pre-existing.

## Verification

- `cargo check` passes for both the `python` and `typescript,simd` and `ffi` feature sets; `cargo test` 473 pass after the core deletion.
- Python: targeted unit set (the two updated tests + iteration-isolation kept green + new signal/fallback tests) — 34 pass; broader LLM/provider sweep — 261 pass.
- `from mesh import MaxIterationsError, ToolExecutionError` verified.

## Test plan

- [ ] CI green (Rust all feature sets + Python runtime tests)
- [ ] Confirm no lingering reference to the deleted `agentic_loop` bindings in any runtime
- [ ] Phase 2 (TypeScript, incl. the Gemini `stopWhen` path) — separate PR
- [ ] Phase 3 (Java, incl. the silent `J1` return → throw) — separate PR, will `Close #1355`
- [ ] Release-note callout for the exhaustion wire-signal change (sequence all runtimes in one release)

## Summary by CodeRabbit

* **New Features**
  * Added structured max-iteration exhaustion signaling for both buffered and streaming agent interactions.
  * Preserves the most recent meaningful assistant text when iteration limits are reached.
  * Exposes `MaxIterationsError` and `ToolExecutionError` through the public Python interface.
* **Breaking Changes**
  * Removed legacy agentic-loop APIs from supported language bindings.
* **Bug Fixes**
  * Standardized typed streaming end-of-stream frames and ensured termination markers don’t leak into user-visible text.
* **Tests**
  * Added/updated unit tests to verify max-iteration propagation and typed frame behavior, including defensive streaming parsing.


dhyansraj:
Revised the **streaming** transport per review (commit `08e58fde0`).

The original streaming terminal signal was a reserved-prefixed sentinel string detected in-band, which can theoretically collide with model output. It is replaced with a **typed stream envelope** aligned with the Anthropic Messages API streaming discipline (the discriminator is the frame TYPE, never the text content). Every stream chunk is now a frame keyed by a reserved `_mesh_frame` discriminator:

- text delta → `{"_mesh_frame": "chunk", "content": "<delta>"}`
- terminal, normal → `{"_mesh_frame": "end"}`
- terminal, exhaustion → `{"_mesh_frame": "end", "stop_reason": "max_iterations"}`

Consumer unwraps: `chunk` → yield content; `end` + `max_iterations` → raise `MaxIterationsError` (not yielded); plain `end` → terminate; non-frame chunk → defensive raw-text passthrough.

The reserved `_mesh_frame` key closes **both** collision surfaces: a framed chunk whose content is literally the end-frame JSON is yielded as text, and an unframed raw delta that merely looks frame-shaped (`{"type":"end"}`) fails the reserved-key check and passes through as text. TS/Java streaming (Phases 2/3) adopt the same frame.

The **buffered** path (`_mesh_stop_reason` sibling field) is unchanged. Independent review of the revision: 0 blockers; the one warning (wider unframed-collision surface) and two infos were fixed in this push.

Issue #1383, opened 2026-07-27
Python: LiteLLM is not required for big-3 vendors but is unavoidable; make it an extra and fail loudly when missing


dhyansraj:
## Summary

`pip install mcp-mesh` installs **110 packages**, including the full LiteLLM tree, for every user — even those who only use `@mesh.tool` DI or one of the big-3 vendors. LiteLLM is **not functionally required for Anthropic / OpenAI / Gemini**: those route through bundled native SDK adapters that are default-ON. It is currently unavoidable only because a handful of `import litellm` statements sit at the *top* of the provider entry points and execute before native dispatch is consulted.

Separately, when LiteLLM is genuinely missing the failure is a bare `ModuleNotFoundError` raised at the **first LLM call** — after install, import, and decoration all succeed — so it escapes build and deploy entirely and surfaces only in production.

## Evidence

Measured against 3.3.1 on linux/amd64.

**Dependency weight** — `pip install mcp-mesh==3.3.1` resolves 110 packages. The LLM subset:

| wheel | size |
|---|---|
| litellm | 20.2 MB |
| hf-xet | 4.4 MB |
| tokenizers | 3.3 MB |
| tiktoken | 1.1 MB |
| regex | 0.8 MB |
| openai / google-genai / anthropic | ~3.6 MB combined |

`litellm` and its transitive tree account for ~30 MB. The three native SDKs total ~3.6 MB.

**LiteLLM is import-only ballast on the big-3 path.** In an env with `anthropic` installed and `litellm` absent:

- `_infer_big3_vendor_from_bare_name()` resolves bare names with no LiteLLM (`claude-*`→anthropic, `gpt-*`/`o1*`→openai, `gemini-*`→gemini)
- `anthropic_native.is_available()` → `True`; `ClaudeHandler.has_native()` → `True`
- native dispatch is **default-ON** (`MCP_MESH_NATIVE_LLM=0` is the opt-out)
- but the call still fails: `ModuleNotFoundError: No module named 'litellm'`

Substituting a **hollow `litellm` stub** (an empty module, zero functionality) lets the same call run end-to-end and reach the Anthropic API (returns a genuine `401 invalid x-api-key` with a `request_id`). LiteLLM is imported but never invoked on this path.

[...]

PR #1386, opened 2026-07-27
fix(llm): make LiteLLM optional for big-3 vendors and fail loudly when missing

dhyansraj:
## Summary

Implements items 1 and 2 of #1383. **Neither is breaking** — item 1 only improves an error message, item 2 makes the big-3 vendors work *without* LiteLLM. The breaking change (moving `litellm` out of the base install behind an extra) is item 3 and stays on a version boundary; this PR does not touch `pyproject.toml`.

LiteLLM is not functionally required for Anthropic / OpenAI / Gemini — they dispatch through the bundled native SDK adapters, and native dispatch is default-ON (`MCP_MESH_NATIVE_LLM=0` is the opt-out). It was unavoidable only because several `import litellm` statements sat at **function-top** in the provider entry points and executed before native dispatch was ever consulted.

Separately, when LiteLLM is genuinely missing, the failure was a bare `ModuleNotFoundError` raised at the **first LLM call** — after install, import, and decoration all succeed. That escapes build and deploy entirely and surfaces only in production.

## Changes

All in `src/runtime/python/mesh/helpers.py`.

- **New `_require_litellm(model=None, vendor=None)`** — imports at the point of genuine use; on failure raises `ImportError` naming the model and resolved vendor, the `pip install 'mcp-mesh[litellm]==<version>'` command, and the `requirements.txt` guidance for containerized agents, chained from the original error via `raise ... from e`. `ImportError` is what the `llm_provider` docstring already declared, so no new public exception type was introduced.
- **Six function-top imports relocated** to the LiteLLM fallback branches (~545, ~1512, ~2047, ~2899, ~3285, plus `_maybe_retry_synthetic_on_validation_failure`, which took the module as a required kwarg and was a transitive blocker).
- `_dispatch_completion` and `_maybe_retry_synthetic_on_validation_failure` now resolve the module lazily; both are private with no external callers.
- **The decorator's vendor detection is deliberately unchanged.** It stays soft-fail — `try/except (ImportError, ...)` falling through to `_extract_vendor_from_model` / `_infer_big3_vendor_from_bare_name` — because that path has a correct non-LiteLLM fallback and must not start raising.

No dispatch logic or control flow was altered; only where the name is bound.

## Verification

- **Full unit suite: 1774 passed** (`tests/unit/` + `provider_handlers/tests/`), 0 failures.
- **Big-3 without LiteLLM.** In a venv with `anthropic` installed and `litellm` uninstalled (precondition asserted via `importlib.util.find_spec`), a bare-name `claude-sonnet-4-5` provider dispatches native and reaches the Anthropic API — 401 on a bogus key, and a normal completion with a real one. Buffered and streaming paths both verified. Pre-fix this raised `ModuleNotFoundError` at function entry.
- **Long-tail vendor.** `cohere-command` and `llama-3` now produce the actionable `ImportError` on both paths instead of a bare `ModuleNotFoundError`.
- Bare-name vendor inference still resolves with LiteLLM absent: logs `using 'unknown'` then `Inferred big-3 vendor 'anthropic' ... normalized to 'anthropic/claude-sonnet-4-5'`.

Refs #1383

## Test plan

- [ ] CI green
- [ ] Manual: an agent using only a big-3 vendor functions with `litellm` uninstalled
- [ ] Manual: a long-tail vendor without `litellm` surfaces the guidance, not a bare import error

## Summary by CodeRabbit

* **Bug Fixes**
  * LiteLLM is now loaded only when a non-native (“long-tail”) execution path is used, so native provider workflows run even when it isn’t installed.
  * Improved ImportError messaging explains when LiteLLM is required and how to proceed.
  * Retry/fallback and both buffered and streaming completion flows now route providers more reliably.
* **Documentation**
  * Updated provider guidance to reflect LiteLLM’s optional, long-tail behavior and improved error expectations.

### DIFF
--- a/src/runtime/python/mesh/helpers.py
+++ b/src/runtime/python/mesh/helpers.py
@@ -39,6 +39,58 @@ logger = logging.getLogger(__name__)
 _MAX_ITERATIONS_UNSET = object()
 
 
+def _require_litellm(model: str | None = None, vendor: str | None = None) -> Any:
+    """Import and return the ``litellm`` module, or raise actionable guidance.
+
+    LiteLLM is an OPTIONAL dependency (issue #1383). The big-3 vendors
+    (Anthropic / OpenAI / Gemini) dispatch through the bundled native SDK
+    adapters and never reach this helper; only the long-tail
+    GenericHandler/LiteLLM path genuinely needs the package. Call this at the
+    point of genuine use — never at function top — so a big-3-only install
+    stays fully functional without litellm present.
+    """
+    try:
+        import litellm
+
+        return litellm
+    except ModuleNotFoundError as e:
+        # Only the top-level 'litellm' package being absent is an optional-dep
+        # problem. A failing transitive import (e.g. 'tokenizers') or any other
+        # ImportError (ABI mismatch, circular import) must propagate untouched —
+        # mislabelling those sends users to an install command that is already
+        # satisfied and cannot fix anything, while hiding the real cause.
+        if e.name != "litellm":
+            raise
+
+        try:
+            from mesh import __version__ as mesh_version
+        except Exception:  # pragma: no cover - version lookup is best-effort
+            mesh_version = None
+
+        pin = (
+            f"mcp-mesh[litellm]=={mesh_version}"
+            if mesh_version
+            else "mcp-mesh[litellm]"
+        )
+        target = ""
+        if model and vendor and vendor != "unknown":
+            target = f" for model '{model}' (vendor '{vendor}')"
+        elif model:
+            target = f" for model '{model}'"
+        elif vendor and vendor != "unknown":
+            target = f" for vendor '{vendor}'"
+
+        raise ImportError(
+            f"This request{target} requires the optional LiteLLM provider path, "
+            "but the 'litellm' package is not installed. mcp-mesh bundles native "
+            "SDK adapters for Anthropic, OpenAI and Gemini; every other "
+            "vendor/model dispatches through LiteLLM.\n"
+            f"  Install it with:  pip install '{pin}'\n"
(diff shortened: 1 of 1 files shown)


dhyansraj:
### Follow-up: narrowed the import handler (review finding)

`_require_litellm` caught `except ImportError` broadly, which mislabelled two distinct failures as "litellm is not installed":

- litellm **is** installed but one of its own transitive imports fails (e.g. `tokenizers`) — the user was told to `pip install mcp-mesh[litellm]`, already satisfied and unable to help, while the real missing module was hidden.
- A non-missing-module `ImportError` (ABI mismatch, circular import) — not a packaging problem at all, reported as one.

Now narrowed to `except ModuleNotFoundError` with `if e.name != "litellm": raise`. Bare `raise` re-raises the same exception object with its original traceback and no added `__cause__`. Non-`ModuleNotFoundError` `ImportError`s bypass the handler entirely. `e.name is None` propagates too, since `None != "litellm"` — failing open toward the real error rather than the guidance.

Behavior for the genuine missing-package case is unchanged, so callers are unaffected (`ModuleNotFoundError` is a subclass of the `ImportError` the `llm_provider` docstring declares).

Verified all paths by patching `builtins.__import__`:

| case | result |
|---|---|
| `name="litellm"` | custom guidance, chained `from e`, pin resolved live to `mcp-mesh[litellm]==3.3.1` |
| `name="tokenizers"` | original object propagates, `__cause__=None`, no `mcp-mesh[litellm]` mention |
| `ImportError("cannot import name ...")` | propagates unchanged, no mention |
| `name=None` | propagates unchanged, no mention |

Full CI-equivalent suite (`tests/unit/ _mcp_mesh/`): **2434 passed, 0 failed, 2 skipped** (the 2 are pre-existing opt-in live-integration gates).

Issue #1396, opened 2026-07-28
@mesh.route handlers mounted via include_router are not discovered on FastAPI >= 0.139


dhyansraj:
## Summary

`@mesh.route` handlers registered on an `APIRouter` and mounted with `app.include_router()` are **not discovered at all** on FastAPI >= 0.139.x. Dependency injection never happens for them; the routes serve, but as plain FastAPI endpoints with no mesh wiring.

This is a documented, exemplified pattern — `examples/simple/simple_fastapi_router.py` exists specifically to demonstrate it, and it is covered in `docs/python/fastapi-integration.md`.

## Cause

Around 0.139.x, `include_router()` stopped flattening the router's `APIRoute` objects into `app.router.routes`. It now inserts a single `_IncludedRouter` entry that has no `.routes` attribute.

Mesh route discovery walks `app.router.routes` directly (`_mcp_mesh/shared/server_discovery.py:302`), so the mounted handlers are invisible to it.

Reproduced:

```python
from fastapi import FastAPI, APIRouter
r = APIRouter()

@r.get("/time")
async def get_time(): return {}

app = FastAPI(); app.include_router(r)
print(any(getattr(x, "path", None) == "/time" for x in app.router.routes))
```

| fastapi | `/time` found in `app.router.routes` |
|---|---|
| 0.136.1 | `True` |
| 0.139.2 | **`False`** |
| 0.140.4 | **`False`** |
| 0.140.11 | **`False`** |

On 0.140.11 the route types present are `['Route', '_IncludedRouter']` — the `APIRoute` is nested inside the `_IncludedRouter` rather than flattened.

## Scope — this is pre-existing, not introduced by the ceiling change

The `fastapi<0.140.5` ceiling from #1387 already permitted 0.139.2 and 0.140.4, both of which exhibit this. CI's "last green" run was on 0.139.2, so the breakage has been present in CI-blessed configurations without being detected — nothing in the suite covers discovery through `include_router`.

Removing the ceiling (#1389) does not cause this and does not worsen it in kind, but it does mean fresh installs land on affected versions.

## Fix

`_extract_route_info` in `_mcp_mesh/shared/server_discovery.py` should recurse into `_IncludedRouter` entries (via `original_router`, or whatever the stable accessor is) rather than assuming a flat `app.router.routes`.

Worth handling defensively across versions — the flattened form is still what 0.136.x produces, so both shapes need to work.

## Test gap this exposes

[...]

PR #1397, opened 2026-07-28
fix(route): rebuild routes through the public FastAPI API, drop the <0.140.5 ceiling

### LINKED ISSUE #1387: SSE routes silently degrade to JSON on FastAPI >= 0.140.5 (removed private get_body_field, swallowed ImportError)
## Summary

`@mesh.route` SSE streaming routes **silently degrade to non-streaming JSON** on FastAPI >= 0.140.5. The response comes back as `application/jsonl` instead of `text/event-stream`, with only a WARNING in the logs and no startup failure.

Because the manifests allow `fastapi>=0.104.0,<1.0.0`, a fresh `pip install mcp-mesh` today resolves a broken version. This affects the current release as shipped — it is not hypothetical.

## Root cause

`_mcp_mesh/pipeline/api_startup/route_integration.py:682` imports four **private** FastAPI internals to rebuild the route handler after SSE wrapping:

```python
from fastapi.dependencies.utils import (
    _should_embed_body_fields,   # explicitly underscore-private
    get_body_field,
    get_dependant,
    get_flat_dependant,
)
from fastapi.routing import request_response
```

FastAPI removed `get_body_field` in **0.140.5**. The `except Exception` at ~line 737 catches the resulting `ImportError`, logs

```
Failed to rebuild route handler for /api/chat: cannot import name 'get_body_field' from 'fastapi.dependencies.utils'
```

and then **`return True`** — reporting success. Without the rebuild, FastAPI keeps calling the original async-generator endpoint and JSON-encodes the result, so the SSE contract is broken while the agent appears healthy.

## Bisected

Reproduced on unmodified `main` by varying only the FastAPI version; `tests/unit/test_route_sse_wrapping.py` (2 tests):

| fastapi | result | `get_body_field` present |
|---|---|---|
| 0.139.2 | PASS | yes |
| 0.140.0–0.140.3 | PASS | yes |
| **0.140.4** | **PASS** | yes |
| **0.140.5** | **FAIL** | **removed** |
| 0.140.7 | FAIL | removed |

CI corroborates: the last green `main` run installed fastapi 0.139.2; the first failing run installed 0.140.7. Nothing else differed (fastmcp 3.4.3, mcp 1.28.1, starlette 1.3.1 in both).

## Severity

The failure mode is the concerning part — not that it breaks, but that it breaks **quietly**:

- Startup succeeds, health checks pass, the route is registered and logs `📡 Route ['POST'] /api/chat -> chat() registered with SSE streaming`.
- Only the content type and framing are wrong, so a client sees a non-streaming JSON body where it expected an event stream.
- Nothing in CI catches it for users, because it depends on which FastAPI version their environment resolved.

This is the same class as #1312 (v3.0.1), where unpinned transitive drift in fastmcp/mcp produced 421s for all Kubernetes Python providers. The remedy recorded there was to pin/lock dependencies; it evidently was not applied broadly enough to cover this.

## Proposed

1. **Pin a ceiling** — `fastapi>=0.104.0,<0.140.5` in both `packaging/pypi/pyproject.toml` and `src/runtime/python/pyproject.toml`. Stops shipping broken SSE and unblocks CI. Immediate.
2. **Stop depending on private FastAPI internals**, or vendor the needed logic. Four private symbols (one explicitly underscore-prefixed) is a standing breakage risk on every FastAPI release; the ceiling only defers it.
3. **Make the failure loud.** Silently degrading SSE to JSON is worse than refusing to start. At minimum this path should not `return True` after the rebuild fails — surface it, or fail the route registration.
4. Consider whether a lockfile or a periodic upper-bound audit is warranted, given this is the second occurrence of the same class.

### LINKED ISSUE #1389: Upgrade to current FastAPI and drop the <0.140.5 ceiling (get_body_field is now private _get_body_field)
## Goal

Remove the `fastapi<0.140.5` ceiling added in #1388 and support current FastAPI. The ceiling is a stopgap: it pins users to 0.140.4 and defers — rather than removes — the risk of depending on FastAPI internals. See #1387 for the breakage it stops.

## What actually changed in 0.140.5

`route_integration.py:682` imports five symbols. Only one moved:

| symbol | 0.140.4 | 0.140.5 |
|---|---|---|
| `get_body_field` | present | **renamed `_get_body_field`, signature changed** |
| `_should_embed_body_fields` | present | present |
| `get_dependant` | present | present |
| `get_flat_dependant` | present | present |
| `fastapi.routing.request_response` | present | present |

The rename is not cosmetic — the first parameter changed shape:

```python
# 0.140.4
def get_body_field(*, flat_dependant: Dependant, name: str, embed_body_fields: bool) -> ModelField | None

# 0.140.5
def _get_body_field(*, body_params: list[ModelField], name: str, embed_body_fields: bool) -> ModelField | None
```

FastAPI's own `routing.py` now calls it as `_get_body_field(body_params=..., ...)`. Since we already compute `get_flat_dependant(...)`, the value needed is `flat_dependant.body_params`, so the adaptation itself is small.

## Approach — two options, and I'd argue for the second

**A. Version-tolerant shim.** Try `get_body_field`, fall back to `_get_body_field` with the reshaped argument. Smallest diff, restores a wide version range immediately.

The cost: FastAPI renamed this symbol *to add a leading underscore*. That is an explicit statement that it is private and unstable. A shim signs us up to chase it on every release, and the failure mode is the one that already bit us — silent SSE degradation rather than a loud break.

**B. Stop depending on the private API.** Determine what the route rebuild actually needs and achieve it through public surface, or vendor the small amount of logic involved so it is version-independent. Larger up-front effort; removes a standing source of the #1312/#1387 failure class.

Whichever is chosen, two things should land with it:

- **Make the failure loud.** The current `except Exception` logs a warning and then `return True`, which is why this shipped silently. A rebuild failure should surface — at minimum not report success.
- **Test across FastAPI versions.** A single pinned version in CI is what let 0.140.5 through unnoticed. Even a two-point matrix (floor and latest) on the SSE tests would have caught it at the source.

## Verification

`src/runtime/python/tests/unit/test_route_sse_wrapping.py` is the reproducer — it fails on >= 0.140.5 and passes below. Bisect and CI evidence are in #1387.

## Related

- #1387 — the breakage and the stopgap ceiling
- #1312 — prior instance of the same class (unpinned transitive drift in fastmcp/mcp causing 421s on Kubernetes Python providers)

dhyansraj:
## Summary

Removes mcp-mesh's dependency on five private FastAPI internals, which lets the `fastapi<0.140.5` ceiling from #1387 come off. Also makes the failure loud, so the next incompatibility can't ship silently like this one did.

## The problem

`route_integration.py` replaced a route's endpoint with an SSE/DI wrapper, then hand-rebuilt FastAPI's cached route state:

```python
from fastapi.dependencies.utils import (
    _should_embed_body_fields, get_body_field, get_dependant, get_flat_dependant,
)
from fastapi.routing import request_response
```

FastAPI renamed `get_body_field` → `_get_body_field` (and reshaped its first parameter) in **0.140.5**. The `ImportError` was swallowed by a broad `except Exception` that logged a warning and then `return True` — so SSE routes silently returned `application/jsonl` instead of `text/event-stream`, with startup succeeding and health checks passing.

## The fix

Everything that block hand-computed — `dependant`, `_flat_dependant`, `_embed_body_fields`, `body_field`, `is_json_stream`, `is_sse_stream`, `app` — is **exactly what `APIRoute.__init__` computes**. So construct a fresh `APIRoute` around the wrapper and let FastAPI do it. Constructor arguments are read back off the existing route **by signature at call time**, so registration options carry over verbatim and a future FastAPI option is picked up rather than silently dropped.

Zero private imports remain; only `from fastapi.routing import APIRoute` (public, documented for custom route classes, and already used by `jobs_cancel_route.py`).

The manual `is_json_stream`/`is_sse_stream` clearing is also gone — FastAPI recomputes both as `False` from the wrapper, which is a coroutine rather than an async generator.

**Route substitution over in-place mutation**, with evidence: `app.routes is app.router.routes` on every matrix version; nothing in `_mcp_mesh/` retains an `APIRoute` (all `.routes` consumers copy out `path`/`methods`/`endpoint`); Starlette walks the list per request with no identity cache. The alternative — harvesting computed state onto the existing object — was rejected because it still reads `_flat_dependant` / `_embed_body_fields`, the same coupling in a different costume.

## Bonus: fixes a second, unreported bug

The old hand-rebuild called `get_dependant()` directly and **silently dropped route-level `dependencies=[Depends(...)]`** — FastAPI re-inserts those, the manual call didn't. Verified against `main`:

| | route-level `Depends` ran | content-type |
|---|---|---|
| old code, 0.136.1 | **False** (dropped) | text/event-stream |
| old code, 0.140.11 | True (rebuild never ran) | **application/jsonl** ← #1387 |
| new code, both | True | text/event-stream |

## Loud failure (#1387's remaining item)

`return False` alone would **not** have been loud: it becomes `{"status": "error"}` → `error_count`, which `execute()` computes and discards, so startup would still succeed. Raising alone wasn't enough either — `_integrate_app_routes` has a per-route `except Exception` that absorbed it into the same discarded counter.

So: a rebuild failure raises `RouteRebuildError`, with a targeted re-raise past that handler → `PipelineStatus.FAILED` → the step is `required=True` → startup aborts. Scope stays tight: only a *rebuild* failure is fatal. The "route not found" path is unchanged (`return False`, nothing was modified).

## Version matrix — the whole point

Passing on a single version is what let this ship. Tests run via isolated `pip --target` dirs so the venv was never modified; counts from `--junitxml` (this suite swallows pytest's stdout summary).

| fastapi | route-integration (61) | full unit suite |
|---|---|---|
| 0.136.1 | 61 passed | 2440 passed / 2 skipped |
| 0.139.2 (last green CI) | 61 passed | — |
| 0.140.4 (last pre-break) | 61 passed | — |
| **0.140.5 (the breaking one)** | **61 passed** | 2440 passed / 2 skipped |
| 0.140.7 | 61 passed | — |
| **0.140.11 (latest)** | **61 passed** | 2440 passed / 2 skipped |

Baseline before the fix on 0.140.7: 2 failures with `Failed to rebuild route handler ... cannot import name 'get_body_field'`.

**CI now pins a floor run** against 0.136.1 for the three route-integration files, placed last so it can't perturb coverage artifacts; the main run covers latest via default resolution.

## Other checks

- 4 new tests prove the loud path: rebuild raises → `RouteRebuildError`, not absorbed by `_integrate_app_routes`, `execute()` returns `FAILED`.
- A test asserts no `fastapi.dependencies` import survives and any `fastapi.routing` import is exactly `APIRoute`.
- Non-SSE DI coverage was thin (skip path only), so `test_non_streaming_di_route_still_works_end_to_end` was added.

## Note — separate pre-existing bug found, filed as #1396

`include_router()` stopped flattening `APIRoute`s into `app.router.routes` around 0.139.x, so `@mesh.route` handlers mounted on an `APIRouter` are **not discovered at all** on ≥0.139. Orthogonal to this change and **not caused by dropping the ceiling** — the `<0.140.5` ceiling already permitted 0.139.2 and 0.140.4, both affected, and CI's "last green" was 0.139.2. Deliberately out of scope; the fix belongs in `server_discovery.py::_extract_route_info`.

Closes #1389
Closes #1387

## Test plan

- [ ] CI green, including the new FastAPI floor job
- [ ] Manual: an SSE `@mesh.route` returns `text/event-stream` on latest FastAPI

## Summary by CodeRabbit

- **New Features**
  - Expanded compatibility with newer FastAPI releases.
  - Improved preservation of route configuration, including dependencies, status codes, tags, response settings, and schema visibility.

- **Bug Fixes**
  - Improved reliability for streaming and SSE-enabled routes across supported FastAPI versions.
  - Route integration failures are now surfaced clearly and prevent startup from continuing with an invalid route configuration.
  - Added safeguards for unsupported route types and route rebuild errors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -282,6 +282,25 @@ jobs:
           name: codecov-${{ matrix.test-group }}
           fail_ci_if_error: false
 
+      # Issue #1387/#1389: @mesh.route route integration is the one place mesh
+      # rebuilds FastAPI's cached route state, so it is the one place a FastAPI
+      # release can break us. Testing a single resolved FastAPI is how 0.140.5
+      # shipped as a silent SSE-to-JSON degradation. The run above uses whatever
+      # pip resolves (latest); re-run the route-integration tests against an
+      # older FastAPI so both ends of the supported range are covered.
+      # Runs last so it cannot perturb coverage/test-result artifacts.
+      - name: Route integration tests against older FastAPI
+        if: matrix.test-group == 'unit' && matrix.python-version == '3.11'
+        env:
+          MCP_MESH_AUTO_RUN: false
+        run: |
+          python -c "import fastapi; print('resolved fastapi:', fastapi.__version__)"
+          pip install "fastapi==0.136.1"
+          python -c "import fastapi; print('pinned fastapi:', fastapi.__version__)"
+          pytest tests/unit/test_route_sse_wrapping.py \
+            tests/unit/test_sse_strips_mcp_mesh_tool_params.py \
+            tests/unit/test_stream_settle.py -q
+
   security-scan:
     name: Security Scanning
     runs-on: ubuntu-latest
--- a/packaging/pypi/pyproject.toml
+++ b/packaging/pypi/pyproject.toml
@@ -59,7 +59,7 @@ requires-python = ">=3.11"
 dependencies = [
     # Rust core runtime (required - no Python fallback)
     "mcp-mesh-core>=3.3.1",
-    "fastapi>=0.104.0,<0.140.5",
+    "fastapi>=0.104.0,<1.0.0",
     "uvicorn>=0.24.0,<1.0.0",
     "httpx>=0.25.0,<1.0.0",
     "aiohttp>=3.8.0,<4.0.0",
--- a/src/runtime/python/_mcp_mesh/pipeline/api_startup/route_integration.py
+++ b/src/runtime/python/_mcp_mesh/pipeline/api_startup/route_integration.py
@@ -799,4 +799,18 @@ class RouteIntegrationStep(PipelineStep):
         kwargs = {
             name: getattr(route, name) for name in init_params if hasattr(route, name)
         }
+        # Anything the constructor accepts but the route doesn't mirror as an
+        # attribute can't be round-tripped, so it falls back to the FastAPI
+        # default on the rebuilt route. That never happens on today's FastAPI
+        # (all parameters round-trip); log it so a future release that breaks
+        # the symmetry is visible instead of silently degrading.
+        skipped = [name for name in init_params if name not in kwargs]
+        if skipped:
+            logger.debug(
+                "Route rebuild for '%s' %s: APIRoute parameter(s) %s not readable "
+                "off the existing route, falling back to FastAPI defaults",
+                route.path,
+                sorted(getattr(route, "methods", None) or []),
+                ", ".join(skipped),
+            )
         return APIRoute(path=route.path, endpoint=wrapped_handler, **kwargs)
--- a/src/runtime/python/_mcp_mesh/pipeline/api_startup/route_integration.py
+++ b/src/runtime/python/_mcp_mesh/pipeline/api_startup/route_integration.py
@@ -14,6 +14,23 @@ from ..shared import PipelineResult, PipelineStatus, PipelineStep
 logger = logging.getLogger(__name__)
 
 
+class RouteRebuildError(RuntimeError):
+    """Raised when a route's FastAPI state could not be rebuilt after wrapping.
+
+    A rebuild failure means the route is still REGISTERED AND SERVING, but
+    FastAPI's cached dispatch state (dependant, body field, per-route ASGI
+    handler) no longer matches the handler mesh installed. For an SSE route
+    that surfaces as ``application/jsonl`` instead of ``text/event-stream``
+    — a broken wire contract on a route that looks healthy.
+
+    This is deliberately fatal (issue #1387): the previous code logged a
+    warning and reported success, which is how FastAPI 0.140.5's rename of
+    ``get_body_field`` shipped as a silent SSE degradation. Route
+    integration is a ``required=True`` pipeline step, so raising here fails
+    the pipeline and aborts startup rather than serving a wrong response.
+    """
+
+
 _SSE_HEADERS = {
     "Cache-Control": "no-cache",
     "X-Accel-Buffering": "no",
@@ -412,6 +429,15 @@ class RouteIntegrationStep(PipelineStep):
                 else:
                     integration_results["error_count"] += 1
 
+            except RouteRebuildError:
+                # Do NOT absorb into error_count like other per-route
+                # failures. A rebuild failure leaves a live route whose
+                # FastAPI dispatch state contradicts the handler mesh
+                # installed (e.g. an SSE route answering application/jsonl).
+                # Let it reach execute(), which fails this required step so
+                # startup aborts instead of serving a broken contract
+                # (issue #1387).
+                raise
             except Exception as e:
                 self.logger.error(f"❌ Failed to integrate route '{route_name}': {e}")
                 integration_results["error_count"] += 1
@@ -642,6 +668,11 @@ class RouteIntegrationStep(PipelineStep):
         """
         Replace the route handler in FastAPI's router.
 
+        The matched route is swapped for a freshly-constructed ``APIRoute``
+        carrying the wrapped endpoint, so FastAPI recomputes its own cached
+        dispatch state instead of mesh hand-patching it through private
+        internals.
+
         Args:
             app: FastAPI application instance
             path: Route path to find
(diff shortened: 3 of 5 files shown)


dhyansraj:
### Review fixes

Two findings reviewed. **One fixed, one skipped.**

#### Fixed — silently discarded `APIRoute` constructor parameters

`_rebuild_api_route` collected constructor arguments with `if hasattr(route, name)`, silently dropping any parameter not mirrored as a same-named attribute. Harmless today — all round-trip — but the reason we rebuild *by signature* is to pick up new FastAPI options automatically, and a silent skip defeats exactly that. A future release breaking attribute symmetry would lose a setting from the rebuilt route with no signal, which is the same silent-degradation class this PR exists to remove.

Skipped names are now collected and reported once at debug level with the route path and methods. **Behaviour is unchanged** — same skips, same return, purely additive (zero removed lines).

Verified in both directions, and on both FastAPI generations (the repo venv is 0.136.1, so a scratch venv on **0.140.12** was used for the other):

- does **not** fire today — 25 of 25 considered parameters round-trip, identical on 0.136.1 and 0.140.12
- **does** fire when it should — with a synthetic keyword-only parameter injected into the signature:
  ```
  DEBUG Route rebuild for '/things/{tid}' ['GET']: APIRoute parameter(s) future_fastapi_option
        not readable off the existing route, falling back to FastAPI defaults
  ```

Two tests added to `TestRouteRebuildIsVersionIndependent`, which already owns the "rebuild carries every registration option" contract. The more valuable one asserts the log *doesn't* fire — a live tripwire that turns a future attribute-symmetry break into a test failure at upgrade time rather than a debug line nobody reads.

Correction to an earlier count: the constructor has 27 parameters, of which **25** are considered (`self`/`path`/`endpoint` excluded, `**kwargs`-kind filtered).

Suite: **2444 passed, 0 failed, 2 skipped** (both pre-existing live-integration opt-ins); `test_route_sse_wrapping.py` 36/36.

#### Skipped — the `fastapi` upper-bound difference between manifests

Not an inconsistency; each file follows its own established convention:

| file | deps carrying an upper bound |
|---|---|
| `packaging/pypi` (published) | 13 of 25 — 52% |
| `src/runtime/python` (source) | 1 of 24 — 4% |

Both lines are also **byte-identical to the pre-ceiling state** (`d43935619~1`) — this PR restored them exactly rather than introducing a difference.

On whether FastAPI 1.x is supported: **no 1.x release exists** (latest 0.140.12), so `<1.0.0` on the published manifest is a forward-looking major-version guard, which is why a distributed package carries one and the source manifest doesn't.

Homogenizing would mean adding 12 upper bounds to the source manifest or stripping 13 from the published one — a dependency-policy change affecting resolution for every user, well beyond this PR. Worth its own issue if a single convention is wanted.

PR #1398, opened 2026-07-28
fix(route): discover and integrate @mesh.route handlers mounted via include_router

### LINKED ISSUE #1396: @mesh.route handlers mounted via include_router are not discovered on FastAPI >= 0.139

dhyansraj:
## Summary

`@mesh.route` handlers registered on an `APIRouter` and mounted with `include_router()` were **not discovered at all** on FastAPI >= 0.139.x — they served as plain endpoints with no mesh DI. This is a documented pattern with its own example file (`examples/simple/simple_fastapi_router.py`, `docs/python/fastapi-integration.md`).

Around 0.139.x, `include_router()` stopped flattening the router's `APIRoute` objects into `app.router.routes`; it now inserts a single `_IncludedRouter`. mcp-mesh walked that flat list in five places.

**Pre-existing, not caused by #1397.** The `<0.140.5` ceiling already permitted 0.139.2 and 0.140.4, both affected — and CI's "last green" was 0.139.2, so this was broken in blessed configurations without detection. No test covered discovery through `include_router`, which is why it survived several releases.

## Approach

New `_mcp_mesh/shared/fastapi_routes.py` — one place that knows how to walk an app's routes, exposing `iter_app_routes(app) -> Iterator[RouteRef]` (route, **effective** path, methods, endpoint, owning list + index, `included` flag).

**It imports nothing from fastapi at all** (asserted by a test), which is what lets it span 0.136.1 → latest with no version branching. On <0.139 nothing exposes `original_router` and the walk collapses to exactly the old flat iteration.

**Discriminator: duck-type on `original_router`, not the class name and not `.routes`.** `.routes` alone would be wrong — Starlette `Mount`/`Host` expose it, and mesh itself mounts FastMCP as a catch-all, so blanket recursion would pull a mounted sub-app's routes into discovery. `original_router` is absent on both, excluding them for free.

**Effective paths come from `effective_route_contexts()`.** This corrects an assumption: `.path` is pre-prefixed only at depth 1. With `outer(prefix="/api/v1")` including `inner`, the nested route's own `.path` is `/inner/deep` while the app serves `/api/v1/inner/deep`. That path becomes the `METHOD:path` route id sent to the registry, so getting it wrong is not cosmetic. FastAPI composes it; mesh shouldn't re-derive it.

## Correction: in-place mutation does not work either

The obvious fix — mutate the nested route in place, since replacing it in `original_router.routes` doesn't take effect — is also wrong. Measured on 0.136.1 / 0.139.2 / 0.140.13:

| strategy | cold cache | **warm cache** (one request first) |
|---|---|---|
| list replacement | wrapper serves | **original still serves** |
| in-place attribute copy | wrapper serves | **original still serves** |
| list replacement + cache invalidation | wrapper serves | **wrapper serves** |

`_IncludedRouter.effective_candidates()` caches `_EffectiveRouteContext` objects keyed on `original_router._get_routes_version()`, and `APIRoute.get_route_handler()` reads `dependant` / `body_field` / `_embed_body_fields` / `is_json_stream` **off that context** when one is in scope. On a warm cache the route object is bypassed entirely. The version counter is only bumped by `add_api_route` / `include_router`, so any list assignment is invisible to it.

Result: **one uniform strategy at every depth** — replace in the owning list (keeping #1397's rebuild), invalidate, then verify. Simpler than a split strategy.

`invalidate_route_caches` calls `_mark_routes_changed()` only if callable — the single private touch, guarded, never imported. It cannot degrade silently, because the swap is verified: if the new route isn't reachable afterwards, `RouteRebuildError` propagates past `_integrate_app_routes`' handler → `PipelineStatus.FAILED` on a `required=True` step → startup aborts. #1397's loud-failure property is preserved and extended to "found, swapped, but didn't take".

## Five call sites

| site | verdict |
|---|---|
| `server_discovery.py:302` | **fixed** — the reported bug |
| `route_integration.py:704`/`:754` | **fixed** — broken twice over: never saw the route, and its replacement wouldn't have taken on a warm cache |
| `a2a_startup/fastapi_discovery.py:94` | **fixed** — not broken today (`mesh.a2a.mount()` uses top-level `add_api_route`), but it answers "does this app host the surface", and answering from a partial view is the multi-app misrouting the filter exists to prevent |
| `mesh/a2a.py:1471` | **fixed — a fifth site not in the issue.** `mount()`'s collision guard scanned only the top level, so on ≥0.139 it was blind to a foreign `include_router` route owning `card_path`/`rpc_path` — the silent hijack its docstring says must never happen — and would then have added a second registration for the same path |
| `jobs_cancel_route.py:112` | **skipped, deliberately** — it asks "did *this step* already insert its route", and that insert is always top-level, so the top-level list is the exact answer. Widening it turns it into a user-collision check whose failure mode is silently dead MeshJob cancellation |

## Verification

**The example is the acceptance test** — driven through the real pipeline (`FastAPIAppDiscoveryStep` → `RouteIntegrationStep`) with a dependency pushed through the heartbeat funnel and a real request:

| | discovered | integrated | `GET /api/v1/time` |
|---|---|---|---|
| before, 0.136.1 | 3 | 3 | `dependency_injected: True` |
| **before, 0.140.13** | **0** | SKIPPED | **`dependency_injected: False`** |
| after, 0.136.1 | 3 | 3 | `dependency_injected: True` |
| **after, 0.140.13** | **3** | **3** | **`dependency_injected: True`** |

Also confirmed on 0.139.2, 0.140.4, 0.140.5.

**Version matrix**, route test files, counts from `--junitxml` via isolated `pip --target`:

0.136.1 (79 passed / 1 skipped), 0.139.2, 0.140.4, 0.140.5, 0.140.7, 0.140.11, 0.140.13 — **80 passed on every version above the floor**. The one skip is `test_rebuild_that_does_not_take_effect_is_loud`, guarded on 0.136.1 since it flattens and has no cache to go stale.

**Full CI-equivalent suite**: 2461 tests, 0 failures on 0.136.1, 0.140.5, and 0.140.13.

**Tests proven real, not tautological** — a detached worktree at `main` with the new test module copied in but the call sites unfixed produces **7 failures on latest** (both discovery tests, all four integration tests, the end-to-end injection test) and **0 on 0.136.1**, matching the bug's version profile exactly.

17 new tests in `test_route_include_router_1396.py`, joined to the FastAPI floor job from #1397 so both ends of the range run them.

## Follow-up

`server_discovery.py:87`'s `router_routes_count` is now inconsistent with `len(routes)` on ≥0.139 (it counts one entry per included router). It has no consumer anywhere in the repo — only ever written — so it was left rather than widen the diff. Worth deleting or renaming if anyone touches that dict.

Closes #1396

## Test plan

- [ ] CI green, including the FastAPI floor job
- [ ] Manual: `examples/simple/simple_fastapi_router.py` gets DI on latest FastAPI

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved support for FastAPI routes registered through `include_router()`, including nested routers.
  * Fixed route discovery, A2A detection, and endpoint integration for included routes.
  * Ensured streaming routes and dependency injection continue working after integration.
  * Added safeguards to detect when route updates fail to take effect.

* **Tests**
  * Added comprehensive regression coverage for included, nested, and mounted routes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -282,12 +282,15 @@ jobs:
           name: codecov-${{ matrix.test-group }}
           fail_ci_if_error: false
 
-      # Issue #1387/#1389: @mesh.route route integration is the one place mesh
-      # rebuilds FastAPI's cached route state, so it is the one place a FastAPI
-      # release can break us. Testing a single resolved FastAPI is how 0.140.5
-      # shipped as a silent SSE-to-JSON degradation. The run above uses whatever
-      # pip resolves (latest); re-run the route-integration tests against an
-      # older FastAPI so both ends of the supported range are covered.
+      # Issue #1387/#1389/#1396: @mesh.route route handling is the one place
+      # mesh reads and rebuilds FastAPI's own route state, so it is the one
+      # place a FastAPI release can break us. Testing a single resolved FastAPI
+      # is how 0.140.5 shipped as a silent SSE-to-JSON degradation, and how
+      # 0.139's include_router() change shipped as silently-undiscovered
+      # routes. The run above uses whatever pip resolves (latest); re-run the
+      # route files against an older FastAPI so both ends of the supported
+      # range are covered — 0.136.1 still flattens included routers, latest
+      # does not, and the same code must handle both.
       # Runs last so it cannot perturb coverage/test-result artifacts.
       - name: Route integration tests against older FastAPI
         if: matrix.test-group == 'unit' && matrix.python-version == '3.11'
@@ -298,6 +301,7 @@ jobs:
           pip install "fastapi==0.136.1"
           python -c "import fastapi; print('pinned fastapi:', fastapi.__version__)"
           pytest tests/unit/test_route_sse_wrapping.py \
+            tests/unit/test_route_include_router_1396.py \
             tests/unit/test_sse_strips_mcp_mesh_tool_params.py \
             tests/unit/test_stream_settle.py -q
 
--- a/src/runtime/python/_mcp_mesh/pipeline/a2a_startup/fastapi_discovery.py
+++ b/src/runtime/python/_mcp_mesh/pipeline/a2a_startup/fastapi_discovery.py
@@ -16,6 +16,7 @@ to the registry.
 import logging
 from typing import Any
 
+from ...shared.fastapi_routes import iter_app_routes
 from ...shared.server_discovery import ServerDiscoveryUtil
 from ..shared import PipelineResult, PipelineStatus, PipelineStep
 
@@ -90,9 +91,14 @@ class A2AFastAPIDiscoveryStep(PipelineStep):
                 app = app_info.get("instance")
                 if app is None:
                     continue
-                app_route_paths = {
-                    getattr(r, "path", None) for r in app.router.routes
-                }
+                # Walk everything the app serves, not just its top-level
+                # list: routes mounted with include_router() are no longer
+                # flattened into it (issue #1396). ``mesh.a2a.mount()``
+                # registers its surfaces top-level, so this filter is not
+                # broken today — but it is a "does this app host the surface"
+                # question, and answering it off a partial view is how the
+                # multi-app mystery this filter exists to prevent comes back.
+                app_route_paths = {ref.path for ref in iter_app_routes(app)}
                 if app_route_paths & wanted_paths:
                     filtered_apps[app_id] = app_info
 
--- a/src/runtime/python/_mcp_mesh/pipeline/api_startup/route_integration.py
+++ b/src/runtime/python/_mcp_mesh/pipeline/api_startup/route_integration.py
@@ -38,6 +38,35 @@ _SSE_HEADERS = {
 }
 
 
+def _methods_overlap(route_methods: Any, call_methods: Any) -> bool:
+    """True when a route's verbs and the verbs being integrated intersect.
+
+    The two sides come from independent places — ``RouteRef.methods`` is
+    whatever FastAPI stored on the route (a set on an ``APIRoute``, a list on
+    an effective-route context), while the integration call's ``methods``
+    comes from ``route_info["methods"]`` — so they can differ in container
+    type, order and (for a hand-built route registration) case. Both are
+    normalised to upper-case sets before comparing.
+
+    Intersection, not equality: a single registration may serve several verbs
+    (``methods=["GET", "POST"]``) and must still be rebuilt when integrating
+    a handler discovered under one of them. Intersection is also immune to
+    FastAPI/Starlette adding ``HEAD`` alongside ``GET`` — an extra verb on
+    the route side can never turn a real match into a miss, and cannot create
+    one either, since ``HEAD`` never appears alone on the call side.
+
+    An empty collection on either side means "unconstrained" and matches, so
+    route objects that carry no ``methods`` at all (websocket routes, which
+    the ``APIRoute`` guard below rejects by design) keep reaching that guard
+    instead of being silently skipped.
+    """
+    route_verbs = {str(m).upper() for m in route_methods or ()}
+    call_verbs = {str(m).upper() for m in call_methods or ()}
+    if not route_verbs or not call_verbs:
+        return True
+    return not route_verbs.isdisjoint(call_verbs)
+
+
 def _resolve_user_function(handler: Any) -> Any:
     """Return the user-authored function underneath a possibly wrapped handler.
 
@@ -673,6 +702,13 @@ class RouteIntegrationStep(PipelineStep):
         dispatch state instead of mesh hand-patching it through private
         internals.
 
+        Every registration of ``original_handler`` at ``path`` whose verbs
+        this call covers is rebuilt, not just the first one found: the same
+        function can be registered twice at one path under different verbs
+        (``add_api_route(..., methods=["GET"])`` then ``methods=["POST"]``),
+        which produces two ``APIRoute`` objects that are indistinguishable by
+        path and endpoint alone.
+
         Args:
             app: FastAPI application instance
             path: Route path to find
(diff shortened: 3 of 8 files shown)


dhyansraj:
### Review fixes (commit `751f1fcc1`)

Four findings, **all valid, none skipped.**

#### 1. Route matching ignored `methods`

`_replace_route_handler` matched on path + endpoint only, then returned on the first hit. Two same-path registrations of one handler are indistinguishable to that, so which gets wrapped was **iteration-order dependent**.

Demonstrated pre/post on FastAPI 0.136.1 (pre-fix run from a worktree at HEAD), handler registered at `/x` under both GET and POST:

| scenario | pre-fix GET / POST wrapped | post-fix |
|---|---|---|
| integrate `["POST"]` | **True / False** → returns True | False / **True** → True |
| integrate `["GET","POST"]` | **True / False** → True | **True / True** → True |
| integrate `["DELETE"]` | **True / False** → True | False / False → **False** |
| `["POST"]`, registration order reversed | **False / True** → True | False / True → True |

Rows 1 and 4 are the order-dependence proof — identical call, only registration order swapped, and the pre-fix wrapper lands on a different verb. Row 3 shows pre-fix wrapping a route whose verb it was never asked about *and reporting success*.

**Comparison semantics — case-normalised set intersection**, chosen against the data flow rather than assumed:
- `ref.methods` derives from `APIRoute.methods` (a `set`), so its list order is hash order — sets are the only stable comparison.
- `APIRoute.__init__` upper-cases, but a hand-built `route_info` need not, so both sides are normalised.
- Intersection, not equality: one registration may serve several verbs and must still rebuild when the call covers one.
- **`HEAD`**: verified on 0.136.1 / 0.139.2 / 0.140.5 / 0.140.13 that `APIRoute.__init__` sets `self.methods` itself and never calls Starlette's `Route.__init__`, so no `HEAD` is auto-added. Intersection is immune either way — an extra route-side verb can't turn a match into a miss, and `HEAD` never appears alone on the call side.
- Empty on either side matches, so routes carrying no `methods` still reach the `APIRoute` guard and fail loudly rather than being silently skipped into "not found".

**No double-wrapping** — verified, not assumed: the walk is snapshotted before any swap (necessary anyway, since the rebuild assigns into the lists `iter_app_routes` reads), and each rebuild reads the pre-swap `ref.route`. A router included twice at the same prefix rebuilds from the original endpoint both times.

#### 2. Test cleanup was conditional

`_route_wrapper_registry` was cleared as a trailing statement, so any failure mid-test leaked it into subsequent tests. Now unconditional. Proven: with a deliberate assertion failure injected, the pre-fix copy produced **two** failures — the deliberate one plus `LEAKED into next test: ['GET:/api/v1/time']` — while the fixed copy produced only the deliberate one.

#### 3. Both early `RouteRebuildError` guards were untested

Only the late "rebuild didn't take effect" guard had coverage. Added focused tests for a `RouteRef` resolving to a non-`APIRoute` and one with `container`/`index` unset.

#### 4. Moved `test_invalidate_route_caches_is_a_noop_without_routers` into `TestIterAppRoutes` — it exercises the shared walker, not integration.

**Tests:** file 21 passed / 1 skipped (was 13/1). Full CI-equivalent suite **2466 tests, 0 failures**. Version matrix on the route files: 0.136.1 (60 passed / 1 skipped), 0.139.2, 0.140.5, 0.140.13 — 61 passed each.

### Scope note worth recording

`fastapi_discovery._map_routes_to_apps` keys `app_routes` by `endpoint_name`, so **the real pipeline collapses two same-name registrations into a single `route_info`** (last wins) and calls integration once per handler name. This fix makes that single call target the verbs mesh actually registered in `DecoratorRegistry` rather than an arbitrary one — but the *other* registration of that handler stays unintegrated by construction, upstream of this function.

Wiring both verbs end-to-end would mean re-keying that mapping to something like `(endpoint_name, path, verbs)`. Separate change, deliberately not folded in.

Issue #1399, opened 2026-07-28
scripts/test_bump_version.py is never run by CI — the over-match guard's own tests are unenforced


dhyansraj:
## Summary

`scripts/test_bump_version.py` is **never executed automatically**. Grepping `.github/workflows/`, the Makefile, and the docs returns zero references to it — the only file mentioning it is the file itself.

It currently holds 18 checks, including the tests added in #1395 for the over-match guard: that the guard flags a third-party pin, that a file path is not accepted as proof of mesh ownership, that `OVERMATCH_ALLOWLIST` entries stay narrow, and that the anchored `version` patterns skip nested third-party pins.

## Why it matters

The guard itself still protects a release — `overmatch_guard()` runs inside `bump_version.py`, so a real bump exercises it. What is unprotected is **the guard's own correctness**: someone editing `bump_version.py` can weaken or break the guard, and nothing fails until the next release bump, which is precisely the moment you least want to discover it.

This is the same shape as the defects the guard was written for. #1379's over-match reached a release because nothing checked for it; a broken guard would reach a release for the same reason.

The narrowness test is the clearest example. It exists so nobody adds a blanket `OVERMATCH_ALLOWLIST` entry that neuters the guard — demonstrated in #1395 by temporarily adding `Exemption(glob="**", pattern=r".+")`, which made three other guard tests fail. That test only helps if it runs.

## Fix

Wire it into CI. It is a plain `python3 scripts/test_bump_version.py` (also `pytest`-compatible) with no dependencies beyond the standard library, so it is cheap to add to an existing job rather than needing its own.

Worth auditing at the same time whether anything else under `scripts/` has tests that never run.

## Context

Found while placing a new manifest-invariant test for #1383. That test was deliberately put in `src/runtime/python/tests/unit/` instead — `ci.yml` runs `pytest tests/unit/ _mcp_mesh/` from `src/runtime/python` on every push, so the invariant is actually enforced there.

dhyansraj:
Worth widening this issue's scope slightly, based on what #1400 turned up.

While bounding the `mcp-mesh[litellm]==` pattern we found the same prefix-match defect in three sibling handlers (`bump_version.py` :462, :488, :577). All are fixed now, but the important finding is **this defect class is invisible to both guards**.

On a deliberately-corrupted run — `mcp-mesh>=3.3.10` rewritten to the nonexistent `>=3.3.20` by a 3.3.1 → 3.3.2 bump — both guards reported green:

```
[...]

PR #1400, opened 2026-07-28
feat(litellm): add the [litellm] extra, pin it from scaffold, document the split

dhyansraj:
## Summary

Items 3-5 of #1383, as the **first of two steps**. This change is deliberately **non-breaking**: `litellm` stays in the base install. The actual removal happens at the next major, at which point the extra added here already works with no further change.

Items 1 and 2 shipped in #1386 (`_require_litellm()` resolving at point-of-use and raising actionable guidance). That guidance already tells users to run `pip install 'mcp-mesh[litellm]==<version>'` — and until now the extra did not exist, so the command emitted `WARNING: mcp-mesh 3.3.1 does not provide the extra 'litellm'` and installed nothing.

## Item 3 — the extra

Added to both manifests, **restating the base requirement verbatim** rather than being an empty stub. It is redundant today (litellm is also in base, so the extra is a harmless no-op), but at the next major the base entry is simply deleted and the extra keeps working. An empty stub would silently become a no-op exactly when it started mattering.

Published-wheel metadata, built the way `release.yml` does:

```
Requires-Dist: litellm!=1.82.7,!=1.82.8,>=1.80.5              <- still base
Provides-Extra: litellm
Requires-Dist: litellm!=1.82.7,!=1.82.8,>=1.80.5; extra == 'litellm'
```

The duplicated pin is guarded by `tests/unit/test_litellm_extra_pin_sync.py`, which asserts base and extra match *and* that both manifests agree with each other. Deliberately placed in the Python suite rather than `scripts/test_bump_version.py` — see the note at the bottom.

## Item 4 — scaffolder emits the pinned extra

New `scaffold/model_dispatch.go` (`RequiresLiteLLM`), consumed by the `llm-provider` requirements template. Only that template carries it: it is the only Python template whose context has a `Model`, and the only one that dispatches — an `llm-agent` consumer delegates to a provider.

Version pinning follows `Dockerfile.tmpl`'s existing mechanism (a literal that `bump_version.py` rewrites) rather than introducing a template variable, since the meshctl binary's own version is `"dev"` in local builds and two version-of-record mechanisms in one template directory would be worse. New handler added; verified both lines move together in a real bump.

## Correction to the plan: `vertex_ai/*` is big-3

I briefed this as long-tail. That was wrong, and verifying it was the right call: `vertex_ai` is a registered key in `ProviderHandlerRegistry._handlers` mapping to `GeminiHandler`, and `gemini_native.supports_model()` matches `_VERTEX_PREFIX`. It dispatches through the bundled `google-genai` SDK exactly like `gemini/*` — only the auth differs. It gets **no** litellm line, and the test asserts that with the reasoning inline.

## Item 5 — docs

`man/content/llm.md`, `man/content/scaffold.md`, `docs/python/llm/index.md`. Wording is **true today**: LiteLLM ships in the base install, declaring the extra changes nothing right now, and it survives the move. The `_typescript`/`_java` man variants were deliberately not touched — they document the Vercel AI SDK and Spring AI, neither of which has a LiteLLM path or a `pip install`.

## Verification

**Non-breaking, measured** — full resolutions with `--ignore-installed` against a `main` worktree and this branch:

```
main   plain    : 111 packages
branch plain    : 111 packages   symmetric-diff vs main = NONE
branch [litellm]: 111 packages   symmetric-diff vs branch plain = NONE
```

**The extra resolves** — the `does not provide the extra` warning is gone.

**Sync test proven real** — appending `,!=1.99.0` to one extra pin fails both the per-manifest and cross-manifest assertions; reverted, 3 pass.

**Bump** — a real `3.3.1 → 3.3.2` in a throwaway worktree: 470 files, over-match guard and coverage guard both clean. The template pin moves in lockstep with the Docker tag, and the four upstream `litellm>=1.80.5` lines (their version, not ours) are untouched.

**Generated requirements.txt** — `anthropic/claude-sonnet-5` and `vertex_ai/gemini-2.5-flash` produce no line; `bedrock/…` and `ollama/llama3` produce the pinned extra with an explanatory comment.

Go: `go build ./...` clean, `go test ./src/core/...` all packages ok, 24 new `RequiresLiteLLM` sub-cases. Python: **2469 tests, 0 failures**.

## Two findings worth separate attention

**1. `bump_version.py`'s coverage guard read past the new pin.** Its pattern was `mcp-mesh(?:>=|==)`, which does not match `mcp-mesh[litellm]==X`. Extended here to allow the optional bracket group — otherwise the new pin could silently go stale across a release, which is exactly the class of bug that guard exists to catch.

**2. `bedrock/*` and `databricks/*` are long-tail despite `anthropic_native.supports_model()` matching them.** Handler selection is by *vendor*, and neither string is a registry key, so they land on `GenericHandler` → LiteLLM and `supports_model` is never consulted. That looks like unreachable capability in `anthropic_native`, and it means the `[anthropic-bedrock]` extra's documented use case may not take the native path today. Not touched — it is a behavioural question outside this issue. Worth its own issue.

Also filed **#1399**: `scripts/test_bump_version.py` is never executed by CI, so the over-match guard's own tests (added in #1395) are unenforced. That is why the new sync test went into the Python suite instead.

Refs #1383

## Test plan

- [ ] CI green
- [ ] `pip install 'mcp-mesh[litellm]'` resolves without warning
- [ ] A scaffolded long-tail provider's `requirements.txt` carries the pinned extra; a big-3 one does not

## Summary by CodeRabbit

* **New Features**
  * `meshctl scaffold` now determines whether a chosen model is routed via bundled native adapters or via LiteLLM, and conditionally adds the `mcp-mesh[litellm]` requirement to generated `requirements.txt`.
  * Added a new Python `litellm` optional install extra (aligned with existing LiteLLM pinning).

* **Documentation**
  * Refined LLM integration and scaffold docs to clarify native vs LiteLLM dispatch and current vs future install behavior.

* **Bug Fixes**
  * Improved version-bump logic to correctly handle `mcp-mesh[litellm]` (including extras) without false matches.

* **Tests**
  * Added coverage for model dispatch decisions and `litellm` extra/version pin synchronization.

### DIFF
--- a/cmd/meshctl/templates/python/llm-provider/requirements.txt.tmpl
+++ b/cmd/meshctl/templates/python/llm-provider/requirements.txt.tmpl
@@ -1,3 +1,12 @@
 # {{ .Name }} dependencies
 # Add your third-party dependencies here
 # Note: mcp-mesh is provided by the runtime environment (local venv or Docker base image)
+{{- if .RequiresLiteLLM }}
+
+# LiteLLM provider path (model: {{ .Model }}).
+# mcp-mesh bundles native SDK adapters for Anthropic, OpenAI and Gemini only;
+# every other vendor dispatches through LiteLLM. LiteLLM ships in the base
+# mcp-mesh install today, so this line currently changes nothing. It is
+# declared so this agent keeps working when LiteLLM becomes an opt-in extra.
+mcp-mesh[litellm]==3.3.1
+{{- end }}
--- a/docs/python/llm/index.md
+++ b/docs/python/llm/index.md
@@ -38,7 +38,10 @@ export OPENAI_API_KEY=sk-...
     release. Declaring `mcp-mesh[litellm]` now is a no-op today and keeps a
     long-tail agent working across that change; `meshctl scaffold` already
     writes that pin into the generated `requirements.txt` when the selected
-    model is not Anthropic, OpenAI or Gemini.
+    model is not Anthropic, OpenAI, Gemini or Vertex AI. `vertex_ai/*` is
+    native despite the long-tail-looking name: it runs the same Gemini models
+    through the same bundled SDK, only the auth differs (ADC / Workload
+    Identity instead of an AI Studio API key).
 
 ## @mesh.llm Decorator
 
--- a/docs/python/llm/index.md
+++ b/docs/python/llm/index.md
@@ -18,11 +18,10 @@ MCP Mesh provides first-class support for LLM-powered agents through the `@mesh.
 
 ## What's Included
 
-The `mcp-mesh` package includes LLM support out of the box via LiteLLM:
+The `mcp-mesh` package includes LLM support out of the box:
 
-- **Claude** (Anthropic)
-- **GPT** (OpenAI)
-- **And 100+ other providers** via LiteLLM
+- **Claude** (Anthropic), **GPT** (OpenAI) and **Gemini** (Google) dispatch through bundled native SDK adapters
+- **100+ other providers** (Bedrock, Cohere, Ollama, …) dispatch through LiteLLM
 
 No additional packages needed. Just set your API keys:
 
@@ -30,6 +29,16 @@ No additional packages needed. Just set your API keys:
 export ANTHROPIC_API_KEY=sk-ant-...
 export OPENAI_API_KEY=sk-...
 ```
+
+!!! note "LiteLLM"
+
+    The big-3 vendors above need nothing beyond the base install. LiteLLM is
+    also part of the base install today, so long-tail providers work with no
+    extra step either — but it is moving to an opt-in extra in a future major
+    release. Declaring `mcp-mesh[litellm]` now is a no-op today and keeps a
+    long-tail agent working across that change; `meshctl scaffold` already
+    writes that pin into the generated `requirements.txt` when the selected
+    model is not Anthropic, OpenAI or Gemini.
 
 ## @mesh.llm Decorator
 
--- a/packaging/pypi/pyproject.toml
+++ b/packaging/pypi/pyproject.toml
@@ -149,6 +149,27 @@ gemini = [
     # error.
     "google-genai>=1.22"
 ]
+litellm = [
+    # LiteLLM long-tail provider path (issue #1383).
+    #
+    # DELIBERATELY REDUNDANT TODAY: ``litellm`` is still a BASE dependency
+    # above, so ``pip install mcp-mesh[litellm]`` resolves to exactly the same
+    # set as ``pip install mcp-mesh``. The extra exists now, ahead of the move,
+    # so that a pin written today (by a user, by ``meshctl scaffold``'s
+    # generated requirements.txt, or by the install guidance in
+    # ``mesh.helpers._require_litellm``) keeps working unchanged when the base
+    # entry is deleted at the next major — at that point THIS is the only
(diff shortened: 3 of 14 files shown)


dhyansraj:
Review follow-ups pushed in `cb896044b`. Both findings verified valid against current code; nothing skipped.

**1. `vertex_ai` missing from the documented native set** — confirmed a real doc/code disagreement. `nativeVendorPrefixes` is `['anthropic','gemini','openai','vertex_ai']`, but `docs/python/llm/index.md`, `meshctl man llm` and `meshctl man scaffold` all named only three. Verified empirically: `--model vertex_ai/gemini-2.5-flash` produces a `requirements.txt` with **no** pin, while `--model ollama/llama3` gets `mcp-mesh[litellm]==3.3.1`. All three surfaces now name `vertex_ai`, with a note on why the name misleads (same Gemini models via `GeminiHandler`, different auth). The generated-comment in `requirements.txt.tmpl` and the `ImportError` text in `helpers.py` carry the same three-vendor sentence but were left alone deliberately — both are only reached on the LiteLLM path, so a `vertex_ai` user never sees them.

**2. Unbounded version regex** — confirmed, and reproduced rather than assumed. On a 3.3.1 → 3.3.2 bump the handler rewrote a *prefix*, turning `mcp-mesh[litellm]==3.3.10` into the nonexistent `==3.3.20`; the guard had the mirror bug, flagging `==3.3.10` as a stale `3.3.1`. Both now carry the terminal boundary.

Two things worth flagging beyond the finding as written:

- **The requested `2.8.0rc1` case needed the shared `boundary` widened**, `(?![\d.\-+])` → `(?![\w.\-+])`. The old form blocks digits so `2.8.00` was already handled, but not letters. I widened the shared constant rather than adding a second one, and checked all 17 pattern/input combinations that use it: three behaviour flips, all letter-suffixed prereleases, no exact-version case changed for any pattern (`mcpmesh/cli:3.3.1-alpine` behaves identically before and after). Those flips are correct — handlers substitute an exact `OLD`, so no handler can ever rewrite `2.8.0rc1`; flagging it produced a survivor the bump was structurally incapable of clearing.
- **Three sibling handlers had the identical defect** and are fixed in the same commit: `bump_version.py:462`, `:488`, `:577` (`(mcp-mesh>=)OLD` / `(mcp-mesh==)OLD`). All three globs cover real files carrying real pins, so this was live, not theoretical.

**This defect class is invisible to both guards.** On the corrupted run the over-match guard reported `all 649 changed lines are provably mcp-mesh-owned` and coverage reported clean — the line genuinely *is* mesh-owned, just rewritten to a version that does not exist. That is the same failure shape as the v3.3.1 Java publish break: a green bump that fails at publish time.

Validation: `23 checks passed` (`python3` and `pytest`), +2 tests covering non-match of `2.8.0rc1` / `2.8.00` / `==2.8.0.post1` plus positive cases so they can't pass by matching nothing. Clean 3.3.1 → 3.3.2 bump in a throwaway worktree: 470 files / 37 categories, both guards green, all four pip pins still rewritten (confirming the boundary doesn't over-block). `go build ./...` and `go test ./src/core/cli/scaffold/...` pass; re-rendered man output carries the new wording.

Follow-up not taken, for a separate issue: since neither guard can see this class, the only regression net is `test_bump_version.py`, and the boundary is copy-paste for any handler added later. A test asserting every `version_format="pep440"` handler's pattern ends in the boundary would make that structural.

PR #1406, opened 2026-07-29
fix(release): stop bump_version.py rewriting dated provenance claims

### LINKED ISSUE #1405: bump_version.py rewrites documentation provenance claims ('Since vX') — three sentences have ratcheted since v3.0.1 and now misdate a v2.2.1 behaviour
## Summary

`scripts/bump_version.py` rewrites **provenance statements** in docs — sentences of the form *"Since vX, <behaviour>"* that record **when a behaviour was introduced**. Bumping them is always wrong: the version is a historical fact, not a coordinate that tracks the current release.

Three sentences across two files have been ratcheted on **every release since v3.0.1**. They now claim a behaviour landed in the release being cut, when it actually shipped in **v2.2.1** (`bcf698d8a`, *"cross-loop affinity fix — lifespan + tools on single user loop"*, #1062).

Found during 3.3.2 release prep. The bump was reverted; the release is paused on this.

## The corrupted claims

**`docs/concepts/stateful-agents.md`** (2 sites)

> Because **v3.3.1** runs your `lifespan` AND your tools on the same user loop, the canonical FastMCP idiom is straightforward

> `lifespan` exit. Since **v3.3.1**, `lifespan` and all tool bodies share the single user loop

**`docs/python/dependency-injection.md:300`** (1 site)

> Since **v3.3.1**, default tool dispatch runs on a single-user loop (was `min(8, max(2, cpu_count()))`)

All three describe the same v2.2.1 change. The ratchet is visible in the history — each release commit walks them forward one version:

```
05ed82d3b release: v3.1.0    Since v3.0.1 -> v3.1.0
aa364fc6d release: v3.2.0    Since v3.1.0 -> v3.2.0
9032eeb75 release: v3.2.1    Since v3.2.0 -> v3.2.1
013d9ad72 release: v3.2.2    Since v3.2.1 -> v3.2.2
0a51c6452 release: v3.2.3    Since v3.2.2 -> v3.2.3
20fb519b3 release: v3.3.0    Since v3.2.3 -> v3.3.0
b4ad7ae99 release: v3.3.1    Since v3.3.0 -> v3.3.1
```

The reader is told a behaviour is new in whatever version they happen to be reading about — which is both false and useless, since it can never indicate an actual upgrade boundary.

## Why both guards missed it

**The over-match guard has an explicit exemption for one of the files** (`bump_version.py`, `EXEMPTIONS`):

```python
Exemption(
    glob="docs/concepts/stateful-agents.md",
    pattern=r"\bvNEW\b",
    reason=("narrative prose ('Since vX, lifespan and all tool bodies share "
            "one loop') dating a runtime behavior change. ..."),
)
```

The guard flagged this prose correctly. The exemption — added in #1395 — silenced it. The reason text even describes the sentence accurately as *"dating a runtime behavior change"*, which is precisely the signal that it must **not** be bumped. The conclusion drawn was "the guard is wrong here" when it should have been "the handler is wrong here".

**`docs/python/dependency-injection.md` is not exempted at all** and still passed, because the guard treats a line as mesh-owned when a mesh token appears within the ±3-line context window, and this one sits in mesh-related prose.

So the guard has two independent blind spots for this class, and neither the coverage guard nor the over-match guard can see it: the line genuinely is mesh-related, it just states a fact the bump makes false.

## Also: a stale release date on the docs landing page

`docs/index.md:415` reads:

> - **Latest Release**: v3.3.2 **(March 2026)**

The version is bump-managed; the date is not. It has read "March 2026" since at least v3.0.0, across every release through July 2026. Either the date should be bump-managed alongside the version, or it should be removed — a hand-maintained date next to an auto-maintained version will always drift.

## Proposed fix

1. **Correct the three sentences to `v2.2.1`.**
2. **Stop bumping provenance prose.** Replace the `stateful-agents.md` exemption with an actual exclusion in the handler, so `Since vX` / `Because vX` / `as of vX` sentences are never rewritten in any doc. An exemption tells the guard to stop complaining; the handler still corrupts the file. The exclusion has to live in the matcher.
3. **Resolve the stale date** — bump-manage it or drop it.
4. **Sweep `docs/` for other provenance prose** that the handler would rewrite, and confirm each surviving match is a genuine "current version" coordinate rather than a historical claim.
5. **Add a regression test** asserting a `Since vX` sentence survives a bump untouched.

## Acceptance

- Version-provenance sentences state the version the behaviour actually shipped in, and survive a bump unchanged.
- No exemption silences a guard hit that indicates a genuine mis-rewrite.
- The docs landing page does not display a release date that contradicts the release.

dhyansraj:
## Summary

`bump_version.py` rewrites **dated provenance claims** in docs — sentences of the form *"Since vX, <behaviour>"* that record when a behaviour shipped. Bumping them is always wrong: the version is a historical fact, not a coordinate tracking the current release.

Four sentences have been ratcheted forward on every release for months. Found during 3.3.2 release prep, which is paused on this.

## Corrections

| file | claimed | actual |
|---|---|---|
| `docs/concepts/stateful-agents.md` (x2) | v3.3.1 | **v2.2.4** |
| `docs/python/dependency-injection.md:300` | v3.3.1 | **v2.2.4** |
| `docs/concepts/architecture.md:219` | v3.3.1 | **v1.0.0** |

The loop-topology change landed in `bcf698d8a`, tagged as "v2.2.1" — but v2.2.1/2.2.2/2.2.3 were aborted npm-publish attempts with no GitHub release and no `RELEASE_NOTES.md` entry. The notes document the whole fix under `## v2.2.4`, and the section headings *directly above* two of these sentences already read `## Loop topology (v2.2.4+)`, as do three sibling surfaces (`docs/environment-variables.md:72`, `man/content/environment.md:95`, `man/content/dependency-injection.md:47`).

Worth recording how the drift compounded: the sentences were written as `v2.2.1`, ratcheted to `v2.2.4` by the 2.2.2/2.2.3/2.2.4 bumps, and then `828c04f77` "corrected" the *headings* from `v2.2.1+` up to `v2.2.4+` — trusting the ratcheted prose. The ratchet fed a false version into hand-maintained text and reached the right answer by accident.

`architecture.md` was not in #1405. It was written as `v1.0.0` and has ratcheted longest of the four.

## Why neither guard caught this

The `architecture.md` line is the clearest case: it contains the literal string "MCP Mesh", so the over-match guard's ownership test clears it. **Ownership is not the question a dated claim raises** — the line genuinely is ours; the bump just makes it say something false.

`stateful-agents.md` was worse. The guard *did* flag it, and #1395 added an exemption to silence it:

```python
Exemption(glob="docs/concepts/stateful-agents.md", pattern=r"\bvNEW\b",
          reason="narrative prose ('Since vX, ...') dating a runtime behavior change. ...")
```

The reason text describes the sentence accurately as *"dating a runtime behavior change"* — precisely the signal it must not be bumped. A correct alarm was read as a false positive. An exemption only quiets the guard; the handler keeps corrupting the file.

## The fix

`PROVENANCE_PROSE` is attached as `line_excludes` on the `Documentation (vOLD)` handler — **in the matcher, not the guard**. The `stateful-agents.md` exemption is removed.

The veto tests **character overlap** with the matched version rather than line membership, which preserves the inverse case: *"Since v2.2.4 this is the default; upgrade to v3.3.1 to get it"* pins the claim and still bumps the coordinate. `**Latest Release**: vX` carries no dating phrase and is unaffected.

One pattern added beyond the issue's list: `(vX+)` availability markers, of which four exist. They are safe today only because `+` falls outside the handler's trailing character class — drop the `+` while editing a heading and the ratchet restarts. `vX or later` was deliberately **not** added; the only repo hits are Python language versions, where it is plausibly a real minimum-version coordinate.

**Stale date dropped.** `docs/index.md` has read `(March 2026)` since at least v3.0.0, through every release into July. Bump-managing it would make the script's output depend on the wall clock and untestable; `RELEASE_NOTES.md` and the releases page already date every release. `**Latest Release**: vX` now stands alone.

## Validation

Real `3.3.1 -> 3.3.2` bump in a throwaway worktree:

```
Summary: 467 files updated across 37 categories
✅ Over-match guard: all 645 changed lines are provably mcp-mesh-owned.
✅ Coverage guard: no stale mesh-shaped references to 3.3.1 remain.
```

Against a control run on pre-fix `f3eccf583` (470 files / 649 lines) the delta is exactly **−3 files / −4 lines**, and a set-diff of the changed-file lists shows only the three prose files stopped bumping, with **nothing newly bumping**. `**Latest Release**` still went to `v3.3.2`.

Since the sentences now name v2.2.4, a 3.3.1 bump would leave them alone trivially — so a `2.2.4 -> 9.9.9` sentinel was run to prove the exclusion is load-bearing rather than inert. Fixed tree: `git diff --stat` on `docs/` and man content is **empty**. The same sentinel on the baseline matcher corrupts all three v2.2.4 sentences to `v9.9.9`.

Tests: **27 checks passed** (`python3` and `pytest`). This required fixing an existing test that *encoded the bug* — `test_docs_v_prefix_handler_skips_docker_tags` asserted `MCP Mesh v3.3.1 adds a media pipeline` → `v3.4.0`, commented `# prose still updates`.

## Exemption audit

All 8 remaining entries are the opposite character — the rewrite is **correct but unprovable**, not wrong-and-silenced: two `workflow_dispatch` inputs (bare YAML scalars), `docs/index.md`'s Latest Release (no mesh token within ±3), two Python `__version__` constants (import package is `mesh`, not `mcp-mesh`), three tsuite package-version blocks (keys name artifacts by role). The distinction is encoded as `test_no_exemption_covers_bare_prose_versions`.

## Flagged, not fixed (different class)

`bump_test_documentation` does a blind `content.replace(old, new)` over three test READMEs — an unanchored global string replace, the loosest operation in the script. Harmless today (no provenance prose in those files), but it is the one remaining place where any `3.3.1`-shaped string is rewritten with no pattern at all.

Closes #1405

### DIFF
--- a/docs/concepts/architecture.md
+++ b/docs/concepts/architecture.md
@@ -216,7 +216,7 @@ Designed for containers and Kubernetes.
 
 ## Multimodal Pipeline
 
-MCP Mesh v3.3.1 adds a media pipeline that lets tools produce and LLMs consume binary content:
+MCP Mesh v1.0.0 adds a media pipeline that lets tools produce and LLMs consume binary content:
 
 ```
 Tool produces media
--- a/docs/concepts/stateful-agents.md
+++ b/docs/concepts/stateful-agents.md
@@ -48,7 +48,7 @@ mcp-mesh runs your agent across two event loops:
 
 The two-loop split means a long-running tool body (LLM call, slow registry hop, multi-minute MeshJob) holds the user loop, but never the framework loop — your K8s liveness/readiness probes stay responsive.
 
-**What this fixes**: loop-bound resources work as expected when created in `lifespan` startup. `asyncpg.Pool`, `redis.asyncio.Redis`, and `aiohttp.ClientSession` are all loop-affine — they bind to the asyncio loop that created them and fail if reused from a different one. Because v3.3.1 runs your `lifespan` AND your tools on the same user loop, the canonical FastMCP idiom is straightforward:
+**What this fixes**: loop-bound resources work as expected when created in `lifespan` startup. `asyncpg.Pool`, `redis.asyncio.Redis`, and `aiohttp.ClientSession` are all loop-affine — they bind to the asyncio loop that created them and fail if reused from a different one. Because v2.2.4 runs your `lifespan` AND your tools on the same user loop, the canonical FastMCP idiom is straightforward:
 
 ```python
 # Module-level — FastMCP's `lifespan` parameter receives a FastMCP server
@@ -104,7 +104,7 @@ long-running work. From mesh's perspective it's an ordinary CRUD
 agent — every tool call is short, idempotent where possible, and
 returns. The pool lives inside the agent process — created in
 `lifespan` startup, stored in a module-level global, and closed in
-`lifespan` exit. Since v3.3.1, `lifespan` and all tool bodies share
+`lifespan` exit. Since v2.2.4, `lifespan` and all tool bodies share
 the single user loop, so a module-level pool created in `lifespan`
 startup is the supported pattern. (FastMCP's `lifespan` receives a
 FastMCP server instance — there is no `.state` namespace as there is
--- a/docs/index.md
+++ b/docs/index.md
@@ -412,7 +412,7 @@ A `kubectl`-style command-line tool that follows you from first agent to product
 
 ## :star: Project Status
 
-- **Latest Release**: v3.3.1 (March 2026)
+- **Latest Release**: v3.3.1
 - **License**: MIT
 - **Languages**: Python 3.11+, TypeScript/Node.js 18+, and Java 17+ (runtime), Go 1.23+ (registry)
 - **Status**: Production-ready, actively developed
--- a/docs/python/dependency-injection.md
+++ b/docs/python/dependency-injection.md
@@ -297,7 +297,7 @@ loop — K8s liveness/readiness probes stay responsive.
 
 ### Default `MCP_MESH_TOOL_WORKERS=1`
 
-Since v3.3.1, default tool dispatch runs on a single-user loop (was
+Since v2.2.4, default tool dispatch runs on a single-user loop (was
 `min(8, max(2, cpu_count()))`). The canonical pattern for loop-bound
 resources works as expected — `lifespan` startup creates the resource
 on the user loop; every tool body uses it on the same loop; `lifespan`
--- a/scripts/bump_version.py
+++ b/scripts/bump_version.py
@@ -142,6 +142,123 @@ def record_changes(
 # ---------------------------------------------------------------------------
 
 
+# Sentences that DATE a behaviour — "Since v2.2.4, tool dispatch runs on a
+# single user loop", "MCP Mesh v1.0.0 adds a media pipeline". The version is a
+# historical fact recording when something shipped, not a coordinate that
+# tracks the current release, so rewriting it is always wrong.
+#
+# Left unprotected it also RATCHETS: every release walks the claim forward one
+# version, so it permanently reads "new in whatever you happen to be reading
+# about" and can never mark a real upgrade boundary. Four sentences ratcheted
+# this way for eight releases before #1405 caught them; one had been walking
+# since v1.0.0.
+#
+# Neither guard can catch this class. The line genuinely is mesh-related — it
+# is usually mesh prose, with "MCP Mesh" written out — so the over-match guard
+# clears it, and the coverage guard only looks for versions we MISSED. The fix
+# has to live here, in the matcher.
+#
+# Each pattern must span the version token itself: the exclusion is applied by
+# OVERLAP with the matched version (see `_sub_skipping_provenance`), not by
+# line. A line may carry both a provenance claim and a genuine current-version
+# coordinate, and only the claim is protected.
+# A version token as it appears in prose. `v` optional so a paren/hyphen form
+# still anchors when the handler's own `v` is the character being protected.
+_PV = r"v?\d+\.\d+(?:\.\d+)?"
+# Filler the lead-in phrases tolerate before the version. "Since MCP Mesh
+# v1.0.0 adds..." is the spelling docs/concepts/architecture.md actually uses,
+# and it was the longest-running corruption of the four.
+_PV_FILLER = r"(?:the\s+)?(?:mcp[-\s]?mesh\s+)?"
+
+PROVENANCE_PROSE: list[str] = [
+    # Lead-in form: a phrase introduces the version.
+    #   "Since v2.2.4, ..." / "(as of v1.4)" / "Prior to v1.4, ..."
+    #   "Since MCP Mesh v1.0.0, ..." / "As of the v3.4.0 release, ..."
+    # The `<verb> in` list mirrors the trailing verb list below — the two forms
+    # are the same claim with the version on the other side of the verb, so a
+    # verb covered in one direction and not the other is an arbitrary hole.
+    r"\b(?:since|because|as\s+of|new\s+in|prior\s+to|before|until"
+    r"|starting\s+(?:with|in|from)"
+    r"|available\s+(?:since|in|from)"
+    r"|(?:introduced|added|shipped|landed|released|removed|deprecated"
+    r"|changed|fixed|renamed|replaced|split|merged|gated|enabled|disabled"
+    r")\s+in"
+    r")\s+" + _PV_FILLER + _PV,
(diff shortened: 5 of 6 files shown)


dhyansraj:
Code review raised 3 WARNINGs. All three verified valid and fixed in `5bdfd7818`.

### 1. An over-veto would have shipped silently — coverage guard now closes the loop

The most important finding. This change trades a loud failure (a doc sentence reading the wrong version) for a silent one (a coordinate that quietly stops tracking), and nothing detected the latter: `_guard_patterns` had no pattern matching a bare `vOLD` in prose. So a future veto of the landing-page line would complete the bump, print green on both guards, and advertise the previous release.

Added `\*\*Latest Release\*\*:\s*v?OLD` to `_guard_patterns`, mirroring the `OVERMATCH_ALLOWLIST` entry that anchors the same line from the other side — the allowlist entry says the rewrite is *legitimate*, this one says it is *mandatory*.

Demonstrated firing rather than merely added, with the same deliberate over-veto injected in both runs:

```
=== with the new guard pattern ===
  EXIT CODE = 1
  ❌ Coverage guard: 1 stale mesh-shaped reference(s) to 3.3.1 survived the bump:
    docs/index.md:415: - **Latest Release**: v3.3.1

=== without it (over-veto retained) ===
  EXIT CODE = 0
  ✅ Over-match guard: all 644 changed lines are provably mcp-mesh-owned.
  ✅ Coverage guard: no stale mesh-shaped references to 3.3.1 remain.
  landing page -> - **Latest Release**: v3.3.1
```

The second block is the failure this PR would otherwise have introduced: exit 0, both guards green, wrong version on the docs landing page.

### 2. The `(vX+)` pattern was dead code

Confirmed: `+` sits outside the handler's trailing class `[\s,\)\]"']`, so `(v2.2.4+)` never produces a match for the veto to suppress. It read as protection, and the tests around it passed vacuously. Its comment also argued for a scenario ("drop the `+` while editing a heading") that it stopped covering the moment the `+` was dropped.

Replaced with the parenthesised-label form plus a migration-arrow rule for the right-hand side of `(v1.3 → v1.4)`, which no paren precedes. Both are genuinely exposed today — `)` and `,` are in the trailing class — at `src/core/cli/man/content/headers.md:51` and `:62`, safe only because those versions are historical.

### 3. Lead-in list missed the spelling our own docs use

`Since MCP Mesh vX` escaped the pattern, which is exactly how `docs/concepts/architecture.md:219` is written — the file with the longest-running corruption of the four. Added an optional `(?:the\s+)?(?:mcp[-\s]?mesh\s+)?` filler, mirrored the trailing verb list into the lead-in form (`removed|deprecated|changed|fixed|released|renamed ... in`), and added `pre-`/`post-vX`.

`vX or newer` / `vX and later` were included after initially being drafted as a deliberate omission. The omission was going to rest on the form being ambiguous between an availability claim and a minimum-requirement floor — but both readings argue against bumping. As an availability claim it is provenance outright; as a requirement floor, auto-bumping would assert that every release raises the floor. No reading makes the current release the right answer, so there was no ambiguity to weigh. The reasoning is recorded in the code.

**INFO 1** fixed (the two vacuous `(v3.3.1+)` assertions now bite via the paren rule). INFO 2/3/4 left alone as the review scoped them.

### Validation

Real `3.3.1 -> 3.3.2` bump, throwaway worktree — **identical to the pre-review baseline**, so the widened patterns cost nothing:

```
Summary: 467 files updated across 37 categories
✅ Over-match guard: all 645 changed lines are provably mcp-mesh-owned.
✅ Coverage guard: no stale mesh-shaped references to 3.3.1 remain.
```

Set-diff against a pre-fix control: the only files that stopped bumping are still the three prose files, with **nothing newly frozen**. `**Latest Release**` still moved to `v3.3.2`.

Over-reach check, re-run after widening: every `v\d+\.\d+` under `docs/**/*.md` and `src/core/cli/man/content/**/*.md` enumerated and classified. Exactly **one live coordinate** exists (`docs/index.md:415`), it carries no parentheses and no dating phrase, and it still bumps. Everything else is a historical claim, a reader-owned artifact behind the `(?<![/:])` lookbehind, or an `A2A v1.0` protocol reference that is structurally unmatchable.

Sentinels: `2.2.4 -> 9.9.9` leaves `docs/` and man content entirely unchanged. A new sentinel exercising all 21 widened phrasings alongside a live coordinate froze all 21 dated claims and bumped only the coordinate.

Tests: **30 checks passed** (`python3` and `pytest`). With `PROVENANCE_PROSE` emptied, exactly four bite; the two that do not are the wiring and coverage-guard tests, orthogonal to the veto contents by design.

Issue #1407, opened 2026-07-29
bump_version.py reminder tells operators to run 'cargo generate-lockfile', which silently re-resolves the whole Rust dependency graph


dhyansraj:
## Summary

After a version bump, `scripts/bump_version.py` prints:

```
Reminder: run 'cargo generate-lockfile' in src/runtime/core to refresh Cargo.lock with the new mcp-mesh-core version
```

`cargo generate-lockfile` **re-resolves the entire dependency graph from scratch**, so it does not merely refresh our own version — it silently advances every transitive crate to its newest compatible release. Following the reminder as written turns a version bump into an unreviewed dependency bump.

## Observed on the 3.3.2 prep

`cargo generate-lockfile` after a 3.3.1 -> 3.3.2 bump changed **9** version lines: ours plus 8 third-party crates.

| crate | from | to |
|---|---|---|
| `napi` | 3.11.0 | 3.12.0 |
| `napi-derive` | 3.6.0 | 3.6.1 |
| `napi-derive-backend` | 6.0.0 | 6.1.0 |
| `napi-build` | 2.3.2 | 2.4.0 |
| `cc` | 1.3.0 | 1.4.0 |
| `ctor` | 1.0.10 | 1.0.11 |
| `displaydoc` | 0.2.6 | 0.2.7 |
| `either` | 1.16.0 | 1.17.0 |

The `napi*` chain is the Node-API binding layer for the TypeScript runtime's native module, and `cc` is the C compiler driver — these change the compiled artifact, not just metadata.

The correct command is targeted:

```
cargo update --package mcp-mesh-core
```

which reports `Locking 1 package to latest compatible version` and produces a **one-line** diff.

## This has already shipped, more than once

| release | third-party crate versions changed in `Cargo.lock` |
|---|---|
| v3.2.2 -> **v3.2.3** | 6 |
| v3.3.0 -> **v3.3.1** | 6 |
| v3.2.3 -> v3.3.0 | 0 |

v3.3.1 is the sharpest case. Its release notes state:

> This is a **release-infrastructure patch — no runtime, wire, registry, resolution, or declaration-syntax changes.** Runtime behavior is byte-for-byte identical to 3.3.0.

[...]

Issue #1402, opened 2026-07-29
Python: declared fastapi floor (>=0.104.0) is unresolvable against the package's own mcp/fastmcp pins


dhyansraj:
## Summary

Both Python manifests declare `fastapi>=0.104.0`, but **FastAPI below 0.133.0 cannot be installed alongside the package's own pinned dependencies**. The declared floor is roughly 30 minor versions below what the package can actually resolve, so it is a compatibility claim we cannot honour.

## Why it is unreachable

```
mcp==1.28.1        -> anyio>=4.5              vs  fastapi 0.104.0 -> anyio<4.0.0
fastmcp==3.4.3     -> fastmcp-slim[server]    ->  starlette>=1.0.1
fastapi <0.133.0   -> starlette<1.0.0             (cap lifted in 0.133.0)
```

Verified empirically:

- `pip install mcp==1.28.1 fastmcp==3.4.3 fastapi==0.132.1` -> `ResolutionImpossible`
- `pip install mcp==1.28.1 fastmcp==3.4.3 fastapi==0.133.0` -> resolves

Anyone pinning `fastapi<0.133` today already hits a hard resolver error. The floor is not protecting a working configuration; it is advertising one that does not exist.

## This is not caused by the route rework

Worth stating explicitly, since the floor was examined while validating #1397/#1398: **the new route code works at every version tested, 0.104.0 through 0.140.13.** A 7-scenario end-to-end sweep across 15 FastAPI versions (direct DI route, `include_router` DI route, direct SSE, `include_router` SSE with a warm cache, registration-options round-trip, nested effective paths, introspection surface) passed 7/7 at every version, with **no silent degradation anywhere** — SSE returned `text/event-stream` with the expected body throughout.

- `_rebuild_api_route` reads `APIRoute.__init__` by signature. Every parameter is optional-with-default in every version and every one round-trips as a route attribute; the 0.104 set (24) is a strict subset of the 0.140 set (25, `strict_content_type` added at 0.132.1). Runtime probe reports `skipped=[]`.
- `iter_app_routes` no-ops cleanly where `original_router` is absent, which is the correct behaviour for versions that flatten `include_router`. `invalidate_route_caches` likewise no-ops, getattr-guarded.
- Remaining production FastAPI imports are only `FastAPI`, `Response`, `Request`, `UploadFile`, `JSONResponse`, `StreamingResponse`, `PlainTextResponse`, `APIRoute` — all long-predating 0.104. No private internals remain.

[...]

PR #1408, opened 2026-07-29
release: v3.3.2

dhyansraj:
Patch release **v3.3.2**.

The headline is a Python `@mesh.route` fix. **3.3.1's declared FastAPI range resolves to versions where routes are silently broken** — SSE endpoints degrade to `application/jsonl` instead of `text/event-stream`, and handlers mounted with `include_router()` are never discovered, so they serve without mesh dependency injection. Neither failure is loud: startup succeeds, health checks pass, logs report success. Anyone installing 3.3.1 today gets this.

Full notes in `RELEASE_NOTES.md`.

## Contents

| Issue | Change |
|---|---|
| #1387, #1389 | Route rebuild moved to the public `APIRoute` constructor; a rebuild failure now aborts startup instead of being swallowed |
| #1396 | `@mesh.route` handlers mounted via `include_router()` discovered again |
| #1402 | FastAPI floor raised to `>=0.135.0` — `>=0.104.0` was unresolvable against the pinned `mcp`/`fastmcp` |
| #1385 | Published manifest reconciled with source |
| #1383 | LiteLLM no longer imported for big-3 vendors; `mcp-mesh[litellm]` extra added (additive — no user action) |
| #1392 | `meshctl man` code-span corruption and unstyled list items |
| #1394 | Over-match guard for `bump_version.py` |
| #1405 | `bump_version.py` no longer rewrites dated provenance claims |
| #1384 | DI positional-contract docs restated; Java examples fixed |

All already merged to `main`; this PR is the version bump + release notes.

## Mechanics

- Version bump **3.3.1 → 3.3.2** via `scripts/bump_version.py` — 467 updates across 37 categories, 645 changed lines, 451 files.
- ✅ Over-match guard: all 645 changed lines provably mcp-mesh-owned.
- ✅ Coverage guard: no stale references to 3.3.1 remain.
- Independent mechanical review of the diff: 19 residual suspects, all legitimate (our own `workflow_dispatch` inputs, `mesh.__version__`, tsuite package tables, and the `**Latest Release**` coordinate). No prefix-corruption artifacts.
- `helm dependency update helm/mcp-mesh-core` → `Chart.lock`, our six subcharts only.
- `cargo update --package mcp-mesh-core` → `Cargo.lock`, **one line**.

### ⚠️ Note on `Cargo.lock`

The script's post-bump reminder says to run `cargo generate-lockfile`. That re-resolves the **entire** dependency graph — running it here advanced 8 third-party crates alongside ours, including the whole `napi` chain (the Node-API bindings for the TypeScript runtime's native module) and `cc`.

This has shipped before: **v3.2.3 and v3.3.1 each moved 6 third-party crates**, and v3.3.1's notes claimed *"Runtime behavior is byte-for-byte identical to 3.3.0"* while its native module was rebuilt against a different dependency set. Nothing surfaces it — the churn hides in a 450-file diff and both guards pass, since they only inspect lines carrying the mesh version.

This release uses the targeted `cargo update --package mcp-mesh-core`. The reminder fix is tracked in **#1407** and deliberately kept out of this PR so the release stays bump-plus-notes.

## Reviewer notes

- **`include_router` version.** The notes say FastAPI stopped flattening included routers at **0.137.0**. PR #1398's body says "around 0.139.x" — that figure was superseded by #1403, which corrected it across 8 sites and verified it live (`original_router` absent at 0.136.3, present at 0.137.0). The note is right; the PR it cites is the stale one.
- **Dependency "before" values** in the Notes section are from the **published** manifest at `v3.3.1`, not the source manifest. The two had drifted (that drift is #1385), and `pip install mcp-mesh` resolves against the published one. An editable dev install saw none of those four changes.
- **FastAPI upper bound is unchanged.** The `<0.140.5` ceiling was added and removed inside this release window and never shipped; the declared range differs from 3.3.1 only by the floor.

## Test plan

- [ ] Tag pushed and inspected before the publish workflow is fired
- [ ] Publish workflow green across PyPI / npm / crates / Maven Central / Docker
- [ ] Post-release: refresh mesh-expert and skybot

### DIFF
--- a/.github/workflows/helm-release.yml
+++ b/.github/workflows/helm-release.yml
@@ -7,9 +7,9 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: "Version tag (e.g., v3.3.1)"
+        description: "Version tag (e.g., v3.3.2)"
         required: true
-        default: "v3.3.1"
+        default: "v3.3.2"
 
 jobs:
   release:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,9 +8,9 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: "Version (e.g., v3.3.1)"
+        description: "Version (e.g., v3.3.2)"
         required: true
-        default: "v3.3.1"
+        default: "v3.3.2"
       environment:
         description: "Environment"
         type: choice
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,10 +1,55 @@
 # MCP Mesh Release Notes
 
-[Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.3.1...HEAD)
+[Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.3.2...HEAD)
+
+[Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.3.1...v3.3.2)
 
 [Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.3.0...v3.3.1)
 
 [Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.2.3...v3.3.0)
+
+## v3.3.2 (2026-07-28)
+
+A patch release, and an important one for Python `@mesh.route` users: **v3.3.1's declared FastAPI range resolves to versions where routes are silently broken.** On current FastAPI, SSE routes degrade to `application/jsonl` instead of `text/event-stream`, and handlers mounted with `include_router()` are never discovered at all, so they serve without mesh dependency injection. Neither failure is loud — startup succeeds, health checks pass, and the logs still claim success. Both are fixed here, alongside a published-manifest reconciliation and the first half of LiteLLM's move to an optional install. No wire, registry, resolution, or declaration-syntax changes.
+
+> **⚠️ Upgrade if you use `@mesh.route` on Python.** The declared FastAPI range is unchanged from 3.3.1 apart from a floor raise — what changed is that the code now works across it. Route integration is also fail-fast now, so an agent that started on 3.3.1 while serving a degraded route can refuse to start on 3.3.2; stage the rollout and see Notes.
+
+### 🌐 HTTP routes
+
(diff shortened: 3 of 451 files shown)


dhyansraj:
Release-notes review complete (our review-agent — CodeRabbit skips this PR automatically: *"449 files exceed the limit of 100"*, which means release PRs have never been auto-reviewed).

**0 BLOCKER, 3 WARNING, 4 INFO.** All applied in `5cf5dbdb4`.

The one that mattered — **route integration is now fail-fast, and the upgrade guidance didn't say so.** `RouteIntegrationStep` is `required=True` and `RouteRebuildError` now propagates in three cases beyond the swallowed-`ImportError` one. So an agent that started fine on 3.3.1 while serving a silently degraded route can **refuse to start** on 3.3.2 — a rolling upgrade turns a wrong content-type into a crash loop. The bullets said this; the ⚠️ callout and Notes did not, and the Notes closed on *"No wire, registry, resolution, or declaration-syntax changes"* — accurate, but blanket reassurance exactly where a reader decides whether to stage the rollout. Now called out in both places, positioned above that closer.

Also fixed:

- **"floor job added to CI" was false.** CI pins `fastapi==0.136.1`; the declared floor is `0.135.0`. We'd have advertised the floor as tested when it isn't — the same supported-on-paper/untested-in-CI shape #1402 fixed, reintroduced one layer up. Reworded to describe the job accurately. Moving the pin is #1409.
- **`orjson` was missing from the fresh-install note** — three of the four listed items are floor moves; `orjson` is the only one that adds a package. Added, with the published-vs-source manifest framing now stated explicitly.
- Dangling "ceiling drop above" reference (residue from an earlier draft order), and `jsonschema`'s import site corrected to `mesh/helpers.py`.

**Verified clean against the tree**, no action: the `[litellm]` extra genuinely requires nothing today (byte-identical pin in base and extra, enforced by `test_litellm_extra_pin_sync.py`); all four dependency floor transitions; the unchanged `<1.0.0` upper bound with no `0.140.5` residue anywhere; "zero private imports" literally true; "nine patterns anchored" counted to exactly nine; ten issue references for ten merged PRs.

Follow-ups filed as **#1409**: move the CI pin to the declared floor and assert it against the manifest so they can't drift; a 9th `0.139`→`0.137` site #1403 missed at `ci.yml:289`; and `renderer_test.go` logging its corpus counts via `t.Logf` rather than asserting them — so the 83/432 figures now shipping in these notes have no in-tree anchor.

PR #1410, opened 2026-07-29
fix(helm): add Deployment strategy passthrough to agent, grafana, tempo, redis

### LINKED ISSUE #1381: mcp-mesh-agent: no Deployment strategy passthrough — agents with ReadWriteOnce persistence deadlock on every upgrade
## Summary

`mcp-mesh-agent` renders a `Deployment` with no `strategy` block and exposes no values key to set one. Any agent using `persistence` with a `ReadWriteOnce` volume therefore **deadlocks on every `helm upgrade`** and requires manual intervention to recover.

Chart version checked: **3.3.1** (`oci://ghcr.io/dhyansraj/mcp-mesh/mcp-mesh-agent`).

## Why it deadlocks

With the default `RollingUpdate` strategy, Helm creates the new ReplicaSet before terminating the old pod. The old pod still holds the RWO PersistentVolumeClaim, so the new pod cannot mount it and stays `Pending` / `ContainerCreating`. The rollout never completes, so the old pod is never terminated. Neither side can make progress.

This is the classic RWO + `RollingUpdate` interaction, and it is normally solved with `strategy: Recreate`.

## Chart specifics

- `templates/deployment.yaml` — no `strategy` block at all, and no `.Values` reference for one.
- `values.yaml` — the only `maxUnavailable` in the file is line 352, a commented-out key under `podDisruptionBudget`, unrelated to Deployment strategy.
- `values.yaml:313` — **`persistence.accessMode` defaults to `ReadWriteOnce`.**

That last point matters: this is not an exotic configuration. Anyone who sets `persistence.enabled: true` and keeps the chart's own defaults will hit this on their second upgrade.

## Reproduce

1. Deploy an agent with `persistence.enabled: true` (or `persistence.existingClaim`) on an RWO storage class.
2. `helm upgrade` with any change to the pod template — a new image tag is enough.
3. The new pod hangs unable to mount the volume; the old pod is never evicted.

## Workarounds, and why they are unsatisfying

- **`kubectl scale --replicas=0` then back to `1`** — works, but is manual, causes downtime, and defeats the purpose of a rolling upgrade.
- **Argo CD `Replace=true`** — does not help. It still performs an update rather than a delete-and-recreate of the Deployment.
- **Switch the volume to `ReadWriteMany`** — viable only when the storage class supports it, and only for content that tolerates concurrent mounts.

None of these is available to a user who only consumes the chart through values.

## Proposed fix

Add a `strategy` passthrough to `templates/deployment.yaml`:

```yaml
spec:
  {{- with .Values.strategy }}
  strategy:
    {{- toYaml . | nindent 4 }}
  {{- end }}
```

and document it in `values.yaml`:

```yaml
# Deployment update strategy. Set to Recreate when the agent mounts a
# ReadWriteOnce volume — RollingUpdate deadlocks because the incoming pod
# cannot mount a claim the outgoing pod still holds.
# strategy:
#   type: Recreate
```

Gating on `with` keeps the rendered output byte-identical for every chart consumer who does not set it, so this is additive and safe to backport.

Optionally, the chart could default `strategy.type` to `Recreate` when `persistence.enabled` is true and `accessMode` is `ReadWriteOnce`, since that combination cannot work any other way — though an explicit values key is probably clearer than inferred behaviour.

## Context

Found while upgrading a production mesh application whose RAG agent mounts an RWO volume for a vector store. Happy to open a PR for the `strategy` passthrough if that shape is agreeable.

dhyansraj:
## Summary

Adds a `with`-gated Deployment `strategy` passthrough so an agent (or bundled component) mounting a `ReadWriteOnce` volume can select `Recreate`. With the default `RollingUpdate`, the incoming pod cannot mount a claim the outgoing pod still holds, so the rollout never completes and the old pod is never terminated — neither side can progress.

Purely additive: 106 insertions, 0 deletions, all defaults `{}`.

## Wider than the issue as filed

#1381 reports this on `mcp-mesh-agent`, where `persistence` is **opt-in**. Two bundled charts have the same defect with persistence **enabled by default**, and both are in the standard `mcp-mesh-core` install path:

| chart | strategy | RWO PVC | replicas |
|---|---|---|---|
| `mcp-mesh-agent` | none | opt-in (`values.yaml:313`) | `replicaCount` |
| `mcp-mesh-grafana` | none | **default-on** (`values.yaml:27-32`) | `1`, hardcoded |
| `mcp-mesh-tempo` | none | **default-on** (`values.yaml:31-32`) | `1`, hardcoded |
| `mcp-mesh-redis` | none | latent (`emptyDir` unless opted in) | — |

So a default core install already ships two workloads that deadlock on any pod-template change. All four get the key.

`mcp-mesh-registry` (hardcoded strategy, stateless) and `mcp-mesh-postgres` (StatefulSet — ordered replacement makes RWO safe) are deliberately untouched.

## Why `Recreate` is NOT the default for grafana/tempo

This was implemented as a default first, on the reasoning that `RollingUpdate` + `replicas: 1` + an RWO PVC can never complete, so the shipped default is already broken. Probing the transition against a real 3-node k3s cluster (v1.34.3) reversed it.

Helm's 3-way merge cannot clear the API-server-defaulted `rollingUpdate` block:

```
Error: UPGRADE FAILED: Deployment.apps "probe" is invalid:
spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is 'Recreate'
```

| path | result |
|---|---|
| `helm upgrade`, Helm 4 default (server-side apply) | **fails** |
| `helm upgrade --force-conflicts` | **fails** — SSA still won't clear the field |
| `helm upgrade --force` | rejected: cannot combine SSA with force replace |
| `helm upgrade --server-side=false` (Helm 3 patch path) | succeeds → `{"type":"Recreate"}` |
| `kubectl apply` (client-side, `retainKeys`) | succeeds |
| rendering `rollingUpdate: null` | no effect — `toYaml` drops null keys |

It fails whether or not the prior manifest carried an explicit `strategy`, so this is not a one-time artifact of this change.

That inverts the trade. Defaulting `Recreate` would turn a **conditional** deadlock (multi-node, and only on a pod-template change) into an **unconditional** failure of the entire `mcp-mesh-core` upgrade for every Helm 4 consumer — including those running grafana/tempo with persistence disabled, who have no exposure at all. The `--server-side=false` migration path is documented in the values comments instead.

Secondary finding: a strategy-only change does **not** roll pods — `.spec.template` is unchanged, and the ReplicaSet count stayed at 1 across the attempted upgrade. The cost of defaulting it was never a rollout; it was the API rejection.

## Validation

- **`helm lint`** — all four charts plus `mcp-mesh-core`, 0 failed.
- **Default renders byte-identical to `dcbcdbd3e`** for all four charts *and* the umbrella.
  Baseline taken from a throwaway `git worktree` with its own `helm dependency update` — necessary because `helm/*/charts/*.tgz` is gitignored, so `git stash` does not revert it. A first attempt contaminated by stale `.tgz` files made a real delta look empty. Grafana's `admin-password` and postgres's `password` churn on every render independently of this change (confirmed by two consecutive renders of unmodified `HEAD`), so both were pinned via `--set` for a clean diff.
- **Overrides render at the correct indent** in all four charts, including nested `rollingUpdate: {maxSurge, maxUnavailable}` maps.
- **Umbrella flow-through** verified through the umbrella's own values file (`mcp-mesh-grafana.grafana.strategy`, `mcp-mesh-tempo.tempo.strategy`, `mcp-mesh-redis.strategy`), not just `--set` on subcharts.
- **`python3 scripts/check_helm_pss.py`** — all 9 charts OK.
- Rendered manifests additionally passed `kubectl apply --dry-run=server` against a live cluster with `Recreate` set.

## Known gap, not closed here

The grafana/tempo **defaults are still wrong out of the box** on multi-node clusters — this PR supplies the escape hatch, not the fix. The clean resolution is likely converting both to StatefulSets, as `mcp-mesh-postgres` already is, since ordered replacement makes RWO safe without the strategy-transition trap. That is a resource-kind change with a PVC ownership migration, so it needs its own issue and upgrade note. Filed separately.

Closes #1381

## Summary by CodeRabbit

* **New Features**
  * Added configurable Kubernetes Deployment update strategies for the agent, Grafana, Redis, and Tempo components.
  * Supports choosing `RollingUpdate` or `Recreate`, including guidance for persistent volumes using `ReadWriteOnce` storage.
  * Preserves existing default rollout behavior when no strategy is specified.

* **Documentation**
  * Added configuration guidance and rollout considerations to the Helm chart values and README files.

### DIFF
--- a/helm/mcp-mesh-agent/README.md
+++ b/helm/mcp-mesh-agent/README.md
@@ -102,6 +102,7 @@ helm uninstall my-agent -n mcp-mesh
 | Parameter                   | Description                            | Default                    |
 | --------------------------- | -------------------------------------- | -------------------------- |
 | `replicaCount`              | Number of replicas                     | `1`                        |
+| `strategy`                  | Deployment update strategy (unset = Kubernetes default `RollingUpdate`; set `{type: Recreate}` for ReadWriteOnce `persistence`) | `{}` |
 | `image.repository`          | Container image repository             | `"mcpmesh/python-runtime"` |
 | `image.pullPolicy`          | Image pull policy                      | `IfNotPresent`             |
 | `image.tag`                 | Image tag (overrides chart appVersion) | `"0.9"`                    |
@@ -109,6 +110,12 @@ helm uninstall my-agent -n mcp-mesh
 | `resources.limits.memory`   | Memory limit                           | `1Gi`                      |
 | `resources.requests.cpu`    | CPU request                            | `100m`                     |
 | `resources.requests.memory` | Memory request                         | `256Mi`                    |
+
+> **`strategy` and ReadWriteOnce volumes:** with the default `RollingUpdate`, Kubernetes starts the incoming pod before the outgoing one is gone. If the scheduler places it on another node it cannot mount a ReadWriteOnce claim the outgoing pod still holds, and the rollout deadlocks until someone deletes the old pod. Both pods landing on one node is why single-node dev clusters (kind, minikube, Docker Desktop) never see this. Set `strategy: {type: Recreate}` whenever `persistence.enabled` is true with the default `accessMode: ReadWriteOnce`.
+>
+> `Recreate` costs a gap in service while the old pod terminates, and when `autoscaling.enabled` is true that gap covers every replica at once, since `Recreate` tears the whole set down before scheduling any replacement. Note that `persistence` creates a **single** claim for the whole release, so ReadWriteOnce is not compatible with more than one replica under either strategy: with `replicaCount > 1` or an active HPA, every replica scheduled off the claim's node stays `Pending`. `Recreate` does not fix that combination — use ReadWriteMany, or keep the agent at one replica.
+>
+> Set it at install time if you can. Switching an **already-installed** release from `RollingUpdate` to `Recreate` is rejected by the API server — it defaulted `.spec.strategy.rollingUpdate` on the live object, and `rollingUpdate` "may not be specified when strategy `type` is 'Recreate'". Under **Helm 4** the default server-side apply does not clear it and `--force-conflicts` does not help, so run that one switch with `helm upgrade --server-side=false`; **Helm 3** patches client-side and needs no flag. Deleting the Deployment and letting Helm recreate it also works.
 
 > **Upgrade note:** `podSecurityContext` (runAsNonRoot, uid/gid 999, fsGroup 999, RuntimeDefault seccomp) is now always applied — it was previously gated behind the removed `podSecurityPolicy.enabled` flag and never rendered. It covers user-supplied `initContainers` (an init container that must run as root needs its own `securityContext`, which overrides the pod level), and `fsGroup: 999` changes group ownership of mounted volumes, including PVCs from `extraVolumes`.
 
--- a/helm/mcp-mesh-agent/templates/deployment.yaml
+++ b/helm/mcp-mesh-agent/templates/deployment.yaml
@@ -19,6 +19,10 @@ spec:
   {{- if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.replicaCount }}
   {{- end }}
+  {{- with .Values.strategy }}
+  strategy:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   selector:
     matchLabels:
       {{- include "mcp-mesh-agent.selectorLabels" . | nindent 6 }}
--- a/helm/mcp-mesh-agent/values.yaml
+++ b/helm/mcp-mesh-agent/values.yaml
@@ -134,6 +134,35 @@ autoscaling:
   targetCPUUtilizationPercentage: 80
   targetMemoryUtilizationPercentage: 80
 
+# Deployment update strategy. Empty leaves the field unset, so Kubernetes
+# applies its own default (RollingUpdate) — right for the stateless default.
+# Set Recreate when the agent mounts a ReadWriteOnce volume (see persistence
+# below, whose accessMode defaults to ReadWriteOnce). RollingUpdate starts the
+# incoming pod before the outgoing one is gone; if the scheduler places it on
+# another node it cannot mount a ReadWriteOnce claim the outgoing pod still
+# holds, and the rollout deadlocks until someone deletes the old pod. Both
+# pods landing on one node is why single-node dev clusters never see this.
+#
+# Recreate trades that for a gap in service while the old pod terminates, and
+# with autoscaling.enabled the gap covers every replica at once, because
+# Recreate tears the whole set down before scheduling any replacement. Note
+# that persistence creates a single claim for the whole release, so
+# ReadWriteOnce is not compatible with more than one replica under either
+# strategy: with replicaCount > 1 or an active HPA, every replica scheduled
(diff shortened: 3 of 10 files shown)


dhyansraj:
Review applied in `59366ef73` — 0 BLOCKER, 1 WARNING, 5 INFO, all addressed. Still purely additive: 144 insertions, 0 deletions, and every changed file this round is comments only.

**WARNING — the redis guidance pointed at a path the chart cannot execute.** `helm/mcp-mesh-redis/templates/` contains only `_helpers.tpl`, `deployment.yaml` and `service.yaml` — no `pvc.yaml`, no `PersistentVolumeClaim` anywhere — while `deployment.yaml:89-96` mounts `claimName: {{ include "mcp-mesh-redis.fullname" . }}` under `persistence.enabled`. An operator following the original comment would set `persistence.enabled: true`, get a pod Pending on a claim nothing creates, and never reach the deadlock the advice was protecting against. The prose now describes the emptyDir reality and frames the key as a passthrough for a volume you have wired yourself. The missing PVC is pre-existing and filed as **#1412**.

**The deadlock claim was overstated in two charts.** RWO is node-scoped, so "every upgrade deadlocks" is false on single-node clusters — a reader testing on kind would have concluded the warning was wrong and ignored it. One canonical sentence now appears in all four charts, verified identical by hash rather than by eye, since drift between hand-maintained near-copies is what produced the inconsistency. Grafana and tempo extend it with a chart-specific consequence (two Grafana processes writing one SQLite database).

**The `--server-side=false` advice would have misfired for most readers.** `helm/mcp-mesh-agent/README.md:8` states Helm 3.2.0+ and `helm/README.md:34` states 3.8+ — and `helm upgrade` on Helm 3 has no `--server-side` flag, so a Helm 3 operator would hit an unknown-flag error and conclude the migration was blocked. All four now say plainly that this applies to Helm 4 and that Helm 3 patches client-side and needs no flag.

**`autoscaling` + RWO no longer reads as workable.** `templates/pvc.yaml:5` creates a *single* claim per release, so with `replicaCount > 1` or an active HPA every replica off the claim's node stays Pending under **either** strategy. The docs previously framed `Recreate` as a bigger-gap tradeoff there, which implied the combination works.

**Umbrella discoverability.** `mcp-mesh-core` ships grafana (2Gi) and tempo (5Gi) RWO persistence on by default but never mentioned `strategy`, and an umbrella user never opens subchart values files. Commented pointers with full dotted paths now sit directly under those persistence blocks.

**Skipped INFO-5** (a template `fail` for `type: Recreate` + `rollingUpdate`): the API server's rejection is immediate and names the exact field and constraint, so a guard would duplicate a good error while adding surface that must track upstream validation.

**Validation:** `helm lint` 0 failed on all five charts; `check_helm_pss.py` 9/9; default render **byte-identical** to `dcbcdbd3e` for all four charts and the umbrella (baseline from a fresh throwaway worktree with its own `helm dependency update` — `helm/*/charts/*.tgz` is gitignored, so `git stash` does not revert built subchart archives and a stale one silently masks a real delta). Post-amend render hashes match pre-amend exactly, confirming pre-commit altered nothing. Umbrella overrides re-verified through `mcp-mesh-grafana.grafana.strategy` and `mcp-mesh-tempo.tempo.strategy`.

Issue #1131, opened 2026-06-03
TypeScript runtime: extract shared MeshRuntimeBase (ApiRuntime/MeshAgent event-loop dedup)


dhyansraj:
Split out from #1116 (which consolidated the rest of the TypeScript MCP-call/LLM duplication across PRs #1126–#1130). This is the one remaining sub-item, deferred because it carries materially higher structural risk than the others.

## Problem
`ApiRuntime` and `MeshAgent` duplicate the registry event loop, the dependency add/remove fallback, and the signal handlers — `api-runtime.ts:292-466`, `agent.ts:1504-1701`.

## Proposed fix
Extract a shared `MeshRuntimeBase` over a small `DependencyStore` interface; `MeshAgent` handles its 3 extra event types via hooks.

## ⚠️ Behavior divergence to resolve first (do NOT paper over)
The two runtimes are **not** currently behavior-identical: on `registry_disconnected`, `ApiRuntime` **clears** its resolved dependencies while `MeshAgent` **does not**. Any shared base must make a deliberate decision here rather than silently picking one — surface it explicitly:
- If clearing-on-disconnect is correct, adopting it for `MeshAgent` is a behavior change (flag for release notes).
- If retaining is correct, `ApiRuntime` changes instead.

This is why the extraction was deferred from the #1116 batch: it is a real semantic reconciliation, not a mechanical dedup.

## Scope
- Shared `MeshRuntimeBase` (event loop + dep add/remove fallback + signal handlers).
- `DependencyStore` interface abstracting the two runtimes' dep storage.
- Hook points for `MeshAgent`'s 3 extra events.
- Resolve the `registry_disconnected` divergence with a documented decision + test.

Source: multi-agent code audit (2026-06-01); adversarially verified.

dhyansraj:
## Divergence resolved (documented decision) — PR #1145

The `registry_disconnected` divergence is resolved in favor of **RETAIN** (the behavior MeshAgent / Python / Java already had); `ApiRuntime` + `MeshExpress` stop calling `clearAllDependencies()`.

**Decision: retain resolved deps through a registry outage.** Rationale:
- The registry is **control-plane**; resolved dependency endpoints are **data-plane** (direct agent→agent) and remain valid through a registry blip — the whole mcp-mesh resilience promise (registry can die; agents keep talking hardwired).
[...]

PR #1413, opened 2026-07-29
fix(helm): keep the core Namespace out of helm uninstall's blast radius

### LINKED ISSUE #1382: mcp-mesh-core owns the Namespace by default — helm uninstall deletes the namespace and every unrelated workload in it
## Summary

`mcp-mesh-core` creates and **owns** the Kubernetes Namespace it is installed into, by default. `helm uninstall` therefore **deletes the namespace and every unrelated resource inside it** — application workloads, Services, Secrets, and PersistentVolumeClaims that the chart never created.

This cost us a production namespace: uninstalling `mcp-core` in order to migrate to an externally-managed core deleted 10 agent Deployments, an application gateway, an nginx front end, a MinIO instance holding user-uploaded media, and all three PVCs (17Gi total). None of it was owned by the chart.

Chart version checked: **3.3.0 / 3.3.1** (`oci://ghcr.io/dhyansraj/mcp-mesh/mcp-mesh-core`).

## Root cause

`templates/namespace.yaml`:

```yaml
{{- if .Values.namespaceCreate }}
apiVersion: v1
kind: Namespace
metadata:
  name: {{ .Values.global.namespace | default .Release.Namespace }}
```

`values.yaml`:

```yaml
global:
  namespace: mcp-mesh     # documented as "used by namespaceCreate"
namespaceCreate: true     # default
```

Because `namespaceCreate` defaults to `true`, the release takes ownership of a `Namespace` object. Helm deletes owned resources on uninstall, and deleting a Namespace cascades to everything in it.

The chart's own documented usage makes this the *normal* path rather than an edge case:

```
helm install ... -n my-namespace --set global.namespace=my-namespace
```

Anyone following that instruction ends up with a core release that owns their application namespace.

## Impact

- **`helm uninstall <core-release>` is a namespace-wide destructive operation**, with nothing in the command or its output indicating that.
- It is silent: `helm uninstall` reports success, and the blast radius is only visible afterwards.
- It is most likely to fire during exactly the operation where it does most harm — **migrating from a self-hosted core to a managed/shared one**, which is when an operator deliberately uninstalls core while intending to keep everything else running.
- `helm uninstall --dry-run` does not make the cascade obvious either, since the Namespace deletion is one line among many.

## Reproduce

1. `helm install mcp-core oci://ghcr.io/dhyansraj/mcp-mesh/mcp-mesh-core -n my-app --create-namespace --set global.namespace=my-app`
2. Deploy anything else into `my-app` — an agent, a Deployment, a PVC.
3. `helm uninstall mcp-core -n my-app`
4. The namespace and everything in it is gone.

## Suggested fixes

**Default `namespaceCreate` to `false`.** Creating the namespace you are being installed into is unusual for a Helm chart; `helm install --create-namespace` and GitOps tooling (Argo CD's `CreateNamespace=true`) both already handle it, and neither takes ownership in a way that cascades on uninstall.

If the default must stay `true` for first-run convenience, then at minimum:

- **Add a Helm resource policy** so the namespace is never garbage-collected on uninstall:
  ```yaml
  metadata:
    annotations:
      "helm.sh/resource-policy": keep
  ```
  This is the standard mechanism for exactly this hazard and is a one-line change.
- **Warn in `NOTES.txt`** when `namespaceCreate` is true, stating that uninstalling the release will delete the namespace and everything in it.
- **Drop the `mcp-mesh` literal default for `global.namespace`.** `.Release.Namespace` is the safer default; the current literal means an install that does not override it creates a *second*, unrelated namespace. (Separately, this causes multiple core releases across different namespaces to contend for one shared `Namespace/mcp-mesh` object — two Argo CD Applications ping-ponged ownership of it in our cluster until we set `namespaceCreate: false`.)

I would consider `helm.sh/resource-policy: keep` the minimum bar here even if the default changes, since it protects existing installations that already have the namespace in their release manifest.

## Context

Reported alongside #1381. Both were found migrating a production mesh application onto managed infrastructure. Happy to open a PR for either.

dhyansraj:
## Summary

`helm uninstall <core-release>` deleted the Namespace it owned — cascading to every unrelated workload, Service, Secret and PVC inside it, with nothing in the command or its output indicating that. This has already cost a production namespace: 10 agent Deployments, an application gateway, nginx, MinIO and three PVCs, none of it chart-owned.

**The fix is one annotation.** `helm.sh/resource-policy: keep` on the Namespace makes Helm skip it on uninstall. `namespaceCreate` stays `true`; there is no migration and no user action.

## This PR originally also flipped the default — live testing showed that flip was destructive

The first version of this PR defaulted `namespaceCreate: false`, on the reasoning that a chart owning its own namespace is unusual. Tested against k3s on Helm 3.19.0 and 4.0.1, upgrading an existing release to that chart:

```
$ helm upgrade core -n nstest-a1 --set global.namespace=nstest-a1 <new-chart>
Release "core" has been upgraded. Happy Helming!
STATUS: deployed

$ kubectl get ns nstest-a1
NAME        STATUS        AGE
nstest-a1   Terminating   36s
```

The unrelated ConfigMap and PVC went with it. Helm reported success — exit 0, no warning. And it is **unrecoverable**: the release secret lives in the namespace being deleted, so afterwards `helm history` returns `release: not found`.

Mechanism: Helm computes `original.Difference(target)`, finds the Namespace absent from the new manifest, checks the **live** object for `resource-policy`, finds none (the old template had no annotations), and deletes it. The `keep` annotation cannot help, because at a `false` default the Namespace never renders and so is never applied.

On our own cluster, 5 of ~11 mesh namespaces were in that shape.

The decisive counter-test: installing this chart with `namespaceCreate: true` and then running `helm uninstall` leaves the namespace **and** an unrelated marker workload intact. **The annotation alone fully fixes the bug; the default flip added no protection and was solely what created the destructive path.** So it is gone.

## Changes

- **`helm.sh/resource-policy: keep`** on the core Namespace. The `annotations:` block is now unconditional, with `commonAnnotations` merged beneath.
- **Deleted `helm/mcp-mesh-registry/templates/namespace.yaml`** — gated on `createNamespace`, a key defined nowhere in that chart's values. Live but unreachable except via an undocumented key. A `createNamespace` case was added to `mcp-mesh-registry.validateNoRemovedKeys`, matching the convention that chart's `_helpers.tpl:237-241` states verbatim: *"a values file carrying one would silently no-op while the user expects effect. Fail loudly with migration guidance instead."*
- **`NOTES.txt` hints now print the release namespace.** They previously printed `mcp-mesh` regardless of where you installed, because `.Values.global.namespace | default .Release.Namespace` short-circuits — `default` never fires against a non-empty literal. Measured against where workloads actually land, 2 of 3 install shapes were wrong; the only one that agreed was `--set global.namespace=`, which is why it survived so long.
- **`NOTES.txt` ownership warning rewritten** to fire only when `global.namespace != .Release.Namespace`. That is the stray-namespace case the annotation does *not* fix — a second, unrelated namespace, which is the Argo CD ownership-contention pathology from the issue.
- **Docs** — `docs/upgrading.md` previously advised existing owners to "move to `namespaceCreate=false`", which is exactly the destructive operation, handed to the users its own diagnostic identifies as at risk. Now states that upgrading applies the annotation automatically and warns explicitly against the flip. README quickstart prose now says `namespaceCreate` creates `global.namespace` (default `mcp-mesh`), **not** the release namespace — so `-n my-app --set namespaceCreate=true` alone creates a stray namespace and then fails.

## Validation

CI runs no helm jobs on a PR — `helm lint` and `check_helm_pss.py` only fire on a release tag — so this is all local plus live cluster.

- **Default render diff vs `3dd7d2cee` is exactly the two annotation lines** (1133 → 1135). Baseline from a throwaway `git worktree` with its own `helm dependency update`; `helm/*/charts/*.tgz` is gitignored so `git stash` does not revert built subchart archives, and grafana/postgres passwords regenerate every render, so both were pinned.
- **Uninstall safety, live on k3s** — installed with `namespaceCreate: true`, added an unrelated ConfigMap, ran `helm uninstall`: namespace `Active`, ConfigMap intact.
- **Removed-key guards fire.** `namespace.yaml:1` holds the repo's only invocation of `mcp-mesh-core.validateNoRemovedKeys`; all three keys `rc=1`, control `rc=0`. New registry `createNamespace` guard `rc=1`, and the chart's existing guards still fire. A comment now marks that invocation as sole, so it is not lost to a future cleanup.
- **NOTES hints correct across four install shapes**, with the stray-namespace warning firing exactly where it should.
- `helm lint` 2 charts / 0 failed · `check_helm_pss.py` 9 charts OK · `go build ./...` clean (the man page is `go:embed`-ed).

## Found while testing, not addressed here

**`namespaceCreate: true` has no working install path on Helm 3 at all.** Verified with a real Helm 3.19.0 client:

| shape | result |
|---|---|
| `-n <new-ns>`, no `--create-namespace` | `create: failed to create: namespaces "..." not found` |
| `-n <ns> --create-namespace --set global.namespace=<ns>` (**the documented recipe**) | `namespaces "..." already exists` |
| `kubectl create ns` first, then install | `invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by"` |

The only Helm 3 shape that succeeds is `global.namespace` **not** matching `-n` — i.e. the stray-namespace misconfiguration this PR warns about. Helm 4.0.1 succeeds on the documented recipe, adopting the namespace.

Reproduced identically against unmodified `3dd7d2cee`, so it is pre-existing and not introduced here. The charts advertise Helm 3.0+/3.2.0+ as the prerequisite. Filed separately.

Closes #1382

## Summary by CodeRabbit

* **Documentation**
  * Clarified that the Helm deployment `-n` namespace must match the core chart’s `global.namespace`; mismatches can leave a stray empty Namespace and cause shared-namespace contention.
  * Expanded Helm upgrade/uninstall guidance to explain safe in-place upgrades, namespace retention behavior, and warnings about changing `namespaceCreate`.
  * Updated installation/status/port-forward commands to consistently target the Helm release namespace.
* **Bug Fixes**
  * Added/expanded validations to prevent unsupported namespace ownership/resource-policy settings that could lead to destructive namespace deletion on upgrade/uninstall.
  * Updated the registry chart to remove conditional Namespace creation when `createNamespace` is set.


dhyansraj:
Verified all four review findings against current code. **Two valid and fixed in `2584f7366`; two rejected.**

## Fixed

**`commonAnnotations` could silently disable the protection this PR adds.** The hardcoded `keep` was merged *before* `.Values.commonAnnotations`, so a user value won on last-wins:

```
$ helm template t helm/mcp-mesh-core --set 'commonAnnotations.helm\.sh/resource-policy=delete'
  annotations:
    "helm.sh/resource-policy": keep
    helm.sh/resource-policy: delete      # ← wins at apply time
```

Exit 0, no error — `helm template` does not reject the duplicate. So the annotation this entire PR exists to add was switchable-off from values, silently, restoring the namespace cascade. Good catch.

Fixed on both sides: `omit` strips the key before merging so a duplicate is never emitted, and a new `mcp-mesh-core.validateNamespaceResourcePolicy` guard `fail`s on any non-`keep` value, naming the key, the hazard, and the alternative. An explicit `keep` stays a silent no-op. The comparison runs through `toString` so a YAML-typed value (`no`, `false`, a number) hits the guard rather than blowing up the comparison.

The guard is deliberately unconditional — it fires even at `namespaceCreate=false`, where `commonAnnotations` is inert. That matches this chart's convention (`validateNoRemovedKeys` also fails on values that render nothing), and since the umbrella applies `commonAnnotations` to *only* the Namespace, that setting is dead config that becomes dangerous the moment anyone enables the flag.

**`afterwards` → `afterward`** at `README.md:416` and `docs/upgrading.md:155`. Both were introduced by this PR; the repo uses "afterward" in four other places, so this is consistency rather than preference. `docs/downloads/*` left alone (generated).

## Rejected

**Setting `namespaceCreate: false`** — this was implemented in the first revision of this PR, live-tested on Helm 3.19.0 and 4.0.1, and **proved destructive**. Upgrading an existing release deleted the namespace and every unrelated resource inside it, reported `STATUS: deployed` with exit 0, and left no rollback because the release secret dies with the namespace. Five namespaces on our own cluster are in that shape. The decisive counter-test: with `namespaceCreate: true`, `helm uninstall` leaves the namespace and an unrelated marker workload intact — so the annotation alone fully fixes the bug, and the flip adds no protection while creating the destructive path. Deliberately reverted; see the PR description.

**Rewriting `docs/upgrading.md:144-156` to say the default is `false`** — same stale premise. The default is `true` and the current text is correct. Swept `namespaceCreate` across `*.md`/`*.yaml`/`*.go`/`*.tpl`: every site states `true`.

Both rejected comments appear to be written against this PR's first revision, before live testing inverted the design.

## Validation

- Override rejected: `--set 'commonAnnotations.helm\.sh/resource-policy=delete'` → exit 1 with the guard message.
- Explicit `keep`: exactly one annotation on the Namespace, no duplicate. (A whole-render grep shows two hits — the second is the pre-existing postgres credentials Secret, which has always carried `keep`.)
- Unrelated annotations: `keep` + `owner` + `team`, correct indent, no duplicate. Mixed case with an explicit `keep` also renders it once.
- Default render **byte-identical** to `56c7502cc`; diff vs `3dd7d2cee` still exactly the two annotation lines. Fresh throwaway worktree with its own `helm dependency update`, passwords pinned.
- Removed-key guards all still fire (three core keys + control, plus registry `createNamespace`).
- `helm lint` 0 failed · `check_helm_pss.py` 9 charts OK · `go build ./...` clean.

dhyansraj:
Verified the three findings against current code in `ee5891e7f`: **two valid, one rejected** — plus a severity correction from the repo owner that mattered more than any of them.

## The framing was wrong, and that's now fixed

The docs inherited the *incident's* severity rather than the *mechanism's*. Corrected in four places (`README.md`, `docs/upgrading.md`, and `man upgrading.md`, which carried the same claim and was additionally stale — still asserting a cascade the annotation now prevents):

- **Registry state is derived, not precious.** Agents re-register on their next heartbeat, so a registry that comes back empty repopulates itself. `Back up data before uninstalling` was misleading advice about mesh data. Confirmed in code: an unknown/unhealthy agent gets 410 → `requires_full_heartbeat()` → full re-registration (`ent_handlers.go:598-603`, `ent_service.go:1434`).
- **The bundled PostgreSQL PVC isn't even deleted.** It comes from the StatefulSet's `volumeClaimTemplates`, which the release does not own, and there is no `persistentVolumeClaimRetentionPolicy` anywhere in `helm/` — so `helm uninstall` leaves the volume behind. Deleting the PVC (or the namespace around it) is what destroys data. "Uninstalling drops the registry's database" was wrong twice over.
- **The real hazard is co-located application data.** The #1382 incident happened because *applications* had used the bundled PostgreSQL/Redis for their own storage. That data isn't derived and no heartbeat returns it. The warning now says that instead.
- **Agents keep serving while the registry is down** — resolved dependencies are never cleared on disconnect (#1131; verified at `rust_heartbeat.py:675-676`, `agent.ts:2296-2298`, with a dedicated regression test at `registry-disconnect-retains-deps.spec.ts`). Deliberately *not* overclaimed: the docs now name what pauses — new capability discovery, re-resolution of changed ones, and lookups a client hasn't already resolved — aligned with the existing `man health.md:145` / `man proxies.md:163` statements.

## Fixed

**Upgrade vs. teardown split.** "Never `helm uninstall` the core release" read as a blanket prohibition, which is wrong now the Namespace survives. Now two bullets: don't use uninstall as an *upgrade mechanism* (it discards release history and values for no benefit), and uninstall remains a legitimate explicit teardown with the consequences stated accurately.

**Unrecoverability narrowed.** The `helm history` → `release: not found` claim only holds when `global.namespace` matches the `-n` target — Helm writes the release secret into the `-n` namespace while the template renders `global.namespace`. When they differ, release metadata survives. The namespace-deletion warning stays unconditional, since the rendered namespace is deleted either way.

## Rejected

**Removing `--create-namespace` from the install examples.** Tested on Helm 4.0.1 against a live cluster:

```
$ helm install cntest ./helm/mcp-mesh-core -n nstest-cn-82733 --set global.namespace=nstest-cn-82733 ...
Error: INSTALLATION FAILED: create: failed to create: namespaces "nstest-cn-82733" not found
```

Helm writes the release secret into the `-n` namespace *before* applying the manifest, so a chart-templated `Namespace` can never create the namespace its own release installs into — only re-declare one that exists. Reproduced on Helm 3.19.0 as well (#1414). Both examples left as-is.

The comment's premise — that `namespaceCreate` can manage the namespace instead — is exactly the impossibility behind #1414, and a reader could reasonably draw the same conclusion, so the surrounding prose now states that `--create-namespace` is **required** and quotes that failure.

## Validation

- Default render **byte-identical** to `2584f7366` (`cmp` clean, 1135 lines) — this is docs-only, so any delta would mean something unintended was touched. Fresh worktree with its own `helm dependency update`; note the correct grafana pin is `mcp-mesh-grafana.grafana.config.adminPassword`, which is the sole source of render nondeterminism.
- `git diff --name-only 2584f7366 HEAD` → exactly three doc files; no template or values changes.
- `go build ./...` clean, and `meshctl man upgrading` rendered to confirm the embedded page shows the new bullets.

dhyansraj:
Verified five findings against current code in `5b61e5067`: **two valid and fixed, three skipped.**

## Fixed

**`helm.sh/resource-policy` is reserved in `commonAnnotations`, and that was undocumented.** The guard added earlier in this PR is template-time, so it blocks `helm upgrade` exactly as it blocks install:

```
$ helm template t helm/mcp-mesh-core --set 'commonAnnotations.helm\.sh/resource-policy=delete'   → exit 1
```

A release currently setting that key would upgrade into a hard failure with no forewarning. Now stated as a prerequisite in both `docs/upgrading.md` and `man upgrading.md`, with the reason: the chart's `keep` on the Namespace is what prevents the cascade, and a `commonAnnotations` value overrides it on last-wins. Good catch — a protection that can block your upgrade needs to say so.

**"Helm leaves it (and everything in it) alone on uninstall" was overstated.** `helm uninstall` still removes the release's own resources — registry, PostgreSQL, Redis, and any observability components. What the annotation preserves is the *Namespace*, and therefore everything the chart did **not** create.

Narrower than reported: exactly one site (`docs/upgrading.md:163`). `man upgrading.md:74` already worded it correctly, so the two surfaces disagreed. The corrected clause is lifted verbatim from the man page so they now match.

## Skipped

**Placeholder for `-n mcp-mesh` in the README uninstall example.** It matches the install examples directly above it, and `helm uninstall` against the wrong namespace is non-destructive — it returns `release: not found`. A placeholder would break copy-paste for the default install those examples describe.

**A `not .Values.namespaceCreate` branch in `NOTES.txt`.** The existing warning is about a *stray namespace*, which can only exist when `namespaceCreate` renders one. At `false` there is no Namespace object, no ownership, and nothing actionable to warn about. NOTES also prints *after* the upgrade completes, so it could not forewarn about the flip even in principle.

**Removing `--create-namespace` from the install examples.** Tested on Helm 4.0.1 and 3.19.0 against a live cluster:

```
Error: INSTALLATION FAILED: create: failed to create: namespaces "nstest-cn-82733" not found
```

Helm writes the release secret into the `-n` namespace *before* applying the rendered manifest, so a chart-templated `Namespace` can never create the namespace its own release installs into — it can only re-declare one that already exists. `namespaceCreate` is not a substitute for `--create-namespace`; that impossibility is tracked as **#1414**. The prose in that section now states the requirement explicitly and quotes the failure, since the comment's premise is one a reader could reasonably reach.

## Validation

- Default render **byte-identical** to `ee5891e7f` — both 1135 lines, identical sha1. Docs-only, so any delta would mean something unintended was touched. Fresh worktrees with their own `helm dependency update`, passwords pinned.
- `git diff --name-only ee5891e7f HEAD` → exactly the two doc files.
- `go build ./...` clean; `meshctl man upgrading` renders the new bullet. (Note for anyone verifying: `meshctl` on PATH may be the npm-installed `@mcpmesh/cli` shim, which serves a stale embedded page — build from `./cmd/meshctl`.)
- Both surfaces now use identical wording for what uninstall preserves versus removes.

PR #1415, opened 2026-07-30
fix(helm): make the Grafana admin password survivable and GitOps-discoverable

### LINKED ISSUE #1391: mcp-mesh-grafana: lookup-based password generation causes permanent Argo CD drift — GitOps consumers get a new random password on every render
## Summary

`mcp-mesh-core`'s Grafana subchart auto-generates its admin-password Secret using `lookup`, which returns nothing under `helm template`. Every render therefore produces a **new random password**, so any GitOps tool that renders manifests (Argo CD, Flux, `helm template | kubectl apply`) sees permanent, unresolvable drift on `<release>-mcp-mesh-grafana-secret`.

The chart is published as an OCI chart and is consumed via GitOps — so the default path does not work in the mode the chart is primarily used in.

Chart version checked: **3.2.3 / 3.3.0**.

## The code

`charts/mcp-mesh-grafana/templates/secret.yaml`:

```gotemplate
{{- if and .Values.grafana.enabled (not .Values.grafana.config.existingSecret) }}
{{- with .Values.grafana.config.adminPassword }}
{{- $passwordB64 = . | b64enc }}
{{- else }}
{{- $existing := lookup "v1" "Secret" .Release.Namespace $name }}
{{- $passwordB64 = dig "data" "admin-password" "" ($existing | default dict) }}
{{- if not $passwordB64 }}
{{- $passwordB64 = randAlphaNum 24 | b64enc }}
{{- end }}
{{- end }}
```

The chart's own doc comment already identifies the problem:

> pure `helm template | kubectl apply` pipelines therefore regenerate the value on every render; helm install/upgrade is the supported path

That is an accurate description of Argo CD. `lookup` requires a live cluster connection that the render step does not have, so the `$existing` branch is dead in GitOps and every render falls through to `randAlphaNum 24`.

## Observed impact

Three Argo CD Applications across three namespaces, all consuming this chart with `grafana.config` unset (`{}`):

- One reports `OutOfSync` on `Secret/…-grafana-secret` and cannot converge. Syncing it silently rotates the Grafana admin password.
- Two report `Synced` — **not because they are correct, but because Argo CD caches rendered manifests** for a fixed chart version with unchanged values. They render once and serve the cached output. Any cache invalidation (re-registering the ApplicationSet, bumping `coreVersion`) surfaces the identical drift.

That second case is the more concerning one: the defect is latent and presents as healthy until an unrelated action exposes it.

## Why the workaround is hard to find

`grafana.config.existingSecret` fixes it completely — when set, `secret.yaml` renders nothing and the Deployment reads `GF_SECURITY_ADMIN_PASSWORD` via `secretKeyRef`. But this is discoverable only by reading `values.yaml` comments or the template source. Nothing surfaces it at install time, and the failure mode is silent drift rather than an error.

## Suggested fixes

**Preferred — do not auto-generate by default.** When neither `grafana.config.adminPassword` nor `grafana.config.existingSecret` is set, `fail` with a message naming both options. A clear error at render time is strictly better than silent per-render rotation, and it makes the GitOps requirement impossible to miss.

**Alternative — keep generation but make it GitOps-safe:**
- Add `helm.sh/resource-policy: keep` and document that GitOps consumers must set `existingSecret`.
- Prominently note the `helm template` limitation in `NOTES.txt` and the README, not only in a template comment.

**Precedent within this same chart:** the Postgres path already does this correctly —

```yaml
global:
  postgres:
    generatedSecret: false
    existingSecret: mcp-core-postgres-auth
    existingSecretPasswordKey: password
```

An operator can supply an externally-managed Secret and the chart generates nothing. Applying that same shape to Grafana — ideally as the default — would resolve this.

## Note on remediation

Adopting `existingSecret` on an existing install needs care: Grafana applies `GF_SECURITY_ADMIN_PASSWORD` only on **first start**, and with persistence enabled the effective password remains the one generated at install time. Operators must seal the password Grafana already has, or reset the PVC. Worth stating in whatever documentation accompanies the fix.

## Context

Third report from the same migration, alongside #1381 (no `strategy` passthrough) and #1382 (`namespaceCreate` owning the namespace). Happy to open a PR.

dhyansraj:
## Summary

`mcp-mesh-grafana` generates its admin password with `lookup`, which returns nothing under `helm template`. Every render therefore produces a **new random password**, so any tool that renders manifests — Argo CD, Flux, `helm template | kubectl apply` — sees permanent, unresolvable drift on `<release>-mcp-mesh-grafana-secret`. The chart is published as OCI and consumed via GitOps, so the default path did not work in the mode the chart is primarily used in.

The latent form is worse than the visible one: an Argo CD Application that renders once and caches reports `Synced` and looks healthy, until any cache invalidation surfaces the identical drift.

This brings Grafana to the credential shape `mcp-mesh-postgres` already uses, and keeps generation as the default.

## Changes

- **`grafana.config.generatedSecret`** (default `true`) with `mcp-mesh-grafana.validateCredentialSource`, mirroring `mcp-mesh-postgres.validateCredentialSource`. Fails when generation is off and neither `adminPassword` nor `existingSecret` is supplied. Invoked inside the `grafana.enabled` gate so it stays quiet when the chart does not render.
- **`helm.sh/resource-policy: keep`** and a `namespace:` field on the generated Secret, matching the Postgres Secret.
- **`mcp-mesh-grafana.secretName` helper**, now used by both `secret.yaml` and the Deployment's `secretKeyRef`. The name was previously computed in three places; this reduces it to two (the umbrella keeps its own inlined helper, with a comment pointing at the new source of truth).
- **Discoverability** — `existingSecret` already worked end to end but was findable only by reading template source. Now surfaced on the umbrella `values.yaml`, in `NOTES.txt`, in a new `### Grafana admin password` section of `helm/mcp-mesh-core/README.md`, and in `helm/README.md`'s "Generated Credentials".

## Why the default path does not `fail`

The issue's preferred fix was to `fail` when neither `adminPassword` nor `existingSecret` is set. That would break the release: `scripts/check_helm_pss.py` runs `helm template` on every chart with **default** values, as does `helm lint`, both in `helm-release.yml`. It would also penalise `helm install`/`upgrade` users, for whom `lookup` works correctly.

So generation stays the default and the GitOps requirement is made discoverable instead. `check_helm_pss.py` passing on defaults is part of the validation below precisely because it is the check that a fail-on-default would have broken.

## The lockout this actually prevents

The `keep` annotation matters more than it appears, but not for the reason first assumed. It was initially justified as "uninstall drops the Secret while the PVC survives" — **that is wrong for this chart.** `helm/mcp-mesh-grafana/templates/pvc.yaml` is a plain release-owned PVC with no `keep`, so `helm uninstall` deletes the volume too (observed `Terminating` immediately after). A fresh Secret and a fresh `grafana.db` match, so that path is not a lockout.

The real path was verified on the cluster against the baseline chart:

1. Install → password generated (`RlpJzxIeQ3WNFCMmI6rGK9kC`).
2. `helm upgrade` adding `existingSecret` — the documented GitOps migration.
3. The generated Secret **is deleted in that same upgrade**, while persistence survives an upgrade.

So the password Grafana is actually running with becomes unrecoverable at the exact moment an operator follows the migration advice. Grafana honours `GF_SECURITY_ADMIN_PASSWORD` only on first start, so the new `existingSecret` value has no effect on the provisioned instance — the result is an instance nobody can log into.

With `keep`, the Secret survives that upgrade (verified), which is what makes the documented "seal the password Grafana already has" remediation performable at all. All four doc surfaces describe this path, not the uninstall one.

## Validation

CI runs no helm jobs on a PR, so this is local plus live cluster.

- `helm lint` grafana + core: 0 failed.
- **`python3 scripts/check_helm_pss.py`: all 9 charts OK on default values.**
- **Default render vs `ee653b6ab`**: only `+ namespace:`, `+ "helm.sh/resource-policy": keep`, and the unrelated postgres password (which regenerates every render). Baseline from a throwaway `git worktree` with its own `helm dependency update` — `helm/*/charts/*.tgz` is gitignored, so `git stash` does not revert built subchart archives and a stale one silently masks a delta.
- **`existingSecret` through the umbrella's values** (not just the subchart's): no chart-managed Secret emitted, `secretKeyRef` → the supplied name/key, generated-password NOTES block suppressed. Two consecutive renders byte-identical — against the default path, which produced two different passwords.
- `generatedSecret: false` alone → fails with the guard message, via both umbrella and subchart. Plus `adminPassword` → renders. Plus `existingSecret` → renders, no generated Secret. `grafana.enabled=false` with `generatedSecret=false` → no fail.
- **Live k3s**: install → `helm upgrade` reused the same password → `helm uninstall` **left the Secret in place** → reinstall under the same release name adopted the kept Secret and `lookup` recovered the same password, with no ownership error.

## Follow-up, not addressed here

`helm/mcp-mesh-grafana/templates/pvc.yaml` carries no `helm.sh/resource-policy: keep`, so `helm uninstall` silently destroys Grafana's dashboards, annotations and users — unlike the bundled PostgreSQL, whose data survives on a `volumeClaimTemplates` PVC the release does not own. Whether that asymmetry is intended is a separate question; filed separately.

Closes #1391


dhyansraj:
Review applied in `c9141755c` — 0 BLOCKER, 1 WARNING, 6 INFO. Five fixed, one skipped.

## WARNING — `keep` cannot protect the upgrade that installs it

The important one, and structurally the same trap as #1382. Helm reads `helm.sh/resource-policy` off the live object as it exists *going into* the upgrade, and the previous release's stored manifest is equally annotation-free. So on a pre-3.3.3 release, a single commit that bumps the chart **and** sets `existingSecret` still deletes the generated Secret — while persistence keeps Grafana serving on the password in `grafana.db`, which is now unrecorded. That is precisely the lockout the annotation was added to prevent.

The docs made it worse by asserting the Secret was safe *because of* the annotation, and offering a remediation branch that never read the old value — together an invitation to skip the one step that makes this safe.

Corrected on **four** surfaces (the review named three; the umbrella `values.yaml` carried the same unqualified claim):

- Retention claims now scoped to "once this chart version has been applied".
- A dedicated bullet titled **"The annotation does not protect the upgrade that installs it."** The old "releases installed with an older chart pick the annotation up on `helm upgrade`" line is gone.
- Adoption is now a numbered two-step procedure: upgrade the chart **alone** and verify with `kubectl get secret … -o jsonpath='{.metadata.annotations}'`, *then* adopt.
- "Retrieve and seal the current value" is **unconditional** — changing to a different password is a follow-on action, no longer an alternative to reading the old one.

## `--set-string generatedSecret=false` silently bypassed the guard

The string `"false"` is truthy in Go templates, so `not $cfg.generatedSecret` never fired and a user disabling generation that way got no Secret and no error. Third guard-that-cannot-fire found today, after the `(vX+)` pattern that could never match and the missing version boundary in `bump_version.py` — same fix shape, an explicit normalized comparison.

`hasKey` dropped (dead once the key ships in values). Verified matrix:

| form | result |
|---|---|
| `--set-string …=false` | **fails** |
| `--set …=false` (bool) | fails |
| `--set-string …=False` | fails |
| `--set-string …=0` | fails |
| `--set-string …=true` / `--set …=true` / default | render |

## Also fixed

- **Fourth secret-name site** (`grafana/NOTES.txt:8`) migrated to `mcp-mesh-grafana.secretName`. It was the exact site the helper's docstring claimed did not exist — i.e. how the drift the helper prevents gets reintroduced.
- **Guard docstring corrected**: it claimed "the chart then creates no Secret of its own", untrue when `adminPassword` is set, since `secret.yaml` never consults `generatedSecret`. Now states the flag is enforced *only* by the guard.
- **`mcp-mesh-core.grafanaSecretName` comment corrected**: it claimed to inline the subchart derivation "with the standard fullname rules", but only reproduces the release-name branch — under `nameOverride` the NOTES retrieval command returns `NotFound`. The comment now names the drift and states plainly that nothing catches it, unlike postgres which treats this class as fail-worthy.
- **PVC remediation order** completed: `scale --replicas=0` → `delete pvc` → `helm upgrade`, with the reason (the `pvc-protection` finalizer blocks while mounted, and nothing recreates the claim).

## Skipped

The `namespace:` field on the Secret. Deliberate parity with `mcp-mesh-postgres/templates/secret.yaml:33`, which already imposes the same `helm template --namespace` requirement. Consistency wins over relitigating it here.

## Validation

- Guard matrix above, via both the umbrella and the subchart standalone. `grafana.enabled=false` with `generatedSecret=false` stays quiet.
- Default render vs `ee653b6ab`: only `+ namespace:`, `+ resource-policy: keep`, and the unrelated postgres password. **Byte-identical** to the pre-amend render.
- `helm lint` 0 failed · `check_helm_pss.py` 9 charts OK.
- All four secret-name sites agree under `nameOverride=grafana`.

**Methodology correction worth recording:** `helm template` does not render `NOTES.txt` at all in Helm 4, so an earlier "NOTES block suppressed under `existingSecret`" check in this PR was vacuous. Re-verified with `helm install --dry-run=client`: the generated-password block is correctly absent, and the new two-step warning renders.

Issue #1416, opened 2026-07-30
helm: mcp-mesh-grafana's PVC is deleted on uninstall, silently destroying dashboards — while the bundled postgres volume survives


dhyansraj:
## Summary

`helm/mcp-mesh-grafana/templates/pvc.yaml` is a plain release-owned `PersistentVolumeClaim` with no `helm.sh/resource-policy` annotation. `helm uninstall` therefore deletes it, destroying every Grafana dashboard, annotation, user, and API key stored in `grafana.db`.

The bundled PostgreSQL behaves the opposite way: its data sits on a `volumeClaimTemplates` PVC that the release does not own, so `helm uninstall` leaves the volume behind. Two bundled stateful components, two opposite retention behaviours, neither documented as a choice.

Observed directly: the PVC goes to `Terminating` immediately after `helm uninstall`.

## Why the asymmetry is the wrong way round

Of the two, PostgreSQL is the one holding **derived** state — the registry's contents rebuild themselves from agent heartbeats, so losing that volume costs a transient topology gap. Grafana's volume holds **authored** state: dashboards someone built, annotations someone wrote, users someone created. None of it is derived and no heartbeat brings it back.

So the component whose data is disposable retains its volume, and the component whose data is irreplaceable discards it.

## Context

Surfaced while fixing #1391 (PR #1415). That PR adds `helm.sh/resource-policy: keep` to the Grafana admin-password **Secret**, and the PVC asymmetry was found while verifying which lockout path the annotation actually prevents. The Secret and the PVC are now inconsistent with each other within the same chart: the credential survives uninstall, the data it unlocks does not.

Note the initial (incorrect) justification for the Secret annotation was "uninstall drops the Secret while the PVC survives" — that framing assumed PVC retention that does not exist here. Correcting it is what exposed this.

## Options

**(a) Add `helm.sh/resource-policy: keep` to the PVC.** Matches the Secret in the same chart and the Postgres volume's effective behaviour. Consistent with `grafana.persistence.enabled` defaulting to `true` — enabling persistence implies the data is worth keeping. Cost: an uninstall no longer fully cleans up, so a later reinstall adopts an existing volume, and operators who expect uninstall to reclaim storage would need to delete the PVC explicitly. `values.yaml` and `NOTES.txt` would need to say so.

[...]

Issue #1412, opened 2026-07-29
helm: mcp-mesh-redis advertises a persistence block but ships no PVC template — enabling it yields a Pending pod on a nonexistent claim


dhyansraj:
## Summary

`mcp-mesh-redis` exposes a full `persistence` values block and mounts a `persistentVolumeClaim` when it is enabled — but the chart contains **no PVC template**. Setting `persistence.enabled: true` therefore produces a pod stuck `Pending` on a claim that nothing ever creates.

## The gap

`helm/mcp-mesh-redis/templates/` contains exactly three files:

```
_helpers.tpl
deployment.yaml
service.yaml
```

No `pvc.yaml`, and no `kind: PersistentVolumeClaim` anywhere in the chart.

`helm/mcp-mesh-redis/values.yaml:33-39` nonetheless advertises the whole tuning surface:

```yaml
# Persistence configuration (using emptyDir for cache)
persistence:
  enabled: false
  storageClass: ""
  accessMode: ReadWriteOnce
  size: 8Gi
  annotations: {}
```

and `helm/mcp-mesh-redis/templates/deployment.yaml:89-96` branches on it:

```yaml
      volumes:
        - name: redis-data
          {{- if .Values.persistence.enabled }}
          persistentVolumeClaim:
            claimName: {{ include "mcp-mesh-redis.fullname" . }}
          {{- else }}
          emptyDir: {}
          {{- end }}
```

So `storageClass`, `accessMode`, `size` and `annotations` are read by nothing at all, and the `enabled: true` branch references a claim name no template produces.

The parenthetical *"(using emptyDir for cache)"* suggests emptyDir was the deliberate design and the rest is vestigial — but the values keys and the `claimName` branch both present persistence as a supported, configurable feature.

## Why it matters now

The three sibling charts with the same values shape (`mcp-mesh-agent`, `mcp-mesh-grafana`, `mcp-mesh-tempo`) all have real `pvc.yaml` templates, so redis reads as consistent with them at a glance. Anyone reasoning by analogy — or reading the values file alone — will conclude persistence works here.

Surfaced while reviewing #1410, which adds a `strategy` passthrough with guidance about RWO volumes. That guidance was written for all four charts uniformly and had to be trimmed for redis specifically, because an operator following it would set `persistence.enabled: true`, get a `Pending` pod, and never reach the deadlock the advice protects against.

## Reproduce

1. `helm install redis helm/mcp-mesh-redis --set persistence.enabled=true`
[...]

PR #1418, opened 2026-07-30
fix(helm): order the grafana and tempo pod swap so a single-replica RWO rollout can finish

### LINKED ISSUE #1411: helm: grafana and tempo default to a Deployment + RWO PVC + replicas:1 — a combination that cannot roll on multi-node clusters
## Summary

`mcp-mesh-grafana` and `mcp-mesh-tempo` ship a `Deployment` with `replicas: 1` hardcoded and a `ReadWriteOnce` PVC **enabled by default**. That combination cannot complete a rolling update on a multi-node cluster: the incoming pod cannot mount a claim the outgoing pod still holds, the rollout never finishes, and the old pod is never terminated.

Both charts are in the default `mcp-mesh-core` install path, so a standard install already carries two workloads that deadlock on any pod-template change.

#1410 added a `strategy` passthrough so operators can select `Recreate`. That is an escape hatch, not a fix — the shipped default is still broken, and it requires the operator to know about the problem in advance.

## Why `Recreate` is not the answer

Defaulting `strategy: {type: Recreate}` was implemented and then reverted in #1410 after testing against a real 3-node k3s cluster. Helm's 3-way merge cannot clear the API-server-defaulted `rollingUpdate` block:

```
Error: UPGRADE FAILED: Deployment.apps "probe" is invalid:
spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is 'Recreate'
```

This fails under Helm 4's default server-side apply and under `--force-conflicts`; only `--server-side=false` succeeds. So flipping the default would convert a conditional deadlock into an **unconditional upgrade failure** for every Helm 4 consumer, including those with persistence disabled who have no exposure at all.

## Proposed: convert both to StatefulSets

`mcp-mesh-postgres` already models this correctly — a StatefulSet with `volumeClaimTemplates`. Ordered replacement makes RWO safe by construction, with no strategy field to transition and therefore none of the trap above.

Scope:

- Resource kind change `Deployment` → `StatefulSet` in both charts.
- PVC ownership migration: today the PVC is chart-managed (`{{ fullname }}-pvc`); under a StatefulSet it becomes a `volumeClaimTemplate`, which names claims `<volume>-<statefulset>-<ordinal>`. **Existing data lives in a differently-named claim**, so this needs a documented migration (retain the PV, pre-create the renamed claim, or a documented backup/restore) — a silent switch would orphan existing dashboards and traces.
- `serviceName` and a headless Service.
- Both charts hardcode `replicas: 1` rather than exposing `replicaCount`; worth resolving in the same change.

## Alternative worth weighing

Expose `persistence.enabled: false` more prominently and let the default be ephemeral. Grafana's dashboards are provisioned from ConfigMaps in this chart, and tempo is a trace buffer — neither is necessarily precious. That would sidestep RWO entirely for the default install, at the cost of losing manually-created dashboards and trace history on restart.

Deciding between these two is the substance of this issue; the StatefulSet route is the one that preserves current behaviour.

## Acceptance

- A default `mcp-mesh-core` install can complete a pod-template change on a multi-node cluster without operator intervention.
- Existing installations have a documented, non-destructive upgrade path for their existing PVCs.
- No upgrade path depends on `--server-side=false`.

dhyansraj:
## Summary

`mcp-mesh-grafana` and `mcp-mesh-tempo` ship a `Deployment` with `replicas: 1` hardcoded and a `ReadWriteOnce` PVC **enabled by default**. With the API-default `maxSurge: 25%`, `RollingUpdate` starts the incoming pod before the outgoing one is gone, so on a multi-node cluster it cannot mount the claim and the rollout never finishes. Both charts are in the default `mcp-mesh-core` path, so a standard install carries two workloads that deadlock on any pod-template change.

The fix is `maxSurge: 0` / `maxUnavailable: 1`, **derived in the template**, leaving `type: RollingUpdate`.

## Why not the StatefulSet conversion this issue proposes

Measured: converting a Deployment + plain PVC to a StatefulSet with `volumeClaimTemplates` **silently destroys the data**.

```
STATUS: deployed / REVISION: 2
storage-v-mig-0   Bound   pvc-98eff722-…     <- brand-new empty volume
                                             <- v-mig-pvc: GONE
cat: can't open '/data/marker.txt': No such file or directory
```

The old PVC is dropped from the manifest with no `keep` on the live object, so Helm deletes it — reporting success. That would have shipped #1416's failure mode across 9 live installs.

Two further traps on that route, for the record: a StatefulSet with the default `podManagementPolicy: OrderedReady` **will not roll a crash-looping pod at all** (measured stuck ~2 min with `updateRevision` never applied; needs a manual pod delete), and `podManagementPolicy` is immutable, so getting it wrong at creation is unfixable in place. `volumeClaimTemplates` is immutable too — a one-way door in both directions.

The plain-PVC StatefulSet variant *does* work, and works structurally (the replacement pod has the same name, so two cannot coexist). It is simply more change than needed. Worth keeping in mind only if the transient attach warning below ever matters in the field — and then: plain PVC plus `podManagementPolicy: Parallel`, never `volumeClaimTemplates`.

## Why the default is derived in the template, not in `values.yaml`

`strategy` is a map and Helm deep-merges it. A populated default in `values.yaml` plus an operator who followed #1410's guidance and set `strategy.type: Recreate` yields *both* keys:

```
The Deployment "x" is invalid: spec.strategy.rollingUpdate: Forbidden:
may not be specified when strategy `type` is 'Recreate'
```

That is exactly the unconditional-failure class #1410 was reverted to avoid, relocated into a values merge. So `values.yaml` keeps `strategy: {}` and the template resolves the default, gating `rollingUpdate` on the resolved type.

## This fixes #1410's `Forbidden` trap rather than sidestepping it

Because the chart now **owns** `.spec.strategy.rollingUpdate` instead of leaving it to API defaulting, switching a live release to `Recreate` now succeeds under Helm 4 SSA with no flags — verified from a release whose previous revision had the block set. `--server-side=false` no longer appears anywhere in the repo.

## `maxUnavailable: 1` is load-bearing

It API-defaults to 25%, which rounds down to **0** at `replicas: 1`, and the API server rejects that:

```
spec.strategy.rollingUpdate.maxUnavailable: Invalid value: 0:
may not be 0 when 'maxSurge' is 0
```

Verified by driving it deliberately (the release went `failed`, then recovered on the next upgrade). Documented as a constraint on overrides in all five values files.

## Deadlock reproduced, then fixed

k3s, 5 nodes, `storageClassName: longhorn` pinned — the cluster has two default StorageClasses and `local-path` node-pins, which would not reproduce the bug. Cross-node move forced with a required `podAntiAffinity`.

Baseline live object confirms the trigger: `{"rollingUpdate":{"maxSurge":"25%","maxUnavailable":"25%"},"type":"RollingUpdate"}`.

**grafana**, wedged:
```
Waiting for deployment rollout to finish: 1 old replicas are pending termination...
error: timed out waiting for the condition          (180s)
…-56b9969586-qwc7d   0/1  ContainerCreating  3m55s  beelink4
…-645d7774f5-d8kr2   1/1  Running            4m32s  beelink5
Warning FailedAttachVolume  Multi-Attach error for volume "pvc-42f70115-…"
```
then `helm upgrade` to this branch, no flags → `deployed`, `{"rollingUpdate":{"maxSurge":0,"maxUnavailable":1},…}`, rolled out in **32s**, single pod.

**tempo**: same shape, `rollout status` timed out at 180s, upgrade converged in **47s**.

**Rescuing a `failed` release**: baseline `--wait --timeout 90s` wedged it to `failed rev 2`; upgrade to this branch reached `deployed rev 3` in 25s, no manual step, no `--server-side=false`.

## Operator-value matrix — rendered and applied live, Helm 3.19.0 and 4.0.1

| values | rendered | live |
|---|---|---|
| `{}` | `RollingUpdate` + `0`/`1` | deployed |
| `type: Recreate` | `Recreate`, no `rollingUpdate` | deployed, live `{"type":"Recreate"}` |
| `rollingUpdate.maxSurge: 1` | `maxSurge: 1` | deployed |
| dropping a prior `Recreate` | `RollingUpdate` + `0`/`1` | deployed |

Plus: an operator already on `strategy.type: Recreate` per #1410 upgrades cleanly keeping the override, and cleanly after dropping it. Umbrella paths render correctly through core. `strategy:` explicitly null falls back to the default. Full core umbrella server dry-run: all 20 objects accepted.

## Scope: grafana and tempo only

`mcp-mesh-agent` and `mcp-mesh-redis` also gained the `strategy` passthrough in #1410, but their persistence is opt-in, so the default path has nothing to fix. On the agent, `maxSurge: 0` would be a regression: `replicaCount` is user-settable and autoscaling is supported, so it would force a capacity dip on every rollout of a stateless multi-replica agent for no benefit. Redis mounts an `emptyDir` by default — a pod swap contends over nothing.

A conditional derivation (`persistence.enabled && replicaCount == 1 && !autoscaling.enabled`) was considered and rejected: it makes `.spec.strategy` a function of unrelated values, so flipping `persistence` would silently rewrite rollout semantics.

Their **docs** were rewritten anyway, so they no longer point at `Recreate`.

## Honest caveat, documented in every rewritten block

`maxSurge: 0` does not eliminate the transient overlap — the scale-down and scale-up are issued in the same second, so one `FailedAttachVolume` / `Multi-Attach` event is still logged per pod-template change. It converges unaided; the window is bounded by `terminationGracePeriodSeconds` plus CSI detach. Measured end to end on a forced cross-node move: grafana **27s**, tempo **50s**. The values comments say "under a minute in practice" rather than a single number.

## Validation

Default render vs `793cd16fa` is exactly two hunks (the new `strategy` block in each chart); `mcp-mesh-agent` and `mcp-mesh-redis` standalone renders byte-identical; Helm 3 and Helm 4 renders identical to each other. Baseline from two fresh detached worktrees each with its own `helm dependency update` (`helm/*/charts/*.tgz` is gitignored), passwords pinned. `helm lint` clean on grafana, tempo, agent, redis and core on both binaries; `check_helm_pss.py` all 9 charts OK.

## Pre-existing snag found while testing, not addressed

Installing `helm/mcp-mesh-grafana` **from the source tree** wedges on `MountVolume.SetUp failed for volume "dashboards": configmap "…-dashboard" not found` — `templates/configmap-dashboard.yaml` reads `files/dashboards/*.json`, which is populated only at release time, so the ConfigMap renders empty while the Deployment mounts it unconditionally. Released chart tarballs are unaffected; anyone testing from a clone hits it. Filed separately.

Closes #1411

## Summary by CodeRabbit

* **Documentation**
  * Updated deployment rollout guidance for persistent single-writer (RWO) volumes, including when to keep the Kubernetes default strategy.
  * Clarified how to avoid rollout deadlocks by pinning `rollingUpdate.maxSurge: 0` (while keeping `maxUnavailable` non-zero) and described expected temporary `FailedAttachVolume`/Multi-Attach events.
  * Rewrote Grafana and Tempo persistence strategy explanations and tightened security context documentation.

* **Improvements**
  * Grafana and Tempo now apply safer explicit rollout defaults when using `RollingUpdate`.
  * Tempo now validates and rejects security context settings that should be configured at the correct pod level.

### DIFF
--- a/helm/mcp-mesh-agent/README.md
+++ b/helm/mcp-mesh-agent/README.md
@@ -102,7 +102,7 @@ helm uninstall my-agent -n mcp-mesh
 | Parameter                   | Description                            | Default                    |
 | --------------------------- | -------------------------------------- | -------------------------- |
 | `replicaCount`              | Number of replicas                     | `1`                        |
-| `strategy`                  | Deployment update strategy (unset = Kubernetes default `RollingUpdate`; set `{type: Recreate}` for ReadWriteOnce `persistence`) | `{}` |
+| `strategy`                  | Deployment update strategy (unset = Kubernetes default `RollingUpdate`; pin `rollingUpdate.maxSurge: 0` for ReadWriteOnce `persistence`) | `{}` |
 | `image.repository`          | Container image repository             | `"mcpmesh/python-runtime"` |
 | `image.pullPolicy`          | Image pull policy                      | `IfNotPresent`             |
 | `image.tag`                 | Image tag (overrides chart appVersion) | `"0.9"`                    |
@@ -111,13 +111,21 @@ helm uninstall my-agent -n mcp-mesh
 | `resources.requests.cpu`    | CPU request                            | `100m`                     |
 | `resources.requests.memory` | Memory request                         | `256Mi`                    |
 
-> **`strategy` and ReadWriteOnce volumes:** with the default `RollingUpdate`, Kubernetes starts the incoming pod before the outgoing one is gone. If the scheduler places it on another node it cannot mount a ReadWriteOnce claim the outgoing pod still holds, and the rollout deadlocks until someone deletes the old pod. Both pods landing on one node is why single-node dev clusters (kind, minikube, Docker Desktop) never see this. Set `strategy: {type: Recreate}` whenever `persistence.enabled` is true with the default `accessMode: ReadWriteOnce`.
+> **`strategy` and ReadWriteOnce volumes:** the unset default gets Kubernetes' `RollingUpdate` with `maxSurge: 25%`, which starts the incoming pod before the outgoing one is gone — correct for a stateless agent, wrong the moment one mounts a ReadWriteOnce claim. If the scheduler places the incoming pod on another node it cannot mount the claim the outgoing pod still holds, and the rollout hangs indefinitely on a `Multi-Attach` error. Both pods landing on one node is why single-node dev clusters (kind, minikube, Docker Desktop) never see this. Whenever `persistence.enabled` is true with the default `accessMode: ReadWriteOnce`, pin the surge to zero:
 >
-> `Recreate` costs a gap in service while the old pod terminates, and when `autoscaling.enabled` is true that gap covers every replica at once, since `Recreate` tears the whole set down before scheduling any replacement. Note that `persistence` creates a **single** claim for the whole release, so ReadWriteOnce is not compatible with more than one replica under either strategy: with `replicaCount > 1` or an active HPA, every replica scheduled off the claim's node stays `Pending`. `Recreate` does not fix that combination — use ReadWriteMany, or keep the agent at one replica.
+> ```yaml
+> strategy:
+>   type: RollingUpdate
+>   rollingUpdate:
+>     maxSurge: 0
+>     maxUnavailable: 1
+> ```
 >
-> Set it at install time if you can. Switching an **already-installed** release from `RollingUpdate` to `Recreate` is rejected by the API server — it defaulted `.spec.strategy.rollingUpdate` on the live object, and `rollingUpdate` "may not be specified when strategy `type` is 'Recreate'". Under **Helm 4** the default server-side apply does not clear it and `--force-conflicts` does not help, so run that one switch with `helm upgrade --server-side=false`; **Helm 3** patches client-side and needs no flag. Deleting the Deployment and letting Helm recreate it also works.
+> `maxUnavailable` must stay non-zero — the API server rejects `maxSurge` and `maxUnavailable` both being `0`. This applies cleanly to an already-installed release under both Helm 3 and Helm 4, with no extra flags, because it replaces the API-defaulted `rollingUpdate` block rather than trying to remove it. It costs a gap in service while the old pod terminates, plus one transient `FailedAttachVolume`/`Multi-Attach` event per rollout — the scale-down and the scale-up are issued in the same second, so the replacement is created while the CSI driver is still detaching. It clears once the detach completes and the rollout converges without intervention. (This is what `mcp-mesh-grafana` and `mcp-mesh-tempo` ship as their default, since both are single-replica with a ReadWriteOnce claim enabled out of the box.)
+>
+> This only makes ReadWriteOnce safe at one replica. `persistence` creates a **single** claim for the whole release, so with `replicaCount > 1` or an active HPA, every replica scheduled off the claim's node stays `Pending` no matter what strategy is set — use ReadWriteMany, or keep the agent at one replica.
 
-> **Upgrade note:** `podSecurityContext` (runAsNonRoot, uid/gid 999, fsGroup 999, RuntimeDefault seccomp) is now always applied — it was previously gated behind the removed `podSecurityPolicy.enabled` flag and never rendered. It covers user-supplied `initContainers` (an init container that must run as root needs its own `securityContext`, which overrides the pod level), and `fsGroup: 999` changes group ownership of mounted volumes, including PVCs from `extraVolumes`.
+> **Upgrade note:** `podSecurityContext` (runAsNonRoot, uid/gid 999, fsGroup 999, RuntimeDefault seccomp) is now always applied — it was previously gated behind the removed `podSecurityPolicy.enabled` flag and never rendered. It covers user-supplied `initContainers` (an init container that must run as root needs both `runAsUser: 0` and `runAsNonRoot: false` in its own `securityContext`, because a container-level context overrides only the fields it sets — `runAsUser: 0` alone still inherits the pod's `runAsNonRoot: true`, and the kubelet refuses to start it with `container has runAsNonRoot and image will run as root` — and only where the namespace's Pod Security Standards level and any cluster policy permit running as root), and `fsGroup: 999` changes group ownership of mounted volumes, including PVCs from `extraVolumes`.
 
 ### Agent Code Configuration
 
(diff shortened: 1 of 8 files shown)


dhyansraj:
Verified both review findings against current code in `d7563374d`: **one fixed, one rejected.**

## Fixed — the init-container guidance invited a pod that never starts

`helm/mcp-mesh-agent/README.md:128` said an init container needing root "needs its own `securityContext`, which overrides the pod level". True but incomplete: a container-level context overrides **only the fields it sets**, so `runAsUser: 0` alone still inherits the pod's `runAsNonRoot: true` and the kubelet refuses with `container has runAsNonRoot and image will run as root`. Both fields are required.

Two things found while verifying that make the trap specific to init containers here:

- `templates/deployment.yaml:69-72` renders `initContainers` as a bare `toYaml` passthrough with **no** chart-injected container `securityContext` — unlike the main container at `:75-76`, which gets `.Values.securityContext`. So a user init container inherits the pod-level context wholesale, which is exactly why the partial-override trap bites there and not on the app container.
- `scripts/check_helm_pss.py` targets the `restricted` profile, so the "where the namespace's PSS level and cluster policy permit it" caveat the review asked for is warranted rather than hypothetical.

Also confirmed the note's existing claims still match the chart: `values.yaml:77-83` is `runAsNonRoot: true`, uid/gid 999, `fsGroup: 999`, `RuntimeDefault` — no drift, so that clause was left alone.

This paragraph is pre-existing rather than introduced here. Happy to split it into its own PR if you would rather keep this branch to rollout ordering; it renders no template change either way.

## Rejected — re-documenting redis persistence would undo a prior fix

The comment asks the redis rollout-strategy comment to reference chart-managed persistence via `persistence.enabled: true` + `accessMode: ReadWriteOnce`.

`helm/mcp-mesh-redis/templates/` contains only `_helpers.tpl`, `deployment.yaml` and `service.yaml` — **no PVC template anywhere in the chart** — while `deployment.yaml:89-96` mounts `persistentVolumeClaim: claimName: {{ include "mcp-mesh-redis.fullname" . }}` when `persistence.enabled` is true. So that path strands the pod `Pending` on a claim nothing creates.

That is tracked as **#1412**, and this exact comment was rewritten in #1410 specifically to stop teaching the path. Re-adding the reference would reintroduce it. `git diff` on that file is empty.

Worth noting this is the third review to reach for redis persistence as a working feature — the chart exposes `persistence.enabled`, `storageClass`, `accessMode`, `size` and `annotations` plus a `claimName` reference, i.e. everything except the template that would make it real. It reads as functional from every angle except the one that matters, which is a decent argument for resolving #1412 by deleting the dead surface.

## Validation

- Default render **byte-identical** to `853641b10` — sha256 `23362a48…e506f3`, 1148 lines both sides. README-only change, so any delta would mean something unintended was touched. Fresh throwaway worktrees per side, each with its own `helm dependency update`, both passwords pinned.
- `git diff --name-only 853641b10 HEAD` → `helm/mcp-mesh-agent/README.md` only.

Issue #1423, opened 2026-07-30
helm: mcp-mesh-tempo ships a 5Gi RWO claim for a 1-hour rolling buffer holding ~3MB


dhyansraj:
## Summary

`mcp-mesh-tempo` ships `persistence.enabled: true` with a **5Gi ReadWriteOnce** claim, while its own configured retention is **1 hour**. The volume is a rolling buffer, not durable storage, and its size and access mode are both out of proportion to what it holds.

```yaml
# helm/mcp-mesh-tempo/values.yaml
persistence:
  enabled: true
  size: 5Gi
  accessModes:
    - ReadWriteOnce
...
retention: "1h"
blockDuration: "5m"
```

## Measured on a live install

Pod up 25 hours, `happyfeet` namespace:

```
3.1M    /var/tempo              <- 0.06% of the 5Gi claim
772.0K  /var/tempo/blocks       2.3M  /var/tempo/wal
blocks/single-tenant/: 4 block dirs, all mtime 04:23-04:28  (now 04:54)
```

Oldest block **31 minutes old**. Retention is working exactly as configured. The claim therefore buys, permanently and by design, ≤1 hour of trace history and ~3MB across a restart.

## Why it is worth reconsidering

The RWO claim is not free. It is what made this chart deadlock on every pod-template change on a multi-node cluster — the defect fixed in #1411/#1418. With `persistence.enabled: false` tempo would use an `emptyDir`, and the RWO constraint that caused that bug would not exist at all.

So the chart is paying a real operational cost — an RWO volume with its attach/detach and rollout-ordering consequences — to preserve at most an hour of traces across a restart.

## The catch, and why this is not a trivial flip

Turning it off on an existing install drops the PVC from the rendered manifest, and with no `helm.sh/resource-policy: keep` on the live object Helm deletes it on the next upgrade. The data lost is worth ~3MB and ≤1h — genuinely trivial — but it is still a silent deletion during a routine upgrade, which is the pattern #1382/#1413 and #1416 exist to eliminate.

A default change here would want the same sequencing discipline: either accept the deletion explicitly and say so in the release notes, or annotate first and flip later.

## Options

- **(a) Default `persistence.enabled: false`.** Honest about what the volume is, and removes the RWO constraint. Costs ≤1h of traces across a restart, and needs the sequencing note above.
[...]

Issue #1421, opened 2026-07-30
CI runs no Helm validation on pull requests — lint and PSS render only fire on a release tag


dhyansraj:
## Summary

No workflow runs any Helm validation on a pull request. `helm lint` and `scripts/check_helm_pss.py` execute **only** in `helm-release.yml`, which triggers on a release tag:

```yaml
on:
  push:
    tags:
      - "v*"
  workflow_dispatch:
```

`grep helm .github/workflows/ci.yml` returns nothing.

So a chart change merges to `main` with zero automated verification, and the first automated check it faces is the release that publishes it.

## Evidence this is a live gap, not a theoretical one

Five chart PRs merged in the last day — #1410, #1413, #1415, #1417, #1418 — every one of them with a green check wall that included **no chart validation at all**. They touched templates in `mcp-mesh-core`, `-agent`, `-grafana`, `-tempo`, `-redis`, added two `fail`-based guards, changed a Deployment `strategy` block, and deleted a template.

Every defect in those PRs was caught by hand — local `helm lint`, `helm template` diffs against the merge base, and live installs on a cluster. That worked, but it is a property of how those PRs happened to be reviewed, not a property of the pipeline. Nothing would have stopped a template typo that produces valid-but-wrong YAML, and `helm lint` does not catch those either — but PSS rendering does, and neither ran.

The failure mode is specific: a chart defect reaches `main`, sits there, and surfaces during a release, which is the worst moment to discover it. That is exactly how the v3.3.1 Java publish broke (#1379) — a defect that only the release exercised.

## What to add

The jobs already exist and take seconds. Run them on PRs that touch `helm/`:

- `helm lint` across `helm/mcp-mesh-*`
- `python3 scripts/check_helm_pss.py` (renders every chart with default values and asserts the PSS `restricted` profile)

Two additions worth considering while wiring it:

- **Render the umbrella and diff against the merge base**, failing only on unexpected deltas. Several of the recent PRs used exactly this check by hand and it caught real problems. Note the trap: `helm/*/charts/*.tgz` is gitignored, so a baseline built without its own `helm dependency update` silently compares stale subchart archives — the check must build the baseline in a clean worktree.
- **Render with a matrix of representative values**, not just defaults. Both guards added recently (`validateNamespaceResourcePolicy`, the grafana `generatedSecret` guard) are invisible to a defaults-only render, since they are inert on the default path.

## Related

[...]

Issue #1419, opened 2026-07-30
helm: mcp-mesh-grafana installed from source never starts — dashboards ConfigMap is gitignored but mounted unconditionally


dhyansraj:
## Summary

Installing `helm/mcp-mesh-grafana` **from a clone of this repository** produces a pod that never starts:

```
MountVolume.SetUp failed for volume "dashboards":
  configmap "<release>-mcp-mesh-grafana-dashboard" not found
```

The released chart tarball is unaffected. This only bites people testing from source — which is everyone developing on the charts.

## Cause

`helm/mcp-mesh-grafana/templates/configmap-dashboard.yaml` builds its data from `files/dashboards/*.json`. That directory is **gitignored** (`.gitignore:256`) and populated by the release workflow's dashboard-sync step, so in a clone the glob matches nothing and the ConfigMap renders with no `data:` — or not at all.

`templates/deployment.yaml` mounts that ConfigMap **unconditionally**, so the volume reference dangles and the kubelet blocks the pod indefinitely.

The two halves disagree about whether dashboards are optional: the ConfigMap's content is conditional on files that may not exist, the mount is not.

## Reproduce

```bash
git clone <repo> && cd mcp-mesh
helm install g helm/mcp-mesh-grafana -n <ns> --create-namespace
kubectl -n <ns> describe pod -l app.kubernetes.io/name=mcp-mesh-grafana
# MountVolume.SetUp failed for volume "dashboards": configmap ... not found
```

Workaround, which is what chart testing has been doing: `--set grafana.dashboards.enabled=false`.

## Why it matters beyond developer friction

- It makes the chart untestable from source without a flag that a first-time contributor has no reason to guess, and the failure names a missing ConfigMap rather than a missing file — so the cause is not discoverable from the error.
- `helm/mcp-mesh-grafana/files/` and `helm/mcp-mesh-tempo/files/` being gitignored means `helm template` on a fresh clone also renders differently from the published chart, which undermines local verification generally. Any "byte-identical render" check done from a clone is comparing something the release does not ship.

## Options

**(a) Gate the mount on the same condition as the content.** If the ConfigMap has no data — or if `files/dashboards/` is empty — do not render the volume or the mount. Smallest change, and makes source and released behaviour differ only in whether dashboards appear, not in whether the pod starts.

[...]

PR #1426, opened 2026-07-30
fix(helm): keep Grafana's data volume on uninstall

### LINKED ISSUE #1416: helm: mcp-mesh-grafana's PVC is deleted on uninstall, silently destroying dashboards — while the bundled postgres volume survives

dhyansraj:
## Summary

`helm/mcp-mesh-grafana/templates/pvc.yaml` was a plain release-owned PVC with no `helm.sh/resource-policy`, so `helm uninstall` deleted it along with `grafana.db`. The bundled PostgreSQL keeps its data on a `volumeClaimTemplates` PVC the release does not own, so its volume survives. The component holding **derived** state retained its volume; the component holding **authored** state discarded it.

This adds `keep` to the Grafana data volume and documents what uninstall now leaves behind.

## Framing

This is a contract fix, not a data-loss recovery. A survey of the live installs found **no authored Grafana state** — one ConfigMap-provisioned dashboard, one default admin user, zero annotations, an identical 1.2M `grafana.db` on each. The harm was latent.

The argument that holds regardless: `persistence.enabled` defaults to `true`, and a volume the chart enables by default should not be discarded by a command that reports success. The Secret in this same chart is already `keep`, so this makes the chart internally consistent — the credential survived while the data it unlocked did not.

## Tempo deliberately untouched

Measured on a live 25-hour-old install: **3.1M on a 5Gi claim**, oldest block **31 minutes** old, `retention: "1h"` actively pruning. It is a rolling buffer, not durable storage, so `keep` would be wrong there. Whether it should have a PVC at all is #1423.

Verified live rather than assumed — see validation (4).

## Also fixed: a recipe this change silently breaks

`helm/mcp-mesh-grafana/values.yaml` offered "reprovision from an empty PVC" as a way to reset the admin password. That recipe worked *because* uninstall deleted the volume; once the PVC is kept it stops working, with no error. Rewritten here, along with the same recipe in `helm/mcp-mesh-core/README.md`.

The README gained a **"What uninstall leaves behind"** table, since the answer is now genuinely mixed across the bundled components.

## Validation

CI runs no helm jobs on a PR (#1421), so this is local plus live cluster, on **Helm 3.19.0 and 4.0.1**.

1. **Annotation lands via plain `helm upgrade`, no pod churn.** Baseline installed from a throwaway worktree at `d385a43a8`, marker seeded into the volume, upgraded with no extra flags: same pod uid (`8998eb93…` h4, `f28b74ea…` h3), same PVC uid, live annotations gained `keep`. Unlike the Namespace case, the PVC is never dropped from the manifest, so there is nothing for SSA to clear.
2. **Survives uninstall**, both versions:
   ```
   These resources were kept due to the resource policy:
   [PersistentVolumeClaim] gf-mcp-mesh-grafana-pvc
   [Secret] gf-mcp-mesh-grafana-secret
   ```
   A scratch pod mounting the claim read the marker back, alongside a retained `grafana.db`.
3. **Re-adopted on reinstall** under the same release name, marker intact, same PVC uid.
4. **Tempo still deleted.** Full `mcp-mesh-core` install then uninstall: remaining PVCs were exactly `mcp-core-mcp-mesh-grafana-pvc` and `postgres-data-mcp-core-mcp-mesh-postgres-0`; `mcp-core-mcp-mesh-tempo-pvc` gone.
5. **Default render delta vs `d385a43a8` is two lines**, on one PVC:
   ```
   @@ -306,6 +306,8 @@  charts/mcp-mesh-grafana/templates/pvc.yaml
   +  annotations:
   +    "helm.sh/resource-policy": keep
   ```
   Tempo's PVC in the same render carries no `annotations:` block. Baseline from a fresh worktree with its own `helm dependency update` — `helm/*/charts/*.tgz` is gitignored, so a stale archive would silently mask a delta.
6. `helm lint` grafana + core: 0 failed. `check_helm_pss.py`: all 9 OK.

## Corrections made while validating

- **The postgres claim is `postgres-data-<sts>-0`, not `data-<sts>-0`** (`volumeClaimTemplates[0].metadata.name: postgres-data`). Caught against live PVCs; the README table and reclaim snippet were wrong before being corrected.
- **A first Helm 3 run produced a false PASS** — `kubectl exec` raced container start, the marker was never written, and an empty-vs-empty comparison "passed". Re-run with a retry loop, explicit `-c grafana`, and a non-empty guard. The numbers above are from the corrected run.
- `mcp-mesh-core.grafanaSecretName` had no fullname helper to build on, so NOTES could not name the PVC or Deployment. Extracted `mcp-mesh-core.grafanaFullname` rather than adding a third copy of the release-name derivation; truncation semantics are identical.

## Note for reviewers

Probes used `--set grafana.dashboards.enabled=false` — installing this chart from a source clone wedges on a missing dashboards ConfigMap, because `files/dashboards/` is gitignored and populated only at release time while the Deployment mounts it unconditionally. That is **#1419**, unrelated to this change.

Closes #1416

## Summary by CodeRabbit

* **New Features**
  * Grafana data is now preserved when uninstalling the Helm release, including dashboards, users, API keys, and credentials.
  * Reinstalling with the same release name can reuse the existing Grafana data.

* **Documentation**
  * Added guidance for reclaiming Grafana storage, including required scaling and PVC deletion steps.
  * Clarified which storage remains after uninstall and how to recover from failed installations.
  * Documented that Tempo storage behavior is unchanged and its PVC is still removed on uninstall.

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -155,6 +155,18 @@ helm get values <release>
   derived is application data: if your own workloads used the bundled PostgreSQL
   or Redis for their own storage, that data is yours to protect before you tear
   anything down.
+- **`helm uninstall` no longer reclaims Grafana's data volume.** The Grafana
+  PVC picks up a `"helm.sh/resource-policy": keep` annotation on upgrade, so
+  Helm skips it on uninstall — `grafana.persistence.enabled` defaults to `true`,
+  and the dashboards, annotations, users, and API keys in `grafana.db` are not
+  derived from anything the mesh can replay. Previously that volume was deleted
+  along with the release. A reinstall under the same release name adopts the
+  existing claim, data intact; reclaim the storage deliberately with
+  `kubectl delete pvc <release>-mcp-mesh-grafana-pvc -n <ns>`. This lands on a
+  plain `helm upgrade` with no pod restart — the claim never leaves the rendered
+  manifest, so it is a metadata-only patch. Tempo's PVC is unchanged and still
+  deleted on uninstall: it is a rolling trace buffer under active retention, not
+  durable storage.
 - **The core release owns its `Namespace`** (`namespaceCreate`, default
   `true`). Older charts let `helm uninstall` delete that namespace and cascade
   to every co-located agent, Secret, and PVC. Upgrading fixes this with no
--- a/helm/mcp-mesh-core/README.md
+++ b/helm/mcp-mesh-core/README.md
@@ -19,12 +19,12 @@ This umbrella chart deploys the core MCP Mesh infrastructure components:
 
 ```bash
 # Install core infrastructure
-helm install mcp-mesh-core ./mcp-mesh-core \
+helm install mcp-core ./mcp-mesh-core \
   -n mcp-mesh --create-namespace \
   --set namespaceCreate=false
 
 # Or with custom values
-helm install mcp-mesh-core ./mcp-mesh-core \
+helm install mcp-core ./mcp-mesh-core \
   -n mcp-mesh --create-namespace \
   --set namespaceCreate=false \
   -f my-values.yaml
@@ -43,7 +43,7 @@ that already exists, and for what applies to a release that already exists.
 
 ```bash
 # Port forward to access registry
-kubectl port-forward -n mcp-mesh svc/mcp-mesh-core-mcp-mesh-registry 8000:8000
+kubectl port-forward -n mcp-mesh svc/mcp-core-mcp-mesh-registry 8000:8000
 
 # Check health
 curl http://localhost:8000/health
@@ -98,7 +98,7 @@ and install with `--take-ownership` (Helm 3.17+ and Helm 4):
 
 ```bash
 kubectl create namespace mcp-mesh
-helm install mcp-mesh-core ./mcp-mesh-core \
+helm install mcp-core ./mcp-mesh-core \
   -n mcp-mesh --set global.namespace=mcp-mesh --take-ownership
 ```
 
@@ -127,14 +127,14 @@ was rendered by 3.4.0 or later — two separate upgrades:
 
 ```bash
 # 1. Upgrade to >= 3.4.0, leaving namespaceCreate alone.
-helm upgrade mcp-mesh-core ./mcp-mesh-core -n mcp-mesh --reuse-values
+helm upgrade mcp-core ./mcp-mesh-core -n mcp-mesh --reuse-values
 
 # Confirm the live namespace picked up the policy before going further.
 kubectl get ns mcp-mesh -o jsonpath='{.metadata.annotations}'
 #   {"helm.sh/resource-policy":"keep","meta.helm.sh/release-name":...}
 
 # 2. Only then, and only if you want it, drop the Namespace from the release.
-helm upgrade mcp-mesh-core ./mcp-mesh-core -n mcp-mesh --reuse-values \
+helm upgrade mcp-core ./mcp-mesh-core -n mcp-mesh --reuse-values \
   --set namespaceCreate=false
 ```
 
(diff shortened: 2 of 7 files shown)


dhyansraj:
Verified the finding against current code. **Real inconsistency, but the proposed direction was backwards** — fixed the other way in `6f467050d`.

## Which release name is canonical

The comment asks to change the reclaim commands from `mcp-core-…` to `mcp-mesh-core-…`, on the grounds that `mcp-mesh-core` is the documented active release name. Measured across the surfaces #1417 rewrote:

| surface | release name |
|---|---|
| `src/core/cli/man/content/deployment.md` | **9× `mcp-core`**, 0× other |
| `docs/04-kubernetes-basics.md` | **5× `mcp-core`** |
| `helm/mcp-mesh-core/README.md` (before) | 4× `mcp-mesh-core`, 1× `mcp-core` |

And this README's own scale-down command at `:391` already read `kubectl scale deploy/mcp-core-mcp-mesh-grafana`.

So the reclaim commands were **already correct**, and the install examples above them were the stale outlier. Applying the comment as written would have moved the one consistent part into the inconsistency.

Fixed the other way: 11 sites swept to `mcp-core` — the four install examples, plus `helm upgrade` (`:130`, `:137`), `helm uninstall` (`:164`, `:645`) and a derived Service name (`:46`) that had the same drift. The README is now `mcp-core` ×27 as a release name, ×0 for the old form. The eight remaining `mcp-mesh-core` occurrences are all chart *paths* (`./mcp-mesh-core`), which are correct in the chart's own README. Placeholder forms like `&lt;release&gt;-mcp-mesh-grafana-pvc` were left as placeholders.

## The scale-down link — the existing procedure only half-applies

The second half of the comment asks to replace the generic scale-down note with exact commands for the Grafana Deployment and the PostgreSQL StatefulSet, or link to the existing instructions. Checking before linking was worthwhile:

- `helm/mcp-mesh-grafana/templates/deployment.yaml:14` → `kind: Deployment`
- `helm/mcp-mesh-postgres/templates/statefulset.yaml:3` → `kind: StatefulSet`, claim template `postgres-data` (`:125-127`)

The difference is not just the noun — **the recreation semantics diverge.** Grafana's PVC is a plain templated object, which is why the `:391` procedure needs a third `helm upgrade` step to recreate the claim. The StatefulSet's `volumeClaimTemplates` recreates `postgres-data-…-0` by itself on the next scale-up, with no `helm upgrade` at all.

So a bare link would have pointed PostgreSQL readers at a step that is wrong for their case. Linked for Grafana, and stated the StatefulSet case inline — one scale target plus the claim-template note. That is new information rather than a second copy of `:391`, which is the duplication that drifts.

## Validation

- Default render **byte-identical** to `8db6c98ea` — sha256 `7f16f29c…0da837`, 1150 lines both sides. README-only change, so any delta would mean something unintended was touched. Fresh detached worktrees per commit with their own `helm dependency update`, both generated passwords pinned.
- `git diff --name-only 8db6c98ea HEAD` → `helm/mcp-mesh-core/README.md` only.
- 3/3 anchor links resolve.

Issue #1404, opened 2026-07-29
examples: orphaned requirements.txt ships stdlib-shadowing 'asyncio'; observability Dockerfile.dev is uneditable due to pre-existing hadolint debt


dhyansraj:
Two unrelated pieces of stale example debt, both surfaced while fixing #1402 and deliberately left out of #1403 to keep a release-path PR scoped.

## 1. `examples/docker-examples/agents/base/requirements.txt` is orphaned, and ships a stdlib-shadowing dependency

Nothing in the repository reads this file — `grep -rln "base/requirements"` returns nothing. Neither sibling Dockerfile consumes it:

- `Dockerfile.base` does `COPY . /src/mcp-mesh` then `pip install src/runtime/python/`
- `Dockerfile.dev` pip-installs its own hardcoded list

So it is dead metadata that a user could still `pip install -r` by hand — and if they do, they get:

```
asyncio
```

which resolves to the **PyPI `asyncio` package (4.0.0), shadowing the stdlib module**. That is a live footgun in a file nothing validates, which is precisely why it has gone unnoticed.

Its `fastapi>=0.104.0` floor was swept to `>=0.135.0` in #1403 for consistency, but that treats a symptom.

**Proposed:** delete the file. If it is meant to be a template, it needs an owner and something exercising it in CI; otherwise it is a trap that only degrades further.

## 2. `examples/observability-test/Dockerfile.dev` carries pre-existing hadolint failures

The file cannot be committed at all without tripping the `hadolint` pre-commit hook:

```
:9  DL3008 warning: Pin versions in apt get install
:9  DL3015 info:    Avoid additional packages by specifying --no-install-recommends
:40 DL3059 info:    Multiple consecutive RUN instructions
```

All three reproduce against unmodified `HEAD`, so they are not introduced by any pending change. The file simply has not been committed since `.pre-commit-config.yaml` was repaired to actually match `Dockerfile.*` — its own comment records that the hook previously mis-parsed `entry:` and linted nothing. There is no `.hadolint.yaml`, and the default failure threshold is `info`, so all three block.

This makes the file **effectively uneditable**: any one-line change drags in unrelated lint debt.

[...]

Issue #1392, opened 2026-07-28
meshctl man: inline code corrupted by italic pass, and list items never styled


dhyansraj:
## Summary

Two rendering defects in `meshctl man`, both in `src/core/cli/man/renderer.go`. Source markdown is correct in every case; the renderer mangles it.

## 1. Inline code spans are corrupted by the bold/italic passes

`styleInline` runs its passes in this order:

```go
line = inlineCodeRe.ReplaceAllString(...)    // `code` -> ANSI; content left as plain text
line = inlineBoldRe.ReplaceAllString(...)    // **x** / __x__
line = inlineItalicRe.ReplaceAllString(...)  // *x* / _x_   <- \*([^*]+)\*|_([^_]+)_
line = inlineLinkRe.ReplaceAllString(...)    // [text](url)
```

Once the backticks are stripped, code-span content is ordinary text, so a line carrying two snake_case code spans has its underscores paired **across** them by the italic pass. Since mcp-mesh uses snake_case capability names throughout, this is pervasive.

`man audit`, from `audit.md:7`:

```
source:   ... as a `dependency_resolved` (or `dependency_unresolved`) event ...
rendered: ... as a ^[[32mdependency^[[3mresolved^[[0m (or ^[[32mdependency^[[0munresolved^[[0m) ...
```

The underscores are consumed and replaced with ANSI italic. Users see `dependency`*`resolved`*.

Same root cause hits links: `inlineLinkRe` runs last, so a URL containing underscores gets ANSI injected mid-URL by the italic pass and then no longer matches as a link. A second instance was found in `llm.md:61`, where the bold pass emitted `\x1b[1m` and the literal `[1m` was subsequently swallowed as link text.

Measured across all topics: **83–85 corrupted code spans**.

## 2. List items are never styled at all

In `renderStyled`, the `- `, nested `  - `, and numbered-list branches write the line raw and `continue`, so `styleInline` is never reached. Markup renders literally:

```
source:   - For `dependency_resolved`:
rendered:   • For `dependency_resolved`:      <- literal backticks, no colour

source:   - **Orphan reroute.** A job whose owner replica is gone
rendered:   • **Orphan reroute.** A job whose owner replica is gone   <- literal asterisks
```

**774 list lines across 65 content files** carry `code` or `**bold**`; 432 are in default-variant topics, the rest reachable via `--typescript` / `--java`. This is the larger of the two defects.

## Fix

[...]

PR #1428, opened 2026-07-30
fix(release): targeted Cargo.lock refresh, CI at the declared FastAPI floor, asserted corpus counts

### LINKED ISSUE #1407: bump_version.py reminder tells operators to run 'cargo generate-lockfile', which silently re-resolves the whole Rust dependency graph

### LINKED ISSUE #1409: CI's 'older FastAPI' job pins 0.136.1 while the declared floor is 0.135.0 — plus a missed 0.139→0.137 site and unasserted renderer counts
Three small, related gaps surfaced while reviewing the v3.3.2 release notes. Kept out of the release PR so it stays bump-plus-notes.

## 1. The "older FastAPI" CI job does not test the declared floor

`.github/workflows/ci.yml:295-306` — job *"Route integration tests against older FastAPI"* — pins:

```
pip install "fastapi==0.136.1"
```

The declared floor is now `fastapi>=0.135.0` (#1402, both manifests). So the lowest version a user can legally install is **below** the lowest version CI exercises. The claim that 0.135.0 is supported rests on a one-off manual sweep, not on anything that runs again.

That is the same shape as the defect #1402 fixed: a supported-on-paper, untested-in-CI band. It is only two versions wide, but it is exactly the band where the auto-streaming attributes the route rebuild depends on are half-present (`is_json_stream` lands in 0.134.0, `is_sse_stream` in 0.135.0), so it is the band most likely to behave differently.

**Proposed:** move the pin to `fastapi==0.135.0` so the job tests the actual floor, and make the pin derive from (or be asserted against) the declared floor so the two cannot drift again. A test that asserts the CI pin equals the manifest floor would make this structural rather than remembered.

## 2. `ci.yml:289` still carries the superseded 0.139 figure

The comment reads *"0.139's include_router() change"*. Lazy `include_router` (`original_router`) actually appears at **0.137.0** — empirically 0.136.3 flattens, 0.137.0 is lazy.

#1403 corrected this across 8 sites; this is a 9th it missed. No functional impact — both shapes are handled — but it now contradicts the corrected figure carried in `fastapi_routes.py`, `route_integration.py:739`, `server_discovery.py:304`, `mesh/a2a.py:1472`, `test_route_include_router_1396.py` and the v3.3.2 release notes.

## 3. `renderer_test.go` logs its corpus counts instead of asserting them

The man-renderer corpus test emits its span/list counts via `t.Logf` rather than asserting on them. The counts quoted in #1392 and in the v3.3.2 release notes (83 corrupted code spans, 432 unstyled list lines) therefore have **no in-tree anchor** — nothing fails if those numbers change.

The test still catches regressions through its other assertions, so this is not a hole in coverage so much as a measurement that cannot be re-derived. Given the figures were published in release notes, an assertion (or a golden count) would make them verifiable rather than historical.

## Acceptance

- The FastAPI version CI exercises is the version the manifests declare as the floor, and cannot silently drift from it.
- No source or CI comment states a superseded `include_router` version.
- Counts quoted in public artifacts are asserted somewhere that fails when they change.

dhyansraj:
## Summary

Four release-tooling defects, bundled because they share one test suite. Three are the same shape: **a check exists but cannot fail.**

## 1. The bump reminder told operators to re-resolve the whole Rust graph (#1407)

`scripts/bump_version.py` printed `run 'cargo generate-lockfile'`, which re-resolves every dependency. Measured on a real 3.3.2 → 3.3.3 bump in a throwaway worktree:

| command | `Cargo.lock` numstat | third-party crates moved |
|---|---|---|
| `cargo generate-lockfile` | `25  25` | **11** |
| `cargo update --package mcp-mesh-core` | `1  1` | **0** |

The reminder now names the targeted command, says explicitly not to use the other one, and points at a new verifier.

**`scripts/check_release_lockfiles.py`** parses `Cargo.lock` / `Chart.lock` before and after and fails if any non-mesh version moved. Gated on `mcp-mesh-core`'s own version having moved, so a deliberate dependency-bump PR is skipped rather than blocked. Run retroactively against every release boundary, it reproduces this issue's history exactly:

```
v3.2.2 -> v3.2.3:  6 third-party moves   gate=FAIL
v3.2.3 -> v3.3.0:  0                     gate=PASS
v3.3.0 -> v3.3.1:  6                     gate=FAIL
v3.3.1 -> v3.3.2:  0                     gate=PASS
```

It catches both incidents and stays quiet on the two clean releases.

Note the count is not stable: #1407 recorded 8 crates during the 3.3.2 prep, the same bump moves **11** today. Nothing in the repo changed — crates.io moved on. A release step whose output depends on when you run it is the defect.

## 2. CI tested above the declared floor (#1409a)

`ci.yml` pinned `fastapi==0.136.1` while both manifests declare `>=0.135.0`. The lowest installable version was below the lowest tested one — the supported-on-paper/untested-in-CI shape #1402 exists to eliminate, reintroduced one layer up.

Pin moved to `0.135.0`, and **`scripts/test_ci_fastapi_floor.py` now asserts the pin equals the declared floor**, so they cannot drift again.

Verified the move does not turn CI red: the four route test files run against fastapi 0.135.0 give **85 passed, 1 skipped** — the skip being `FastAPI flattens included routers; no cache to go stale`, which independently corroborates the 0.137.0 figure below.

## 3. A missed `0.139` → `0.137.0` site (#1409b)

`ci.yml:289` still attributed the lazy `include_router` change to 0.139. It landed in **0.137.0** (0.136.3 flattens, 0.137.0 is lazy). #1403 corrected 8 sites; this was the 9th. The floor test also asserts this attribution.

## 4. Corpus counts were logged, not asserted (#1409c)

`renderer_test.go` emitted its counts via `t.Logf`, so the figures published in #1392 and the v3.3.2 release notes had no in-tree anchor. Now golden assertions plus a new `TestManCorpusMarkupListLineCount`.

Two honest notes on those numbers:

- **The published `432` was correct and has since drifted.** At `428dd1eb3` the corpus measured exactly 432; it is **433** today because #1413 and #1417 added man content. Verifiable in principle, stale in practice — which is the argument for the golden.
- **The `83` corrupted-code-spans figure is not re-derivable** and no anchor was fabricated for it. It was a property of the pre-fix renderer; the fixed renderer corrupts zero. What is anchored is the corpus the measurement was taken over: 1605 spans / 488 list-item spans / 433 markup list lines.

## 5. Guard and handler disagreed about `3.3.1rc1` (surfaced by #1427)

Seven raw image-tag handlers used `(?![\d.\-+])`, which **admits letters** — so a tag `mcpmesh/cli:3.3.1rc1` would be rewritten by a 3.3.1 bump. Meanwhile `test_bump_version.py:69` already asserted the *guard* ignores that form.

The guard was right. Handlers widened to `(?![\w.\-+])`, verified exhaustively before changing — **all seven match identically before and after**:

| handler | files | before | after |
|---|---|---|---|
| Scaffold Dockerfile.tmpl | 15 | 15 | 15 |
| Markdown | 185 | 95 | 95 |
| Dockerfiles | 194 | 191 | 191 |
| docker-compose | 18 | 21 | 21 |
| Go handler source | 3 | 6 | 6 |
| Scaffold source | 3 | 16 | 16 |
| language_test.go | 1 | 2 | 2 |

Every real tag in the tree (`3.3.2`, `latest`, `custom-tag`, `0.5`, `0.5.1`, `0.5.2`, `3.2.2`) and every `@mcpmesh/*` specifier behaves the same. **Nothing stopped being bumped.**

## The finding nobody was looking for

Extending the meta-test past its pep440 scope exposed **three handlers with no terminal boundary at all** — `Documentation (--version OLD)`, `Docker Example Helm Values`, and `CI/CD Workflows (e.g., vOLD)`. A 3.3.1 → 3.3.2 bump against `--version 3.3.10` would have produced `--version 3.3.20`: exactly the `==3.3.10` → `==3.3.20` corruption the pip pins were fixed for.

Nothing is damaged today because every site carries the exact version. It went unlooked-for precisely because the meta-test's own scope line said the raw handlers were out of scope — a boundary in a test comment kept a defect invisible. **All 34 handlers are now terminally bounded.**

## Helm side, checked and asserted

`helm dependency update` cannot pull external chart versions here: all six `mcp-mesh-core` dependencies are `file://../<subchart>` at exact pins, no chart in `helm/` declares a non-`file://` repository, and every other chart is `dependencies: []`.

**Caveat worth recording:** the command *does* hit the network — it refreshes the operator's local repo cache (`harbor, jetstack, sealed-secrets, argo, bitnami`). None of that can reach the lock, but the safety is a property of today's `Chart.yaml`, not of the command. So the checker asserts it, and goes red the day someone adds a `bitnami/postgresql` dependency or a `^3.3` range.

## Validation

Every fix proven to fail before being trusted:

- **Floor assertion** — raising the manifest to `0.141.0` → 2 failed; reverting the CI pin to `0.136.1` → `CI pins fastapi==0.136.1 but ... declares a floor of 0.135.0`; restoring the `0.139` comment → `attributes the include_router change to ['0.139']; it landed in 0.137.0`.
- **Renderer goldens** — 1605→1604, 488→500, 433→432 each fail with a delta message.
- **Meta-test** — boundary deleted → fails; reverted to the narrow form → both checks fail; `--version OLD` left unbounded → `--version 3.3.1rc1` becomes `3.3.2rc1`, caught.
- **Lockfile gate** — clean run 3 ✅ exit 0; after `generate-lockfile`, `❌ Cargo.lock: 11 non-mesh version(s) moved alongside ours` exit 1.

Tallies: `test_bump_version.py` 32 → **37**; `pytest scripts/ -q` 32 → **56**; `go test ./src/core/cli/man/...` 35 → **36**. `go build`, `go vet` green.

**A real 3.3.2 → 3.3.3 bump is byte-identical to the pre-change baseline** — 467 files / 37 categories / 645 lines, both guards clean, and the two bumps' diffs compare to zero. Nothing newly stopped bumping.

## One scope call for review

`examples/docker-examples/agents/base/requirements.txt` is a **third** site declaring a fastapi floor. It is included in `test_every_declared_fastapi_floor_agrees`, so a split between the shipped example and the published contract also fails. Easy to drop if that is too strict — note this file is itself orphaned (#1404).

Closes #1407
Closes #1409

## Summary by CodeRabbit

* **Compatibility**
  * Updated the supported older FastAPI validation level to match the declared minimum version across project components.

* **Bug Fixes**
  * Improved version updates so they no longer accidentally modify longer or similarly named versions.
  * Release lockfiles now detect unintended dependency changes and enforce pinned chart dependencies.

* **Tests**
  * Expanded automated coverage for version updates, dependency integrity, FastAPI compatibility, and documentation rendering consistency.


dhyansraj:
Verified five findings against current code in `6dbab9607`: **three fixed, two skipped.**

## Fixed

**1. The lockfile gate could not fail — and that is the defect this PR is about.**

`_git_show` returned `None` for *both* an unresolvable ref and a path genuinely absent at a valid ref, and `_check_pair` turned that into `⏭ skipped` → `(True, False)` → **pass**. Any unfetched `origin/main`, `HEAD^` on a first commit, or typo made the gate report success without comparing anything. Reachable straight from the CI step, which builds `BASE="origin/$BASE_REF"`.

Added `RefError` + `_resolve_ref()` (`git rev-parse --verify --quiet '&lt;ref&gt;^{commit}'`). The ref is resolved before anything is read; `_git_show` now takes an already-resolved ref and returns `None` only for a genuinely missing path. Proven both ways:

```
$ python3 scripts/check_release_lockfiles.py --base origin/no-such-branch
❌ lockfile comparison could not run: base ref 'origin/no-such-branch' does not resolve to a commit.
   In CI this usually means the branch was not fetched (checkout needs fetch-depth: 0)...
exit=1
```

Valid ref with the path absent still skips at exit 0. And the new test was proven to **fail when the validation is removed** — replacing `ref = _resolve_ref(base)` with `ref = base` reproduces the exact silent-pass output the fix deletes. A test for "this cannot silently pass" that itself silently passed would have repeated the problem.

**2. `github.base_ref` now goes through step env** — at **both** sites, not just the one flagged: `lockfile-integrity` (~380) and the helm render diff (~468). No direct `${{ github.base_ref }}` interpolation into any `run:` remains in `.github/workflows/`. Worth being accurate about the severity: this workflow only triggers on PRs to `main`/`develop`, so `base_ref` is not attacker-controlled here — hardening and consistency with documented Actions practice, not a live injection fix.

**3. The image-name alternation is now a shared constant.** Promoted from nitpick because it is a real recurrence vector: the guard has had a shared `img` all along while seven handlers repeated the alternation inline, and that asymmetry is *how* guard and handler came to disagree about `3.3.1rc1`. The new meta-test catches boundary drift but not alternation drift, so a new image name added to one side and not the other would still slip through.

Both irregularities were checked before flattening, and **both are deliberate, so both were preserved**: `:797` covers runtime images only because all 15 scaffold `Dockerfile.tmpl` files use `{python,typescript,java}-runtime` as a base and registry/ui/cli are services, never bases; `:842`'s `FROM ` prefix exists so the compose handler at `:872` can deliberately catch bare comment references. No handler's match set was widened.

Proof of no behaviour change — two detached worktrees, real `3.3.2 → 3.3.3` bump in each:

```
6584 diff-before.patch   3f5155c8397bf6f7ed2ea83fdeff9e9789dc7086
6584 diff-after.patch    3f5155c8397bf6f7ed2ea83fdeff9e9789dc7086   ← identical sha1
```

Plus all seven handlers' match counts identical before/after, and every handler tuple and `_guard_patterns` output string-identical across `(3.3.2,3.3.3)`, `(3.3.2,3.4.0)` and `(3.3.2,None)`.

## Skipped

**Scoping `_ci_pin()` to the job block.** `test_ci_fastapi_floor.py` already asserts `len(pins) == 1`, which fails loudly on a second pin anywhere in the file — another job or a comment. Scoping to a block would *weaken* it: a stray pin outside the block would be silently ignored instead of failing. Fail-loud-on-ambiguity is the safer design, and it is what is there.

**Extracting a `_probe_handler` context manager.** Both tests already use `try/finally`, so there is no pollution bug — it is DRY across two call sites, adding indirection for no correctness gain.

## Validation

`pytest scripts/ -q` **56 → 59**; `test_bump_version.py` 37 checks; `test_check_release_lockfiles.py` 16 → 19. `actionlint` identical before and after — the same two pre-existing `constant expression "false"` findings at `ci.yml:18` and `:51`, nothing new, nothing raised on the `$BASE_REF` shell. Diff scope: exactly the four intended files.

PR #1433, opened 2026-07-30
fix(registry): bound telemetry aggregates and decouple the event cap from retention

### LINKED ISSUE #1424: Telemetry aggregates grow without bound — MetricsProcessor maps, accumulator edgeStats, and Redis consumer-group entries are never pruned
## Summary

The trace *data* pipeline is bounded — `MCP_MESH_TRACE_RETENTION` (default `24h`) trims the `mesh:trace` stream, and Tempo's `block_retention` bounds the durable store. What has **no bound at all** is the in-memory aggregate layer built on top of it, plus the Redis consumer-group bookkeeping around it.

None of these is reachable by any existing knob, and no cleanup loop visits them.

### 1. `MetricsProcessor` maps — never pruned

`src/core/ui/metrics_processor.go:14-22`:

```go
type MetricsProcessor struct {
	mu sync.RWMutex

	// Per-agent metrics keyed by agent name
	agentMetrics map[string]*agentMetrics

	// Per-model token usage keyed by model name
	modelMetrics map[string]*modelMetrics
}
```

Written on every processed event; **never deleted, never expired, no cap**. These back the dashboard's per-agent and per-model statistics, and the counters are all-time — the only reset is restarting meshui.

Cardinality is small in a stable mesh, which is why this has not bitten yet. It grows without limit under any agent-name churn: per-pod names, ephemeral or short-lived agents, version-suffixed names, or a model identifier that varies.

### 2. `TraceAccumulator.edgeStats` — never pruned

`src/core/registry/tracing/accumulator.go:55-58`:

```go
	// Per-edge stats keyed by "source -> target"
	edgeStats map[string]*edgeAccum
```

Each entry caps its latency samples (`maxEdgeLatencies = 10000`), but the **number of edge keys is unbounded**. Same cardinality exposure as (1), and it exists in both the registry and meshui processes.

### 3. Redis consumer-group bookkeeping — never cleaned

- **A dead consumer per pod restart.** `src/core/registry/tracing/consumer.go:143-145` names the consumer from hostname + PID:
  ```go
  hostname, _ := os.Hostname()
  pid := os.Getpid()
  consumerName = fmt.Sprintf("registry-%s-%d", hostname, pid)
  ```
  Every restart creates a new permanent entry in the group. There is **no `XGROUP DELCONSUMER`** anywhere in the repo.
- **Permanently-pending messages.** `XREADGROUP` is called with `">"` (new messages only), and there is **no `XAUTOCLAIM`, `XCLAIM`, or `XPENDING`** anywhere. A message whose processing fails is logged, skipped, never ACKed, and never re-read — it stays in the PEL indefinitely.

Note the code comment at `consumer.go` claiming "pending messages first, then new ones" is inaccurate given the `">"` argument.

## A related gap worth fixing alongside

**Trimming depends entirely on the registry.** The `XADD` itself is uncapped (`src/runtime/core/src/tracing_publish.rs:372` — no `MAXLEN`, no `MINID`), and `XTRIM MINID ~` is issued only by the registry consumer. meshui reads the same stream through its own consumer group but **never trims and never reads `MCP_MESH_TRACE_RETENTION`** (`manager_ui.go` builds its config with no retention field).

So if the registry is down, or has tracing disabled while agents keep publishing, nothing bounds the stream. The only remaining backstop is Redis `maxmemory` with `allkeys-lru` — and that is a **cliff, not a policy**: `mesh:trace` is a single key, so eviction drops the entire stream at once rather than shedding old entries, and with `appendonly: yes` session data goes with it.

## Proposed

1. **Bound the three maps.** The correlator already has the pattern to copy — `src/core/registry/tracing/correlator.go:388-396`, `pruneExpiredCompletedTraces`, age-based with a `retention <= 0` no-op. Either that or an LRU cap. Both `MetricsProcessor` maps and `edgeStats` want the same treatment, and it should be reachable by a knob rather than a constant.
2. **`XGROUP DELCONSUMER`** for consumers idle beyond a threshold, or a stable consumer name that does not embed the PID.
3. **Reclaim or drop stuck PEL entries** — `XAUTOCLAIM` with a retry limit, then acknowledge-and-log rather than leaving them pending forever.
4. **Make trimming not depend on one process** — either have meshui honour `MCP_MESH_TRACE_RETENTION` too, or cap at the producer with `XADD ... MAXLEN ~ N`. Producer-side is a single call site (`tracing_publish.rs:372`) since Python, TypeScript, Java and Go all publish through the same Rust core.

## Not in scope here

`OTLPExporter.activeSpans` / `pendingSpans` (`exporters.go:798,801`) are never `delete()`d and hold live contexts and spans — a genuine leak, but currently **latent**: only `EventType == "span_start"` reaches that path and no runtime emits it today. Worth its own issue if the dual-event shape is ever revived.

`JSONFileExporter` writes one unrotated file per trace, but is only reachable via a non-default `TRACE_EXPORTER_TYPE`.

## Acceptance

- No telemetry structure grows without a bound an operator can see and adjust.
- Stream trimming does not depend on a single process being alive.
- Consumer-group entries and pending messages do not accumulate across restarts.

### LINKED ISSUE #1425: MCP_MESH_RETENTION=0 silently removes the registry_events row cap along with the sweep
## Summary

Setting `MCP_MESH_RETENTION=0` disables the registry sweep job, which is documented and intended. What is not obvious is that it **also silently removes the 100,000-row cap on `registry_events`**, because the cap is enforced inside the sweep goroutine that never starts.

An operator who disables the sweep to preserve agent history for forensics unknowingly unbounds an unrelated table.

## Mechanism

The cap is a constant, `src/core/registry/sweep.go:36`:

```go
	eventMaxRows         = 100_000
```

It is enforced by `enforceEventCap`, which runs inside `runOnce` — i.e. inside the sweep loop. And the sweep returns before starting anything when retention is non-positive, `sweep.go:221-226`:

```go
	if s.cfg.Retention <= 0 {
		if s.logger != nil {
			s.logger.Info("🧹 Registry sweep job disabled (MCP_MESH_RETENTION=0)")
		}
		return
	}
```

So the two behaviours — "purge old agents/schemas" and "cap the events table" — share a single on/off switch, despite being unrelated concerns with different reasons to be disabled.

The log line is accurate about what the operator asked for and silent about what else they got.

## Why it matters

`registry_events` is append-heavy: it records registry lifecycle activity and backs the meshui Events page. Losing its row cap on a busy mesh means unbounded table growth in Postgres, discovered whenever someone notices disk usage or query latency — with nothing connecting it back to a retention setting changed for an unrelated reason.

The two plausible reasons to set `MCP_MESH_RETENTION=0` — debugging agent lifecycle, or preserving history for an incident — are both cases where the operator specifically does *not* want unbounded growth elsewhere.

## Proposed

Decouple them. Options, roughly in order of preference:

1. **Enforce the event cap independently of the sweep's retention setting.** The cap is a safety limit on table size, not a retention policy; it has no reason to be conditional.
2. **Give the cap its own configuration**, so it can be raised, lowered, or deliberately disabled on purpose rather than as a side effect.
3. At minimum, **say so in the log line** — "sweep disabled; the registry_events row cap is no longer enforced" — so the consequence is visible at the moment it takes effect.

(1) plus (2) is the honest fix; (3) alone leaves a footgun with a warning label.

Also worth reviewing: `eventMaxRows` is a hard-coded constant with no way to tune it even when the sweep *is* running.

## Context

Found while mapping telemetry retention in response to a question about unbounded growth. Unrelated to the trace pipeline — filed separately from the telemetry-aggregate issue for that reason.

## Acceptance

- Disabling the agent/schema sweep does not change whether `registry_events` is capped.
- Any setting that removes a bound says so where the operator will see it.

dhyansraj:
## Summary

Two registry-side bounds that silently disappear. The trace *data* pipeline was already bounded; the **in-memory aggregate layer** on top of it was not, and no knob reached it. Separately, disabling the sweep also removed an unrelated table cap.

## Design: age-based prune, hard cap as backstop

Not one or the other. The cardinality here is *names*, so "has this key stopped reporting?" is the semantically correct question — an agent that stopped should age out, one still reporting should never be dropped.

A pure LRU cap gets that backwards under a churn burst: it evicts by recency-of-write regardless of liveness, so a live-but-low-traffic agent can vanish from the dashboard while a burst of ephemeral pod names fills the map. Age alone has only a soft ceiling (new-name rate × retention), so a hard cap remains — but it engages **only after age pruning has failed to keep up**, and evicts least-recently-seen first, degrading *into* LRU rather than starting there. Every eviction batch logs both knob names.

Both are knobs: `MCP_MESH_TELEMETRY_AGGREGATE_RETENTION` (24h) and `MCP_MESH_TELEMETRY_AGGREGATE_MAX_ENTRIES` (10000). `0` disables either and is logged as a removed bound. **A negative or unparseable value falls back to the default, never to 0** — a typo must not silently unbound the map, which is the defect class this PR is fixing.

Eviction is batched to ~90% of cap so the sort amortises to `O(log n)` per insert. `MetricsProcessor` prunes off the write path on a 10s gate rather than owning a goroutine — it has no lifecycle today, and `replayWindow` constructs throwaway instances that would otherwise need Start/Stop wiring. The accumulator reuses its existing 10s `cleanupLoop`.

**No knobs on the Redis housekeeping thresholds** (`pelMinIdle` 5m, `pelMaxDeliveries` 5, `consumerIdleTimeout` 1h). Those are timing thresholds for a loop that runs unconditionally, not memory/retention trade-offs — the bound is never off, so there is nothing to trade.

## The trace-stream single point of failure — both halves fixed

Both were small:

- **meshui now trims.** `manager_ui.go` passes `Retention` through, `cmd/mcp-mesh-ui/main.go` sets it from the shared `tracing.TraceRetentionFromEnv()`. Helm wires `ui.tracing.retention` next to the registry's, with a comment in both charts that they must agree.
- **The producer now caps.** `tracing_publish.rs:372` uses `StreamMaxlen::Approx(n)` — one call site covering Python, TypeScript, Java and Go.

Previously the `XADD` was uncapped and only the registry trimmed, so with the registry down the only backstop was `allkeys-lru`, which evicts the **entire** `mesh:trace` key rather than shedding old entries.

## Validation

**Each bound proven to bound** — driven past the limit, asserted after *every* write, not just at the end:

| test | result |
|---|---|
| `TestEdgeStatsUnboundedWithoutBounds` — 2000 churned names, bounds off | **2000** keys (documents pre-fix growth) |
| `TestEdgeStatsKeyCapBoundsGrowth` — same, cap 50 | never exceeds 50 |
| `TestMetricsProcessorUnboundedWithoutBounds` | agents **2000**, models **2000** |
| `TestMetricsProcessorKeyCapBoundsGrowth` — cap 50 | never exceeds 50 |
| `TestEnforceMaxEntriesBoundsGrowth` — 5000 inserts, cap 100 | never exceeds 100; newest survives, oldest gone (proves LRU, not arbitrary) |
| age-prune tests | idle key removed **and** still-active key survives with counters intact |

**#1425 before/after, run against the actual pre-fix code** in a worktree at `b83e1b9ee`, same assertion (`MCP_MESH_RETENTION=0`, cap 3, 5 events):

```
PRE-FIX:  registry_events count -> 5    FAIL
POST-FIX: registry_events count -> 3    PASS
```

`TestEventCapOnlyTickLeavesStaleAgentsAlone` confirms the other direction — a 30-day-stale agent survives, so the cap-only tick really is cap-only. `TestEnforceEventCapDisabledDoesNotDeleteEverything` guards the obvious way to get this wrong (cap 0 ⇒ every row exceeds it ⇒ delete the table).

**Redis, against a real server** (skipped without `MCP_MESH_TEST_REDIS_URL`, so `go test ./...` needs no service): a stuck PEL entry is proven stuck first — a second `XREADGROUP` does *not* redeliver it — then reclaimed; a poison pill is dropped after `pelMaxDeliveries`; 3 dead consumer incarnations reaped down to the 1 live one; and the reap provably discards no PEL entry.

Producer `MAXLEN` on the same Redis: uncapped 20000 XADDs → `XLEN 20000`; `MAXLEN ~ 1000` → `XLEN 1000`. Asserted on the wire by `publish_span_caps_stream_with_maxlen`, which captures raw RESP and checks the `~` flag.

**Tallies** (same machine, `-count=1`): `go test ./...` 778 → **811 PASS, 0 FAIL, 0 races**; `cargo test` 473 → **475**. Same 4 pre-existing skips both sides. `pytest scripts/` 85; `helm lint` 9/9; PSS OK; render matrix 34/34.

**No default-path change**: a normally-sized mesh (4 agents / 3 models / 500–1000 traces) run through bounded and unbounded processors produces byte-identical rows and totals.

## Worth reviewer attention

- **One pre-existing test was intentionally changed.** `TestSweepDisabledStartIsNoop` asserted `running == false` at `Retention=0` — precisely the contract #1425 inverts. It now covers the only remaining no-op path (both bounds off), with `TestSweepRetentionZeroStillRunsForEventCap` as its counterpart.
- **`EventCapDisabled = -1` sentinel.** 47 existing test literals construct `SweepConfig{Retention: ...}` leaving `EventMaxRows` zero-valued; making `0` mean "disabled" would let a forgotten struct field silently remove the bound — the same defect class as the issue. So `0 = unset → default`, `-1 = deliberately off`, and env `"0"` maps to the sentinel with a warning.
- **`replayWindow` throwaway instances inherit the same bounds**, so a 1h/1d traffic window with more distinct keys than `MaxEntries` will evict. Intended and consistent, but a behaviour change at that (pathological) scale.
- **Pre-existing formatting left alone**: `sweep.go` has a `gofmt` struct-alignment violation and `tracing_publish.rs` has 6 `rustfmt` hunks, both predating this branch. Verified zero new violations added (rustfmt hunks 6 → 6); unrelated lines not reformatted.
- **Noted, deliberately out of scope**: `MCP_MESH_RETENTION=0` also disables the orphan-job reroute, expired-lease reclaim and total-deadline expiry, which sit unguarded inside `runOnce`. Arguably the same bug class as #1425, but fixing it changes job-lifecycle behaviour well beyond what either issue asked for. Happy to file separately.

Closes #1424
Closes #1425

## Summary by CodeRabbit

* **New Features**
  * Added configurable limits for trace stream size, event-table rows, and dashboard telemetry aggregates.
  * UI tracing now continues trimming trace data during registry outages.
  * Added automatic recovery of stalled tracing messages and cleanup of inactive consumers.
  * Registry event limits can now be configured independently from general retention settings.

* **Documentation**
  * Expanded observability, environment-variable, and Helm configuration guidance, including defaults and disable options.

* **Bug Fixes**
  * Improved protection against unbounded telemetry growth and stuck trace-processing entries.

### DIFF
--- a/cmd/mcp-mesh-ui/main.go
+++ b/cmd/mcp-mesh-ui/main.go
@@ -167,7 +167,10 @@ func main() {
 			BatchSize:     100,
 			BlockTimeout:  2 * time.Second,
 			TraceTimeout:  5 * time.Minute,
-			TempoQueryURL: resolveTempoURL(tempoURL),
+			// meshui honours MCP_MESH_TRACE_RETENTION too, so stream trimming
+			// does not depend on the registry process being alive (#1424).
+			TraceRetention: tracing.TraceRetentionFromEnv(),
+			TempoQueryURL:  resolveTempoURL(tempoURL),
 		}
 
 		metricsProc = ui.NewMetricsProcessor()
--- a/docs/07-observability.md
+++ b/docs/07-observability.md
@@ -60,7 +60,10 @@ export TEMPO_URL=http://localhost:3200
 | `TELEMETRY_PROTOCOL` | `grpc` | OTLP protocol: `grpc` or `http` |
 | `TRACE_EXPORTER_TYPE` | `otlp` | Exporter: `otlp`, `console`, or `json` |
 | `TEMPO_URL` | `http://localhost:3200` | Tempo query URL for the trace API |
-| `MCP_MESH_TRACE_RETENTION` | `24h` | Redis `mesh:trace` stream retention (`0` disables trimming) |
+| `MCP_MESH_TRACE_RETENTION` | `24h` | Redis `mesh:trace` stream retention (`0` disables trimming). Honoured by both the registry and meshui, so trimming survives either being down |
+| `MCP_MESH_TRACE_STREAM_MAXLEN` | `100000` | Producer-side `XADD MAXLEN ~` ceiling on `mesh:trace`, applied by every agent runtime. Bounds the stream even with no consumer alive (`0` disables) |
+| `MCP_MESH_TELEMETRY_AGGREGATE_RETENTION` | `24h` | Age-out window for the in-memory per-agent / per-model / per-edge dashboard aggregates (`0` disables age pruning) |
+| `MCP_MESH_TELEMETRY_AGGREGATE_MAX_ENTRIES` | `10000` | Hard key ceiling per aggregate map, evicting least-recently-seen first (`0` disables the ceiling) |
 
 See the [environment variables reference](environment-variables.md) for the full list.
 
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -273,14 +273,23 @@ export DEFAULT_EVICTION_THRESHOLD=60
 ```bash
 # How long unhealthy/unknown agents are kept in the registry before the
 # sweep job purges them. Go duration string (e.g. "30m", "2h", "48h").
-# Default: 1h. Set to "0" to disable the sweep entirely (forensic mode —
-# keeps all rows). Affects `meshctl list` (purged agents disappear from
-# output), the underlying RegistryEvent table (event rows are governed
-# by a separate hardcoded 100,000 rolling cap, not by this variable),
-# and orphan schema_entries (rows in the content-addressed schema store
-# that are no longer referenced by any capability — purged when both
-# orphan and older than this retention window; #842).
+# Default: 1h. Set to "0" to disable the agent/schema sweep (forensic
+# mode — keeps all agent and schema rows). Affects `meshctl list` (purged
+# agents disappear from output) and orphan schema_entries (rows in the
+# content-addressed schema store that are no longer referenced by any
+# capability — purged when both orphan and older than this retention
+# window; #842). It does NOT govern the registry_events table: that has
+# its own row cap, MCP_MESH_EVENT_MAX_ROWS, which stays enforced even
+# with the sweep disabled.
 export MCP_MESH_RETENTION=1h
+
+# registry_events row cap — a table-size safety limit, not a retention
+# policy. When the table exceeds this many rows the oldest are deleted
+# until it is back under the cap. Enforced independently of
+# MCP_MESH_RETENTION, so disabling the sweep does not unbound this table.
+# Default: 100000. Set to "0" to disable the cap deliberately (logged as
+# a removed bound at startup).
+export MCP_MESH_EVENT_MAX_ROWS=100000
 
 # How often the periodic job/agent sweep runs (reaping, lease recovery,
 # retention purges). Go duration string (e.g. "30s", "5m", "1m30s").
@@ -310,7 +319,9 @@ export MCP_MESH_JOB_STALE_TIMEOUT=2h
   `1m` with the default 5m sweep) will not speed up the purge cadence —
   it only affects when an agent becomes eligible for purge. Lower
   `MCP_MESH_SWEEP_INTERVAL` to tighten the cadence.
-- Internal constant: event hard cap = 100,000 rows.
+- `MCP_MESH_RETENTION=0` disables only the agent/schema/job phases. The
+  `registry_events` row cap keeps running on the same schedule; use
+  `MCP_MESH_EVENT_MAX_ROWS=0` to turn that off too.
 
 ### Logging and Debug
 
@@ -456,10 +467,25 @@ export TRACE_EXPORTER_TYPE=otlp
 export TRACE_BATCH_SIZE=100
 export TRACE_TIMEOUT=5m
 
-# Redis trace stream retention — registry trims mesh:trace entries older
-# than this on connect and periodically. Go duration string. Default: 24h.
+# Redis trace stream retention — the registry AND meshui both trim mesh:trace
+# entries older than this on connect and periodically, so trimming does not
+# depend on a single process being alive. Go duration string. Default: 24h.
 # Set to "0" to disable trimming entirely.
 export MCP_MESH_TRACE_RETENTION=24h
+
+# Producer-side ceiling on the mesh:trace stream. Every agent runtime publishes
+# with XADD MAXLEN ~ <n>, so the stream stays bounded even when no consumer is
+# running. This is a safety ceiling, not a retention policy — time-based
+# retention stays with MCP_MESH_TRACE_RETENTION. Default: 100000. "0" disables.
+export MCP_MESH_TRACE_STREAM_MAXLEN=100000
+
+# In-memory telemetry aggregates (dashboard per-agent, per-model and per-edge
+# stats). Keyed by name, so they grow with agent/model name cardinality rather
+# than trace volume. A key not written to within the retention window ages out;
+# the max-entries ceiling is a backstop for name churn, evicting the
+# least-recently-seen key first. "0" disables the respective bound.
+export MCP_MESH_TELEMETRY_AGGREGATE_RETENTION=24h
+export MCP_MESH_TELEMETRY_AGGREGATE_MAX_ENTRIES=10000
 
 # Trace output options
 export TRACE_PRETTY_OUTPUT=false
--- a/helm/mcp-mesh-core/values.yaml
+++ b/helm/mcp-mesh-core/values.yaml
@@ -185,7 +185,9 @@ mcp-mesh-registry:
         enabled: true
         # mesh:trace Redis stream retention (XTRIM MINID on connect and
         # periodically); "0" disables trimming. The stream is a transport
-        # buffer — long-term trace history lives in Tempo.
+        # buffer — long-term trace history lives in Tempo. meshui trims to the
+        # same window (see mcp-mesh-ui.ui.tracing.retention below), so trimming
+        # does not depend on the registry being alive.
         retention: "24h"
       telemetryEndpoint: "mcp-core-mcp-mesh-tempo:4317"
 
@@ -323,6 +325,12 @@ mcp-mesh-tempo:
 mcp-mesh-ui:
   ui:
     registryURL: "http://mcp-core-mcp-mesh-registry:8000"
+    tracing:
+      enabled: true
+      # Keep equal to mcp-mesh-registry.registry.observability.distributedTracing.retention
+      # above: both processes trim mesh:trace, so trimming survives either
+      # being down, but only if they agree on the window.
+      retention: "24h"
     basePath: "/ops/dashboard"  # default; override with custom image for different path
     tempo:
       url: "http://mcp-core-mcp-mesh-tempo:3200"
--- a/helm/mcp-mesh-registry/values.yaml
+++ b/helm/mcp-mesh-registry/values.yaml
@@ -271,10 +271,11 @@ registry:
   observability:
     distributedTracing:
       enabled: true
-      # How long span events stay in the mesh:trace Redis stream before the
-      # registry trims them (XTRIM MINID on connect and every 5m). The stream
-      # is a transport buffer — long-term trace history lives in Tempo. Go
-      # duration string; "0" disables trimming entirely.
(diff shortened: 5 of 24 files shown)


dhyansraj:
Verified five findings against current code in `f2693f6ee`: **three fixed, two skipped.** Each fix was proven red against the pre-amend commit before being trusted.

## Fixed

**1. A stray negative silently disabled the event cap — in this PR's own code.**

`NewSweepJob` had `if maxRows < 0 { maxRows = 0 }`, so `-2`, a sign typo, or an arithmetic slip unbounded `registry_events` silently. That is exactly the defect class #1425 is about, and the `EventCapDisabled = -1` sentinel was introduced *specifically* so a forgotten field could not disable the cap — then the guard accepted every negative anyway.

Now an explicit switch: only the sentinel disables, `0` still means unset → default, every other negative **warns and falls back to the default**. Proven red first — against `da314e688` with only the new test file added, `non_sentinel_negative_falls_back`, `large_negative_falls_back` and `min_int_negative_falls_back` all reported `eventMaxRows = 0, want 100000`.

**2. Per-message ACK context in the reclaim loop.**

The pass context is `WithTimeout(sc.ctx, 10s)`, shared across up to `pelReclaimBatch = 100` messages with a `processMessage` between each. A slow pass exhausts it, so later `XAck`s fail — the message is processed but never acknowledged, stays in the PEL, and is reclaimed again next pass. Duplicate work indefinitely.

Successful-processing `XAck` now gets its own 5s context, cancelled immediately after. `XPendingExt`, `XClaim` and the poison-pill `XAck` keep the pass context — none is downstream of `processMessage`. Still derived from `sc.ctx`, so shutdown cancels it.

Proven red with the fix reverted but the test seam kept:
```
WARN: failed to acknowledge reclaimed message 1785441190881-1: context deadline exceeded
WARN: failed to acknowledge reclaimed message 1785441190881-2: context deadline exceeded
reclaimed=1 dropped=0, want 3 and 0
```

**3. Trace-stream maxlen cached out of the publish hot path.**

`trace_stream_maxlen()` ran `env::var` + parse on **every span**, and on a misconfigured value emitted `warn!` per span — a log flood, not just overhead. Now resolved once in `init_trace_publisher`, stored on `TracePublisherState`, read from the existing read-lock block.

A state field rather than a `OnceLock`, deliberately: a process-wide `OnceLock` would freeze whichever test initialises first and break `publish_span_caps_stream_with_maxlen`'s 777 assertion.

Proven **on the wire** rather than by counting log lines: init with `321`, corrupt the env to `not-a-number`, publish 3 spans. With the per-span read reverted, captured RESP showed `XADD mesh:trace MAXLEN ~ 100000` — re-parsing, warning and falling back per span. With the fix, `321` every time. The same test re-inits with `0` and asserts no `MAXLEN` on the wire at all.

## Skipped

**Gating the reaper on Redis 7.2+ — the premise does not hold.** `reapIdleConsumersWithTimeout` reads `c.Idle`, which `XINFO CONSUMERS` has returned since Redis **5.0**; `Inactive` (7.2) is never read. go-redis v9.11.0's parser errors only on an *unexpected* field (`default: return fmt.Errorf(...)`) — an older Redis returns a subset, not extras, so it parses cleanly. `XPendingExt`, `XClaim` and `XGROUP DELCONSUMER` are all 5.0+. Nothing in that function was touched, and no contrary evidence turned up while working.

**Injectable logger for `MetricsProcessor`.** `log.Printf` is the only logging in that package and was introduced by this PR for three call sites; `go test` hides output for passing tests, and eviction is batched, so this is test-tidiness with no correctness gain.

## Validation

`go test ./...` and `-race`: 10 packages ok, 0 FAIL, both pre- and post-amend. `cargo test --no-default-features`: 475 → **476** (+1 new). Redis-gated suite against a throwaway server on **:6399**: all 3 packages ok, 37 PASS in `tracing`, 0 FAIL/SKIP; `test_trace_stream_maxlen_env` and `publish_span_caps_stream_with_maxlen` pass unmodified.

Default path unchanged in all three: sweep `0`→default / positive→itself / `-1`→disabled are bit-identical and only non-sentinel negatives moved; the ACK succeeds identically when a pass is not slow; Rust keeps the same 100000 approximate cap and the `maxlen == 0` plain-`xadd` branch.

`git diff --name-only da314e688 HEAD` → exactly the five intended files.

Issue #1401, opened 2026-07-29
Java DI: four matching rules across five injection sites — align @MeshRoute and @MeshA2A on the positional contract (3.4.0, breaking)


dhyansraj:
## Summary

Java has **four different dependency→parameter matching rules across five injection sites**, and two of them contradict each other inside the same annotation family. `@MeshTool` pairs declared dependencies with parameters **positionally**; `@MeshRoute` and `@MeshA2A` match **by name**. Python is positional at every site.

This issue proposes aligning Java on the positional contract, and records two things that must be settled first: a missing validation layer in Java, and an undecided divergence in TypeScript.

**Breaking change** — targeted at 3.4.0.

## Current state

| runtime | `tool` | `route` | `a2a` |
|---|---|---|---|
| Python | positional | positional | positional |
| TypeScript | positional | **by capability key** | positional |
| **Java** | **positional** | **by name** | **by name** |

### Java `@MeshTool` — positional

`src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshToolWrapper.java:396-419` builds the pairing table: `k` indexes the declared `dependencies = {...}` list, `eligiblePositions` holds injectable parameter positions in signature order. Nth declared dependency → Nth injectable parameter. Parameter names are never consulted.

The field javadoc at `:128-137` states the contract and explicitly claims parity: *"the same contract as Python/TypeScript."*

### Java `@MeshRoute` — by name

`.../spring/web/MeshInjectArgumentResolver.java:63-107` derives a capability string from `@MeshInject`'s value, falling back to the parameter name, then does a map lookup:

```java
capability = annotation.value();            // :69-72
if (capability == null) {
    capability = parameter.getParameterName();   // :77, requires -parameters
}
McpMeshTool tool = dependencies.get(capability);  // :101
```

The map is double-keyed by capability and by derived parameter name (`MeshRouteHandlerInterceptor.java:155-157`). Declaration order of `dependencies = {...}` is irrelevant.

### Java `@MeshA2A` — by name, third code path

`.../spring/web/MeshA2ADispatcher.java:765-788` — same semantics as `@MeshRoute` but a hand-rolled linear scan rather than the shared resolver.

### Other Java sites, for completeness

[...]

Issue #547, opened 2026-02-11
Schema Registry — type-safe capability matching with schema compatibility


dhyansraj:
## Summary

Add schema registry support to MCP Mesh so the registry can validate producer output schemas against consumer expected schemas at wiring time, catching mismatches early instead of at runtime deserialization.

## Motivation

Currently, when a consumer depends on a capability, the registry matches on capability name, tags, and version. There's no contract on what a tool returns — a consumer just hopes the response matches what it expects. Schema mismatches surface as runtime deserialization errors (e.g., the `isError` guard added in #544).

A schema registry catches these at wiring time: "this consumer expects `{name, department}` but the producer returns `{id, fullName}` — incompatible, don't wire."

### The cross-runtime parity problem

Pydantic, Jackson, and `zod-to-json-schema` produce *different* JSON Schema for semantically identical types — enum representations, date formats, optionality (`null` vs absent), discriminated unions, generics. A naive design has each runtime emit its own schema and asks the registry to reconcile across vendors. That doesn't scale; the divergence space is unbounded.

**Solution: Rust as the single normalization point.** All three runtimes already cross the Rust FFI boundary at registration. Each runtime extracts schema using its native generator, hands the raw output to a Rust normalizer over FFI, and the normalizer emits a canonical JSON Schema. Producer and consumer can be in different languages and still reconcile, because both went through the same normalizer.

```
┌─────────────────┐
│ Python runtime  │ Pydantic → raw schema ──┐
└─────────────────┘                         │
┌─────────────────┐                         ▼
│  TS runtime     │ Zod → raw schema ──┬─→ Rust Normalizer ─→ canonical schema ─→ Go Registry
└─────────────────┘                    │
┌─────────────────┐                    │
│ Java runtime    │ Jackson → raw ─────┘
└─────────────────┘
```

### Per-runtime responsibilities (before crossing FFI)

Runtimes strip language-specific wrappers that aren't part of the contract:
- Python: `Awaitable[T]`, `AsyncIterator[T]` → emit `T`
- TS: `Promise<T>` → emit `T`
- Java: `Mono<T>`, `CompletableFuture<T>` → emit `T`

The Rust normalizer never sees these wrappers; it only sees the inner type's schema.

### Three-tier validation policy

The normalizer reports a verdict per schema:

- **OK** — schema normalized cleanly. Agent starts; canonical schema registered.
[...]

Issue #936, opened 2026-05-11
feat(a2a): auto-inject MeshJobSubmitter for @MeshA2A / mesh.a2a.mount task=true deps


dhyansraj:
## Context

Surfaced during PR #934 (Java A2A producer) and PR #935 (TypeScript A2A producer). Both runtimes have a framework gap that affects every user who wants to write a long-running A2A producer.

## The gap

`@MeshA2A` (Java) and `mesh.a2a.mount(...)` (TypeScript) handle DDDI for declared dependencies the same way `@MeshRoute` / `mesh.route()` do — they inject `McpMeshTool` proxies into handler parameter slots. But they **do NOT auto-inject `MeshJobSubmitter`**, which is the primitive long-running producers need to return a `JobProxy` from their handler.

`@MeshTool` handlers (inside mesh agents) DO get `MeshJobSubmitter` auto-injection via `JobsRuntimeManager.wireConsumers`. The user-owned-app producer path is missing this.

### Java (example: `examples/java/producer-report-agent/`)
```java
@Component
public class ReportSkill {
    @Lazy @Autowired private MeshRuntime meshRuntime;  // @Lazy needed to break boot cycle

    @MeshA2A(...)
    public JobProxy generateReport(...) {
        MeshJobSubmitter submitter = new MeshJobSubmitter(
            "generate_report",
            meshRuntime.getAgentSpec().getAgentId(),
            meshRuntime.getAgentSpec().getRegistryUrl()
        );
        return submitter.submit(payload).get(30, TimeUnit.SECONDS);
    }
}
```

The `@Lazy` is itself a workaround for the broader Java cycle pattern flagged separately. Users copy-pasting this example need to understand BOTH workarounds.

### TypeScript (example: `examples/typescript/producer-report-agent/`)
```typescript
const submitter = new MeshJobSubmitter(
  "generate_report",
  getApiRuntime().getServiceId(),
  process.env.MCP_MESH_REGISTRY_URL ?? "..."
);
```

Hand-constructed via internal `getApiRuntime()` API. Same shape, different language idiom.

## What "fixed" looks like

The handler should be able to declare `MeshJobSubmitter` as a parameter and get it injected by the framework — just like `McpMeshTool` proxies for declared dependencies today.

### Java
```java
@MeshA2A(skillId = "generate-report", ...)
public JobProxy generateReport(
    Map<String, Object> payload,
    MeshJobSubmitter jobSubmitter,    // ← auto-injected
[...]

PR #1436, opened 2026-07-30
feat(java)!: bind @MeshRoute and @MeshA2A dependencies by position

dhyansraj:
## Summary

**PR B of the #1401 sequence — the breaking Java conversion.** `@MeshRoute` and `@MeshA2A` dependencies now bind **by position**, matching Python and Java's own `@MeshTool`. Parameter names are never consulted.

Four sites convert together. Three had to, or the fix would have created new silent failures.

## The four sites

1. **`MeshInjectArgumentResolver`** — the parameter's injectable-slot ordinal (from `MeshInjectableSlots.routeSlots`, cached per `Method`) indexes the positional list. No name read at request time.
2. **`MeshRouteHandlerInterceptor`** — `MESH_DEPENDENCIES_ATTR` is now `List<McpMeshTool>`.
3. **`MeshRouteBeanPostProcessor.enrichDependencyReturnTypes`** — the `McpMeshTool<T>` generic follows the slot ordinal.
4. **`MeshA2ADispatcher`** — `invokeHandler` consults a cached `ArgPlan` from `planArguments`.

## Why site 3 had to be in this PR — demonstrated, not argued

The return-type enrichment feeds `injector.getToolProxy(cap, returnType)`, driving response deserialization and the #547 schema payload. Converting the resolver while leaving it by-name would make a reordered handler deserialize into the **wrong type**.

The deliberate-break run shows exactly that flip:

```
before = {employee_count=DepartmentStats, get_employee=Employee}
after  = {get_employee=DepartmentStats, employee_count=Employee}
```

A new silent failure, created by the fix, now closed by it.

## The three traps

**(1) Null-padding — avoided by construction.** `MESH_DEPENDENCIES_ATTR` is pre-sized with `Collections.nCopies(declared.size(), null)` and written with `set(i, …)`; the loop is index-driven and there is **no `add()` anywhere in the file**. The old map simply omitted unavailable dependencies, which is harmless for a map and shifts every later slot in a list — precisely the defect #1390 pinned.

**(2) `MeshRouteUtils` kept working.** `getDependencies` is rebuilt from `MESH_ROUTE_METADATA_ATTR` plus the positional list on each call, so there is one source of truth. Capability keys first, `@MeshDependency(name=…)` aliases via `putIfAbsent` so an alias can never shadow a capability. Added `getDependency(request, int)` — the attribute is positional now and had no public positional accessor.

**(3) A2A precedence is explicit.** `planArguments` assigns roles in four ordered passes — dependency slots → `MeshJobSubmitter` → message → null — rather than relying on if/else ordering, where the tool branch previously won a collision only because it was tested first.

## `@MeshInject` is now a checked assertion

Its value must equal what position assigns, or boot fails — **unconditionally, not gated on `MCP_MESH_STRICT_DI`**. Correctly-ordered code behaves identically; a redundant annotation becomes a safety net. It is now also honoured on `@MeshTool` parameters, where it was silently ignored.

```
BOOT FAILED: com.example.api.ApiController
@MeshRoute ...getEmployee(int, McpMeshTool, McpMeshTool): a @MeshInject value
contradicts the dependency its parameter's POSITION binds to.
  2 declared dependencies: [0] 'get_employee', [1] 'employee_count'
    slot 0 = parameter 1 (McpMeshTool statsTool)
        @MeshInject("employee_count") asserts:  dependency[1] 'employee_count'
        binds (by position):                    dependency[0] 'get_employee'
  Fix — pick one:
    • reorder dependencies = {...} to: [0] 'employee_count', [1] 'get_employee'
```

## No in-tree app changes behaviour — verified independently of PR A

PR A's detector predicted this: 0 findings across 102 modules. Confirmed here by a different method — all 79 example/fixture modules compiled against the modified SDK (`ok=79 fail=0`), then a throwaway scanner carrying a **verbatim reimplementation of the three deleted name-matching paths** run against the real shipped components:

```
handlers=21 (route=14 a2a=7)  slots=14  SLOT DIFFERENCES=0  RETURN-TYPE DIFFERENCES=0
shipped guard verdicts: OK=21 WARN=0 ERROR=0
```

14/14 route and 7/7 a2a matches PR A's census exactly. Two independent methods, same answer.

## Validation

- **Null-padding**, asserted on the interceptor's stored list: an unavailable middle dependency leaves index 2 holding `c`. Also the throwing-resolution and `isAvailable()==false` variants, and A2A equivalents driven through real `tasks/send`.
- **`MeshRouteUtils`** — 5 tests across `getDependencies`, `getDependency(String)`, `requireDependency` (including the throw), `hasDependency`, the alias key, the index accessor and the no-attribute path.
- **Return-type enrichment** — `TypedController` declares identical `(McpMeshTool<Alpha>, McpMeshTool<Beta>)` parameters under two declaration orders; the reordered case asserts `beta→Alpha, alpha→Beta`, with a separate test proving `getToolProxy("beta", Alpha.class)` reaches the injector.
- **Suite**: 1080 → **1112** (+32). `MeshPositionalBinderTest` 12/12 and `MeshToolWrapperUnresolvedDepPositionTest` 5/5 pass **unedited**. `go build ./...` clean.

*Baseline note:* PR A reported starter 739 / total 1082; this environment measures 737 / 1080 at that same SHA. A 2-test environmental gap, not a regression — both sides of the comparison above were measured here, and every touched class is accounted for.

## Judgement calls for review

- **Parameter-name crossover stays WARN (strict-promotable), only `@MeshInject` is fatal.** The brief said "detector goes WARN → ERROR", but that paragraph was about `@MeshInject`. Making a *name* coincidence permanently fatal — under a contract that says names are never consulted — would be unfixable except by renaming a parameter. One-line change if you want it fatal in 3.4 as a forcing function.
- **`@MeshInject` accepts the `@MeshDependency(name=…)` alias** as well as the capability, at the paired index only, so it can never point at a different dependency. Without it, `capability="base-cap"` + `@MeshInject("baseCap")` would fail boot on a spelling the framework itself generates.
- **Class name `MeshLegacyBindingDetector` kept** though it now enforces rather than forecasts — so the PR A ↔ PR B diff reads as a content change rather than delete+add. Rename is cheap.

## Out of scope

TypeScript (PR C), docs/man/release notes (PR D), and the `jobBeforeProxy` discriminator from PR A — no guard added, still awaiting a decision.

Refs #1401

## Summary by CodeRabbit

- **New Features**
  - Mesh dependencies are now injected consistently by parameter position across routes, A2A handlers, and tools.
  - `@MeshInject` validates the expected dependency position during startup, with clear errors for contradictions.
  - Dependency handling preserves parameter slots, including unavailable dependencies.

- **Bug Fixes**
  - Prevented silent mismatches caused by dependency names or parameter names.

- **Documentation**
  - Updated guidance and examples to explain positional dependency binding and startup validation.

### DIFF
--- a/examples/java/producer-date-agent/src/main/java/com/example/dateproducer/ProducerDateAgentApplication.java
+++ b/examples/java/producer-date-agent/src/main/java/com/example/dateproducer/ProducerDateAgentApplication.java
@@ -39,8 +39,11 @@ import java.util.Map;
  *
  * <p>The mesh dependency on {@code date_service} demonstrates DDDI
  * inside an A2A handler: at request time the framework resolves the
- * {@link McpMeshTool} proxy through {@link MeshDependency} and injects
- * it at the {@link MeshInject} parameter slot — same wiring used by
+ * {@link McpMeshTool} proxy declared by {@link MeshDependency} and injects
+ * it into the first injectable parameter — dependencies pair with
+ * parameters by POSITION, in signature order. The {@link MeshInject}
+ * annotation does not select the dependency; it asserts the one position
+ * assigns, and startup fails if the two disagree. Same wiring used by
  * {@code @MeshRoute}.
  *
  * <h2>Stack</h2>
--- a/examples/java/rest-api-consumer/src/main/java/com/example/api/RestApiConsumerApplication.java
+++ b/examples/java/rest-api-consumer/src/main/java/com/example/api/RestApiConsumerApplication.java
@@ -84,8 +84,9 @@ class ApiController {
      * Greet a user via a mesh agent.
      *
      * <p>The @MeshRoute annotation declares a dependency on the "greeting" capability.
-     * At request time, the framework resolves the dependency and injects it as a
-     * McpMeshTool parameter (matched by parameter name).
+     * At request time, the framework resolves it and injects it into the first
+     * McpMeshTool parameter: dependencies pair with McpMeshTool parameters by
+     * POSITION, in signature order. The parameter's name is irrelevant.
      */
     @GetMapping("/greet")
     @MeshRoute(dependencies = @MeshDependency(capability = "greeting"))
--- a/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshToolWrapper.java
+++ b/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshToolWrapper.java
@@ -418,6 +418,14 @@ public class MeshToolWrapper implements McpToolHandler {
         int minSlots = (task && meshJobParamIndex != null) ? slotCount - 1 : slotCount;
         MeshDiValidator.checkArity("@MeshTool", binding, minSlots, slotCount);
 
+        // Issue #1401: @MeshInject is a checked assertion at EVERY injection
+        // site, not just the web ones. It was silently ignored on @MeshTool
+        // parameters — the wrapper never even imported it — which meant the
+        // annotation quietly meant two different things depending on where it
+        // was written. Here it asserts the dependency positional pairing
+        // assigns; a value that disagrees fails the boot. It never selects.
+        io.mcpmesh.spring.web.MeshLegacyBindingDetector.inspectTool(binding);
+
         // Generate input schema (excluding injected params)
         this.inputSchema = generateInputSchema();
 
--- a/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/web/MeshA2ABeanPostProcessor.java
+++ b/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/web/MeshA2ABeanPostProcessor.java
@@ -70,9 +70,9 @@ public class MeshA2ABeanPostProcessor implements BeanPostProcessor {
                 deps.add(MeshRouteRegistry.DependencySpec.fromAnnotation(dep));
             }
 
-            // Issue #1401: report handlers whose bindings would change when
-            // @MeshA2A moves from name-based to positional pairing. Warning
-            // only — this PR changes no binding behaviour.
+            // Issue #1401: @MeshA2A binds positionally. Fail the boot on a
+            // @MeshInject value that contradicts the position, and warn on a
+            // handler still shaped for the pre-3.4 name-based binding.
             MeshLegacyBindingDetector.inspectA2A(method, deps);
 
             String handlerMethodId = targetClass.getName() + "." + method.getName();
--- a/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/web/MeshA2ADispatcher.java
+++ b/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/web/MeshA2ADispatcher.java
@@ -3,6 +3,7 @@ package io.mcpmesh.spring.web;
 import io.mcpmesh.JobProxy;
 import io.mcpmesh.MeshJobSubmitter;
 import io.mcpmesh.spring.MeshDependencyInjector;
+import io.mcpmesh.spring.MeshPositionalBinder;
 import io.mcpmesh.spring.MeshRuntime;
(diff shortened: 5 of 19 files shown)


dhyansraj:
Review finding verified and fixed in `f142df510` — **and the audit it prompted found two more instances, one of which is on the route path.**

## The reported defect was real and reachable

`argPlans` was keyed by `surface.handlerMethodId()`, which is `ClassName.methodName` with **no parameter types** (`MeshRouteBeanPostProcessor:82`). Two overloaded `@MeshA2A` handlers therefore share a key, and `MeshA2ARegistry:65` dedupes by `path()` rather than handler id — so both register as distinct surfaces sharing one cache entry.

Proven pre-fix, three overload pairs:

```
differingArityDoesNotReuseTheNarrowPlan: Index 2 out of bounds for length 2
equalArityDoesNotMisbindTheDependency:   expected <null> but was <alpha-proxy>
sameRoleLayoutStillUsesItsOwnCapabilities: expected <beta-proxy> but was <alpha-proxy>
```

The third is the one that matters — an overload with an **identical role layout** receiving the other surface's declared capability, because `binding().capabilities()` is baked into the leaked plan. It cannot throw. It just injects the wrong tool.

Also worth noting: the route path was already keyed correctly (`MeshInjectArgumentResolver:71`, `Map<Method, int[]>`), so the two halves of this PR disagreed about how to identify a handler.

## `Method` is a safe key — argued, not assumed

`@MeshA2A` is not `@Repeatable`, and `MeshA2ABeanPostProcessor` derives the whole `SurfaceMetadata` from that single annotation — so any two registrations of one `Method` are metadata-identical by construction, which means identical `path`, which `putIfAbsent` rejects at boot. That covers both ways a `Method` can be seen twice: two bean instances of one `@Component`, and a subclass override inheriting via `findAnnotation`.

## Two siblings folded in

**`jobSubmitterCache`** had the identical defect, and it is the worse of the two — `resolveJobSubmitter` picks its capability from `deps.get(0)` or `skillId`, so the second overload submitted every job under the first's capability. Proven on the submitter itself, not on cache identity:

```
jobSubmitterIsNotSharedAcrossOverloads: expected <job_second> but was <job_first>
```

**`unbuildableSurfaces`** — re-keyed for key-space consistency, but **no test was written for it**, deliberately. The differential case is not reachable today: one writer (`runtimeProvider == null`) is dispatcher-wide so there is no sibling to wrong, and the other is dead because `pickSubmitterCapability` falls back to `skillId.replace('-','_')` and the compact constructor rejects an empty `skillId`. Writing a test that passes identically before and after would have been evidence of nothing. The reasoning is in the field's javadoc instead.

## The audit turned up a third instance — filed as #1437

Two instances in one file read as a habit rather than an oversight, so I asked for a sweep. `MeshA2ADispatcher` is now clean (all three collections `Method`-keyed), but **`MeshRouteRegistry.routesByHandler`** (`:39`) has the same key with a bare `put` — last-write-wins, no guard, unlike `MeshA2ARegistry`'s `putIfAbsent`. Probed:

```
PROBE getByHandlerMethodId(com.example.Ctrl.handle) -> beta
      <-- what MeshRouteHandlerInterceptor.preHandle uses for BOTH overloads
```

Arguably worse than the A2A instances: the interceptor builds the **injected dependency list** from that metadata, so `/a` receives `/b`'s proxies. Not fixed here — `getByHandlerMethodId` is a public accessor and the change touches a pre-existing contract rather than this PR's new code. **#1437**, along with two weaker pre-existing instances (`MeshToolWrapperRegistry` keys on the bare method name with no class qualifier; `MeshLlmRegistry.configsByMethod`).

One fingerprint worth recording: the pre-existing `MeshA2ADispatcherJobSubmitterInjectionTest` fixture builds its handler id as `"SubmitterCapturingBean.handler[" + skillId + "]"` — the test author hand-disambiguated a key production does not. That workaround has been in the test tree since #936.

## Validation

`MeshA2ADispatcherOverloadedHandlerTest` 4/4. `MeshA2ADispatcherJobSubmitterInjectionTest` still 6/6 — it pins the per-surface caching contract including `assertSame` across repeat requests, so the re-key did not weaken caching. Suite **1112 → 1116**; no pre-existing test changed state. `MeshPositionalBinderTest` 12/12 and `MeshToolWrapperUnresolvedDepPositionTest` 5/5 unedited.

PR #1438, opened 2026-07-30
feat(typescript)!: bind mesh.route and mesh.a2a.mount dependencies by position

dhyansraj:
## Summary

**PR C of the #1401 sequence — the breaking TypeScript conversion.** `mesh.route` and `mesh.a2a.mount` dependencies now bind **by position**, matching Python and Java (#1436). `mesh.addTool` was already positional and is untouched.

The issue's table was wrong on one point: **`a2a.mount` was also capability-keyed**, so this converts three sites, not one.

## No in-tree app changes behaviour

A base↔head comparator (TS compiler API) built each handler's `local identifier → capability` table under each tree's own rule and diffed by identifier:

```
handlers=35 (route=27 a2a=8)  bound-slots=30  BINDING DIFFERENCES=0   UNMATCHED: none
```

Java's equivalent run was 21 handlers / 0 differences. Every in-tree handler declares exactly one dependency, so this is 30 single-slot rewrites plus one `deps["greet-lucky"] → [greetLucky]` rename class.

## The migration guard — and why it is bounded to declared capabilities

Silent misbinding is structurally impossible here: string keys and numeric indices are disjoint, so a stale `deps.cap` on an array yields `undefined`, never a wrong proxy. But `undefined` surfaces far from the cause as `Cannot read properties of undefined (reading 'call')`, and casts like `deps["cap"] as McpMeshTool | null` — present in our own examples — defeat the compiler entirely. So the array is Proxy-wrapped and throws with the index named:

```
mesh.route dependencies are positional as of 3.4.0.
You accessed `deps.add`; "add" is declared dependency [0].
Rewrite the handler as:  async (req, res, [add, greet_lucky]) => { ... }
```

**The originally-specified unbounded trap with a positive allowlist is not shippable, and this is the one deliberate deviation from the brief.** Its first host was vitest, which probes `$$typeof`, `@@__IMMUTABLE_ITERABLE__@@`, `@@__IMMUTABLE_RECORD__@@`, `nodeType`, `tagName`, `hasAttribute` and `toJSON` on a single `toHaveBeenCalledWith` — and the throw surfaced as `PrettyFormatPluginError`, **hiding the real assertion**. That key set is library-invented and not enumerable in advance, so no allowlist can close over it, and getting it wrong breaks users' currently-passing route tests.

Inverted instead: the trap fires on **declared capabilities**, checked before anything else. That catches 100% of the actual hazard — an un-migrated handler reading a declared capability by name, cast form included — while being immune to host probing. A capability named `map` or `length` still throws rather than silently returning the array method. Reading an *undeclared* key returned `undefined` before this change too: a typo, not a migration signal, with no rewrite to prescribe.

The rationale and the measured probe list are in the module docstring, and the pass-through half is pinned by 13 tests so a future "tighten this" refactor fails loudly.

Verified live against the built `dist/`: `deps[0]`, destructuring, `.length`, `Array.isArray`→true, spread, `for...of`, `JSON.stringify`, `console.log`/`util.inspect`, `.filter`, and undeclared-key→`undefined` all behave correctly.

## Two decisions

**`A2ADependencies` kept and repointed, not deprecated.** Repointing it at the array type achieves both goals at once: the name survives so barrels and imports are unchanged, *and* `mount<{ date_service: McpMeshTool }>` stops compiling, because an object type no longer satisfies the constraint (TS2344, pinned by `@ts-expect-error`). No `@deprecated` marker: the type was not *replaced*, it changed meaning, so there is no "use X instead" to point at — and it is still `A2AHandler`'s default type argument, so the marker would strike through the recommended usage. The generic is now more useful than before: `mount<[McpMeshTool, McpMeshTool | null]>` gives per-slot typing. `RouteDependencies` gets the same treatment; both alias a newly-exported `PositionalDependencies`.

**Duplicate capabilities — the required perimeter reverts to per-index.** The old per-capability dedupe was correct *because* injection collapsed duplicates into one `deps[cap]`. Positionally each declaration owns its index, so deduping would wave through a required dependency the handler reads as `null`. Two tests: required-dead-sibling now 503s, and both-live gives each slot its own distinct proxy.

## Two more things the conversion forced

**`route()`'s union handler parameter had to go, or every converted example fails to compile.** A union of the 3- and 4-param signatures defeats contextual typing for an array binding pattern — TS falls back to the implied tuple `[add] → [any]` and rejects the assignment. The keyed form masked this, because `{ add: any }` was satisfied by `Record<string, …>`. `route`/`routeWithConfig` are now generic and take `MeshRouteHandlerWithNext<D>`, to which a 3-param handler is assignable.

Verified against the **built SDK**: base `express-api` had 37 type errors, head has 8 — and those 8 are a pre-existing `express@^4` + `@types/express@^5` mismatch in that example's own `package.json`, unrelated and untouched. Properly-typed deps also surfaced two latent example bugs (`return res.json()` from a `void` handler, `parseFloat(unknown)`), both fixed.

**`meshctl scaffold api --lang typescript` was a missing site.** `cmd/meshctl/templates/typescript/api/src/index.ts.tmpl` emitted `async (req, res, { hello })` — a freshly scaffolded app would not have compiled. The plan had recorded scaffold as unaffected; it was wrong. Converted and verified end to end: built `meshctl`, scaffolded a gateway, `npm install`, `tsc -p tsconfig.json` → **exit 0**.

## Validation

`npm test` **1213 → 1240** (85 → 87 files). `tsc --noEmit` clean on both the express4 and express5 configs. `go build ./...` clean. Slot preservation extended to route and a2a — an unavailable middle dependency leaves later slots in place — with the a2a case driven through a real `tasks/send`.

## Out of scope

Java (#1436, merged), docs/man/release notes (PR D), and the `jobBeforeProxy` finding from PR A — still awaiting a decision, no guard added.

Refs #1401

## Summary by CodeRabbit

* **New Features**
  * TypeScript route and A2A handlers now receive dependencies as positional arrays, aligned with their declared order.
  * Unavailable dependencies remain represented by `null`, preserving dependency positions.
  * Added a public type for positional dependencies.
* **Bug Fixes**
  * Improved handling of duplicate and unresolved dependencies.
  * Missing required dependencies now return clear 503 responses without continuing execution.
* **Documentation**
  * Updated TypeScript examples and guidance to demonstrate positional dependency access and migration behavior.
* **Tests**
  * Added comprehensive coverage for positional injection, unresolved slots, validation, and compatibility safeguards.

### DIFF
--- a/cmd/meshctl/templates/typescript/api/src/index.ts.tmpl
+++ b/cmd/meshctl/templates/typescript/api/src/index.ts.tmpl
@@ -34,14 +34,16 @@ app.get("/health", (req: Request, res: Response) => {
  * GET /api/hello
  *
  * Example endpoint that calls the "hello" mesh capability.
- * The mesh.route() middleware resolves the dependency and injects
- * it into the handler via the deps object.
+ * The mesh.route() middleware resolves the dependency and injects it into
+ * the handler BY POSITION: deps[i] is the i-th declared dependency, so
+ * destructure the deps parameter as an array.
  */
 app.get(
   "/api/hello",
-  mesh.route([{ capability: "hello" }], async (req, res, { hello }) => {
+  mesh.route([{ capability: "hello" }], async (req, res, [hello]) => {
     if (!hello) {
-      return res.status(503).json({ error: "hello capability unavailable" });
+      res.status(503).json({ error: "hello capability unavailable" });
+      return;
     }
 
     try {
@@ -56,20 +58,21 @@ app.get(
 
 // ===== MULTI-DEPENDENCY EXAMPLE (uncomment and adapt) =====
 //
-// Declare multiple dependencies on a single route.
-// Each dependency is available on the deps object by capability name.
+// Declare multiple dependencies on a single route. Dependencies bind BY
+// POSITION — the Nth declared dependency is deps[N] — so destructure them
+// in the same order they are declared.
 //
 // app.get(
 //   "/api/combined",
 //   mesh.route(
 //     [{ capability: "time_service" }, { capability: "system_info" }],
-//     async (req, res, deps) => {
+//     async (req, res, [timeService, systemInfo]) => {
 //       const results: Record<string, unknown> = {};
-//       if (deps.time_service) {
-//         results.time = await deps.time_service({});
+//       if (timeService) {
+//         results.time = await timeService({});
 //       }
-//       if (deps.system_info) {
-//         results.system = await deps.system_info({});
+//       if (systemInfo) {
+//         results.system = await systemInfo({});
 //       }
 //       res.json(results);
 //     }
@@ -82,13 +85,14 @@ app.get(
 //
 // app.post(
 //   "/api/calculate",
-//   mesh.route([{ capability: "add" }], async (req, res, { add }) => {
+//   mesh.route([{ capability: "add" }], async (req, res, [add]) => {
 //     if (!add) {
-//       return res.status(503).json({ error: "Calculator unavailable" });
+//       res.status(503).json({ error: "Calculator unavailable" });
+//       return;
 //     }
 //     const { a, b } = req.body;
 //     const result = await add({ a, b });
-//     res.json({ operation: "add", a, b, result: parseFloat(result) });
+//     res.json({ operation: "add", a, b, result: parseFloat(String(result)) });
 //   })
 // );
 
--- a/docs/a2a/producer.md
+++ b/docs/a2a/producer.md
@@ -117,7 +117,7 @@ The simplest case — the upstream returns within seconds, so there is no parkin
 
     ```typescript
     import express from "express";
-    import { mesh, type McpMeshTool } from "@mcpmesh/sdk";
+    import { mesh } from "@mcpmesh/sdk";
 
     process.env.MCP_MESH_HTTP_PORT = process.env.MCP_MESH_HTTP_PORT ?? "9090";
     process.env.MCP_MESH_AGENT_NAME = process.env.MCP_MESH_AGENT_NAME ?? "date-a2a-agent";
@@ -135,12 +135,13 @@ The simplest case — the upstream returns within seconds, so there is no parkin
         tags: ["system", "date"],
         dependencies: ["date_service"],
       },
-      async (deps, _payload) => {
-        const dateService = deps.date_service as McpMeshTool | null;
+      async ([dateService], _payload) => {
+        // Dependencies bind BY POSITION (#1401): deps[0] is `date_service`.
+        // A slot holds `null` until the registry resolves it.
         if (dateService == null) {
           return { error: "date_service not yet resolved" };
         }
-        return { date: await dateService.call({}) };
+        return { date: await dateService({}) };
       },
     );
 
--- a/examples/typescript/express-api/index.ts
+++ b/examples/typescript/express-api/index.ts
@@ -61,19 +61,21 @@ app.get("/health", (req, res) => {
  */
 app.post(
   "/api/add",
-  mesh.route([{ capability: "add" }], async (req, res, { add }) => {
+  mesh.route([{ capability: "add" }], async (req, res, [add]) => {
     if (!add) {
-      return res.status(503).json({ error: "Calculator service unavailable" });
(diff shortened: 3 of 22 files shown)


dhyansraj:
Verified four findings against current code in `efdd8d1c4`: **three fixed, one skipped.** One "nitpick" turned out to be a live bug in two production code paths.

## The `.call({})` nitpick is a real defect — proven at a real dispatch site

`McpMeshTool` is a **callable interface** (`types.ts:625-637`) whose only named method is `callTool`. There is no `call` member — so `.call` resolves to **`Function.prototype.call`**, consuming the first argument as `thisArg`:

```
[.call({format:'ISO'})] tool received args: []                    ← ZERO arguments
[({format:'ISO'})]      tool received args: [ { format: 'ISO' } ]
tool.call === Function.prototype.call : true
```

It type-checks silently because both parameters (`args?`, `options?`) are optional, so `strictBindCallApply` accepts `call(thisArg)` with nothing further.

Five sites, and **two were live**:

| site | reachable | would a test have caught it? |
|---|---|---|
| `mount.ts:198` (JSDoc) | no | teaches the bug |
| `positional-deps-types.spec.ts:48` | no — never invoked | no; compile-only, and it compiled clean |
| `examples/typescript/producer-date-agent/index.ts:107` | **yes**, every `tasks/send` | no test |
| `uc29` integration fixture `:51` | **yes**, per dispatch | **no** — tc02 asserts envelope shape, never the args, and `date_service` takes no required arguments, so the dropped payload produced an identical result. Green throughout |
| `docs/a2a/producer.md:143` | no | — |

All rewritten to `dateService({})`. `examples/complex/DEVELOPMENT_GUIDE.md:346` also matches the pattern but is Python, where `.call()` is a real method — left alone.

## A doc surface this PR's own sweep had missed

While fixing those lines: **`docs/a2a/producer.md` was still on the pre-#1401 capability-keyed shape** — `const dateService = deps.date_service as McpMeshTool | null`. As shipped it would hit the new migration guard and **throw at request time**, since `date_service` is a declared capability. Migrated to `async ([dateService], _payload)`.

Worth noting because `positional-deps.ts`'s own header cites that file as one of the two places using the defeating cast — so the module documented the hazard while the doc still demonstrated it. Re-grepped `deps\.\w` across `docs/`, `man/content/` and the TS scaffold templates: **zero remaining**.

## mount's doc example

Now narrows before calling and the tuple hint is nullable, matching `route.ts:69`. The identical unachievable `mount<[McpMeshTool]>` hint on `A2ADependencies` in `dispatcher.ts` is fixed too — leaving it would have re-split the two surfaces the same way.

## Shared helper

`resolvePositionalDeps` in `positional-deps.ts`, both callers delegating. Module-exported only, so the public API is unchanged. Two design points:

- The registry parameter is **structural** (`PositionalDepsSource`), not `RouteRegistry` — `route.ts` owns that class and imports this module, so importing it back would create a cycle.
- It returns `{ values, deps }` rather than just the wrapped array. `route()`'s 503 guard reads slots by index, and index access on the guarded proxy goes through the `get` trap as the string `"0"` — so a capability literally named `"0"` would turn a 503 into a thrown 500. Returning the unguarded `values` for framework-side pre-flight keeps the guard bit-identical.

## Skipped

**Sanitising `req.method` / `req.path` in `console.warn`.** Real class, but **pre-existing**: `main:route.ts:479` carries the identical interpolation and `:518` builds `routeName` the same way. Reachability is low — Node's HTTP parser rejects control characters in the request target and `req.path` does not percent-decode. Out of scope for a positional-binding PR; no evidence found that it is more reachable than that.

## Validation

- Extraction changed no behaviour: a comparator running the pre-extraction inline logic from both files verbatim against the helper, over declared-count 0–3 × every resolved-slot bitmask × route-registered/missing — **19 scenarios, 38 handlers, 0 binding differences**.
- `npm test` **1240 / 87 files**, identical to baseline. `tsc --noEmit` clean on both express4 and express5 configs.
- Required-dependency guard green, including both per-index duplicate-capability cases.
- `git diff --name-only` → exactly 8 intended files.

## Worth its own look, independent of this PR

The uc29 suite exercises a dependency whose tool takes **no arguments**, so no A2A producer test can notice a dropped payload — which is precisely why this bug lived there undetected. A fixture that passes and asserts a non-empty argument would close that hole.

PR #1439, opened 2026-07-31
docs(di): state the positional dependency-injection contract on every surface

### LINKED ISSUE #1401: Java DI: four matching rules across five injection sites — align @MeshRoute and @MeshA2A on the positional contract (3.4.0, breaking)

dhyansraj:
## Summary

**PR D — the final PR of the #1401 sequence.** Documentation only, no code changes.

Dependency injection is now positional in every runtime and every decorator (#1434, #1435, #1436, #1438). This states that on every surface that described the old behaviour, and ships a migration guide covering both runtimes.

36 files. The deliverable is `docs/migration/3.4-positional-di.md` — what breaks, what does not, how to find your exposure on 3.3.x before upgrading, and before/after for all four shapes. Mirrored offline in `man upgrading`, with a Migration section added to `mkdocs.yml`.

## Golden delta, with the reason

`wantInlineCodeSpans` **1605 → 1626 (+21)**. `wantListCodeSpans` 488 and `wantMarkupListLines` 433 **unmoved**.

The tests iterate `ListGuides()` and call `GetGuide(name)`, which returns the **default** variant — the `_java` and `_typescript` files are never sampled. Of the +61 spans added across all variants, only `a2a.md` (+6) and `upgrading.md` (+15) are counted, and neither added a markdown list line. Arithmetic verified per-file before touching the constant.

## Two findings that changed the work

**1. `docs/` must not be wholesale-regenerated — the brief's instruction was wrong.**

Running `scripts/generate_docs_from_man.py` at HEAD produces a **23-file, +1145/−587** diff. The generator-managed pages carry substantial hand-authored content the man sources do not have: `!!! note` admonitions, the DDDI cross-link, an entire multimodal section, required-dependency prose. `git log -- docs/java/dependency-injection.md` shows six feature PRs editing it directly.

Regenerating would have silently deleted all of it. The managed trees were restored to HEAD and the man-derived changes applied to their `docs/` counterparts by hand.

Separately: `fastapi`, `spring-boot` and `express` have **no man source at all**, so `docs/*/{fastapi,spring-boot,express}-integration.md` are orphaned and now hand-maintained. Worth its own issue.

**2. 25 TypeScript `addTool` snippets taught an object destructure — and `addTool` has always been positional.**

Across `tags`/`health`/`proxies`/`capabilities`/`headers`/`decorators` `_typescript.md`, `schema-matching.md` and `docs/concepts/dddi.md`. These could never have run. Unrelated to this change; fixed while sweeping.

## Every stale surface found

The brief said to treat the inventory as untrusted, because PR C found `docs/a2a/producer.md` still teaching the keyed shape *after* the plan declared docs clean. That proved correct again — three surfaces below were not in any prior inventory.

| Surface | |
|---|---|
| `dependency-injection_java.md:74‑90` — the paragraph *stating* the by-name rule, plus 5 more sites | fixed |
| `api_java.md:59` "injected by parameter name", `:102` "when the parameter name does not match" | fixed |
| `service-views_java.md:49`, `decorators_java.md:81`, `headers_java.md:173,233` — `@MeshInject` described as "only fires on `@MeshRoute`", now false | fixed |
| `api_typescript.md` ×7 keyed handlers + "maps to the first key" prose; `decorators_typescript.md:73` + 3 snippets | fixed |
| **`docs/a2a/producer.md:30`** — prose still taught the keyed shape after #1438 fixed the snippet 100 lines below | fixed |
| **`docs/java/spring-boot-integration.md:54,76‑92`** — an entire "Parameter Name Matching (Recommended)" section | fixed |
| **`docs/typescript/express-integration.md`** ×5 keyed handlers | fixed |
| `docs/concepts/{routes,service-views,streaming}.md` | fixed |
| 25 `addTool` object destructures (see above) | fixed |
| 2 scaffold templates: `python/api/main.py.tmpl:37` "matched by parameter name" — wrong for Python too — and the Java `api` template | fixed |
| `docs/concepts/schema-matching.md:167`, `docs/reference/kwargs.md:257` | checked, correct as-is |
| `examples/**` | clean — converted in #1436/#1438 |

## Validation — every snippet was executed, not eyeballed

This sequence has repeatedly turned up doc examples that do not work, so none were taken on trust.

- **Java**: every snippet compiles against the starter (`javac exit=0`), and the **shipped `MeshLegacyBindingDetector`** was run over each one — 8 `OK`, and the migration guide's deliberate BEFORE example correctly `ERROR`s. All quoted error text was captured from the real components via a harness, not transcribed from PR descriptions.
- **TypeScript**: 13 route/express + 3 a2a snippets typechecked against the built `dist/` under `strict` — **0 errors**. The documented negative was confirmed to actually fail: `mount<{ date_service: McpMeshTool }>` → `TS2344`. Also fixed 9 pre-existing `result.message` accesses on `Promise<unknown>` in snippets touched.
- `go build ./...` clean; `go test ./src/core/cli/man/...` passes with the updated golden.
- `meshctl man` rendered from a **locally built** `./cmd/meshctl` (bypassing the npm `@mcpmesh/cli` on PATH, which serves a stale embedded page): new text confirmed in `upgrading`, `dependency-injection --java/--typescript`, `api --java/--typescript`, `a2a`, `environment`. All 9 topics render; `man --list` still 62.
- `mkdocs build --strict`: no broken-link or nav warnings.

## Pre-existing, not fixed here

**The documented `addTool` dependency idiom does not typecheck under `strict`.** `ToolConfig.execute`'s rest parameter is `...deps: (McpMeshTool | MeshJob | null)[]`, so the form taught at `dependency-injection_typescript.md:51`, in the SDK's own JSDoc (`types.ts:589-592`), and used by `examples/typescript/greeter` fails `TS2322` (`MeshJob` not assignable to `McpMeshTool | null | undefined`). Verified verbatim against HEAD.

The fix is making `ToolConfig` generic over its dependency tuple — exactly what #1438 did for `route<D>` — which is a code change and out of scope for a docs PR.

## Not included

`RELEASE_NOTES.md` — the previous nine touches are all `release:` commits, so the 3.4.0 note belongs in the release PR. The durable surfaces (migration guide, mkdocs nav, `man upgrading`) carry the content meanwhile.

The `jobBeforeProxy` guard (#1435) and whether parameter-name crossover should be fatal (#1436) are both described as still open; nothing presents them as settled.

Closes #1401

## Summary by CodeRabbit

* **Documentation**
  * Clarified that Java and TypeScript dependencies bind by declaration order rather than parameter or capability names.
  * Added 3.4.0 migration guidance, including updated examples, types, diagnostics, and error handling.
  * Documented nullable unresolved dependencies and recommended 503 responses where applicable.
  * Expanded strict dependency-injection diagnostics for Java and Python, including startup failure behavior.

### DIFF
--- a/cmd/meshctl/templates/java/api/src/main/java/{{ .JavaPackagePath }}/{{ .NamePascal }}Application.java.tmpl
+++ b/cmd/meshctl/templates/java/api/src/main/java/{{ .JavaPackagePath }}/{{ .NamePascal }}Application.java.tmpl
@@ -55,8 +55,10 @@ class ApiController {
      * Call the "hello" mesh capability.
      *
      * <p>The @MeshRoute annotation declares a dependency on the "hello" capability.
-     * At request time, the framework resolves the dependency and injects it as a
-     * McpMeshTool parameter (matched by @MeshInject name).
+     * At request time, the framework resolves the dependency and injects it
+     * BY POSITION: the Nth declared dependency fills the Nth McpMeshTool
+     * parameter. @MeshInject does not select a dependency -- it asserts the
+     * one position assigns, and boot fails if the two disagree.
      */
     @GetMapping("/api/hello")
     @MeshRoute(dependencies = @MeshDependency(capability = "hello"))
--- a/cmd/meshctl/templates/python/api/main.py.tmpl
+++ b/cmd/meshctl/templates/python/api/main.py.tmpl
@@ -33,8 +33,9 @@ async def hello_endpoint(request: Request, hello: McpMeshTool = None):
     Example API endpoint that calls the "hello" mesh capability.
 
     The @mesh.route decorator declares dependencies on mesh capabilities.
-    At request time, the framework resolves each dependency and injects
-    it as a McpMeshTool parameter (matched by parameter name).
+    At request time, the framework resolves each dependency and injects it
+    BY POSITION: the Nth declared dependency fills the Nth McpMeshTool
+    parameter. Parameter names are never consulted.
     """
     if not hello:
         raise HTTPException(status_code=503, detail="hello capability unavailable")
--- a/docs/a2a/producer.md
+++ b/docs/a2a/producer.md
@@ -21,13 +21,13 @@ Issue #972: the same code path also flips `a2a_producer=true` on the agent paylo
 
     `@MeshA2A(path = "/agents/<skill>", ...)` on a Spring Boot bean method (sibling to `@MeshRoute`). The framework auto-mounts both routes on the application's `DispatcherServlet` — the user owns the Spring Boot lifecycle (`SpringApplication.run(...)`).
 
-    Mesh dependencies are declared via `@MeshDependency` entries on the annotation and injected at `@MeshInject` parameter slots, identical to the `@MeshRoute` DDDI path.
+    Mesh dependencies are declared via `@MeshDependency` entries on the annotation and injected **by position** into the handler's `McpMeshTool` parameters, identical to the `@MeshRoute` DDDI path. (Changed in 3.4.0 — `@MeshA2A` used to bind by name; see [Migrating to positional DI](../migration/3.4-positional-di.md).)
 
 === "TypeScript"
 
     `mesh.a2a.mount(app, config, handler)` on a user-owned Express app (sibling to `mesh.route(...)`). The user owns the Express app AND the `app.listen()` lifecycle — same shape as `mesh.route(...)` HTTP handlers. The mesh api-runtime pipeline picks up the mounted A2A surface from the `A2AProducerRegistry` and registers the agent with the registry as `agent_type=a2a` on each heartbeat.
 
-    Mesh dependencies are declared via the `dependencies` array on the mount config and supplied to the handler under the `deps` argument keyed by capability name, identical to how `mesh.route(...)` injects resolved `McpMeshTool` proxies.
+    Mesh dependencies are declared via the `dependencies` array on the mount config and supplied to the handler as a **positional** `deps` array — `deps[i]` is the i-th declared dependency, `null` until it resolves — identical to how `mesh.route(...)` injects resolved `McpMeshTool` proxies. (Changed in 3.4.0 — the argument used to be an object keyed by capability; see [Migrating to positional DI](../migration/3.4-positional-di.md).)
 
 ## Sync handler
 
--- a/docs/concepts/dddi.md
+++ b/docs/concepts/dddi.md
@@ -64,7 +64,10 @@ DDDI uses **capability-based discovery** with tag matching, not hardcoded servic
       name: "greet",
       capability: "greeting",
       dependencies: ["calculator"],
-      execute: async (args, { calculator }) => {
+      execute: async (args, calculator: McpMeshTool | null = null) => {
+        if (!calculator) {
+          return "Calculator unavailable";
+        }
         const result = await calculator({ expression: "2+2" });
         return `The answer is ${result}`;
       },
--- a/docs/concepts/routes.md
+++ b/docs/concepts/routes.md
@@ -26,7 +26,7 @@ route handler  ──── injects proxy for capability "session_state.record_q
 resolved provider agent ──► result ──► HTTP response
 ```
 
-The per-runtime surface differs — Java pairs `@MeshRoute(dependencies = {…})` with a `@MeshInject` parameter typed `McpMeshTool<…>`; Python and TypeScript declare the dependency on the route and receive the proxy — but the shape is the same everywhere: one route, the specific capabilities it names, one resolved proxy per capability, called inline in the handler. The exact declaration syntax lives in the dependency-injection guides linked below.
+The per-runtime surface differs in syntax — Java declares `@MeshRoute(dependencies = {…})` and takes `McpMeshTool<…>` parameters, Python declares on the decorator and takes typed parameters, TypeScript declares on `mesh.route(...)` and destructures a `deps` array — but the shape is the same everywhere: one route, the specific capabilities it names, one resolved proxy per capability, called inline in the handler. **Binding is positional in all three**: the Nth declared capability lands in the Nth injectable slot, and names are never consulted. The exact declaration syntax lives in the dependency-injection guides linked below.
 
 Because the proxy is an ordinary resolved dependency, everything DDDI gives a consumer applies here too: the capability hot-swaps to a better or healthier provider between requests, calls thread call context and appear in the audit trail, and an unavailable optional capability injects a null/unavailable proxy that the handler can fall back around.
 
@@ -68,8 +68,9 @@ A route is a DDDI consumer with an HTTP trigger. Each capability it injects is o
 This page stays at the concept level; the exact route declaration, dependency wiring, and injection syntax live in the runtime guides:
 
 - [Dependency Injection (Python)](../python/dependency-injection.md#required-dependencies) — `@mesh.route`, required deps, the 503 perimeter
-- [Dependency Injection (Java)](../java/dependency-injection.md#views-and-meshroute) — `@MeshRoute` + `@MeshInject`, per-capability injection
+- [Dependency Injection (Java)](../java/dependency-injection.md#views-and-meshroute) — `@MeshRoute`, positional pairing, per-capability injection
 - [Dependency Injection (TypeScript)](../typescript/dependency-injection.md#required-dependencies) — `mesh.route()`, required deps, the 503 perimeter
+- [Migrating to positional DI (3.4)](../migration/3.4-positional-di.md) — the Java and TypeScript route binding change
 
 ## See Also
 
--- a/docs/concepts/schema-matching.md
+++ b/docs/concepts/schema-matching.md
@@ -114,7 +114,7 @@ agent.addTool({
     },
   ],
   parameters: z.object({}),
-  execute: async ({}, { lookup_employee }) => { /* ... */ },
+  execute: async ({}, lookupEmployee: McpMeshTool | null = null) => { /* ... */ },
 });
 ```
 
--- a/docs/concepts/service-views.md
+++ b/docs/concepts/service-views.md
@@ -185,7 +185,7 @@ A service view is a grouping convenience, not a default. The [Zero wire and regi
 
 ### When per-capability injection is the better fit
 
-- **A handler uses only 1–2 capabilities of the group** — declare exactly those with per-capability injection: a Python dependency on the specific `@mesh.tool` handler, a Java `@Selector` + `@MeshInject` → `McpMeshTool`, or a TypeScript dependency → `McpMeshTool`. Don't inject a 13-method view to reach two methods.
+- **A handler uses only 1–2 capabilities of the group** — declare exactly those with per-capability injection: a Python dependency on the specific `@mesh.tool` handler, a Java `@Selector` / `@MeshDependency` → `McpMeshTool`, or a TypeScript dependency → `McpMeshTool`. Don't inject a 13-method view to reach two methods.
 - **The consumer is a route handler** (`@mesh.route` / `mesh.route(...)` / `@MeshRoute`) — views are a tool-parameter surface and are **rejected** on routes anyway; a route declares the specific dotted capability it needs and injects that proxy.
 - **Precise `required`-gating matters** — a view's `required` methods gate the handler as a set; if those methods span multiple providers, the handler is gated on capabilities it never calls. Per-capability injection gates on exactly what the handler invokes.
 - **Dep-count legibility / registry footprint matters** — when you want `meshctl list` to show what a handler actually depends on, not the whole group behind one type.
--- a/docs/concepts/streaming.md
+++ b/docs/concepts/streaming.md
@@ -91,8 +91,11 @@ Once a producer streams, two consumer code paths are available — and both work
 
     app.post("/api/chat", mesh.route(
       [{ capability: "chat" }],
-      async (req, res, { chat }) => {
-        if (!chat) return res.status(503).json({ error: "chat unavailable" });
+      async (req, res, [chat]) => {
+        if (!chat) {
+          res.status(503).json({ error: "chat unavailable" });
+          return;
+        }
         await mesh.sseStream(res, chat.stream({ prompt: req.body.prompt }));
       }
     ));
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -108,13 +108,19 @@ export MCP_MESH_TOOL_WORKERS=1
 export MCP_MESH_TOOL_ISOLATION=true
 ```
 
-### Strict DI Diagnostics (Python)
+### Strict DI Diagnostics (Python, Java)
 
 ```bash
 # Default: false. When truthy, ambiguous or skipped dependency-injection
-# configurations raise StrictDIError at decoration/startup instead of
-# warning. Injection semantics are unchanged — only the diagnostic
-# severity is promoted.
+# configurations fail at decoration/startup instead of warning. Injection
+# semantics are unchanged — only the diagnostic severity is promoted.
+#
+# - Python: raises StrictDIError at decoration/startup.
+# - Java: promotes the boot-time dependency/parameter arity mismatch and
+#   the @MeshRoute / @MeshA2A legacy-order warning to a startup failure.
+#   A contradicting @MeshInject value is fatal either way — it asserts the
+#   dependency position already assigns, so a false assertion is never
+#   survivable. Accepts 1 / true / yes.
 export MCP_MESH_STRICT_DI=true
 ```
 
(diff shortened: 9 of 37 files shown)


dhyansraj:
Nine findings verified and addressed in `b5536ddba`. **Two were factual errors in the guide whose whole job is making the migration safe** — worth calling out, since docs-last does not mean lower-stakes.

## The two that mattered

**`@MeshTool` was listed as unchanged, and it changed.** Its *binding* is unchanged — always positional — but #1436 made `@MeshInject` **honoured** on `@MeshTool` parameters, where it had been silently ignored. A disagreeing annotation there now fails boot. A reader was being told they had no exposure at exactly the site that gained one. Qualified at all three surfaces.

**"Dropping them entirely is equally valid" was unsafe on 3.3.** True on 3.4; on 3.3 `@MeshInject` still *selects*, so dropping falls back to parameter-name matching and can rebind — or inject `null` where the name matches nothing. The guide actively encourages doing the sweep *before* upgrading, which made that sentence a live trap. Now scoped to "once you are on 3.4", with the 3.3 behaviour spelled out.

The broader `:91-96` claim was **narrowed, not deleted**: the two *reorder* fixes genuinely are behaviour-preserving on 3.3 (name-neutral under the old rule) and stay as-is; only *adding* an annotation where none existed is called out as preserving 3.3 only where the parameter name already selected that dependency.

`docs/upgrading.md:45-47` — **skipped, no such text exists.** Its entire 3.4 content is the admonition at `:13-21`; the drop-annotations advice lives only in the man page and the migration guide, both now qualified.

## The judgement calls both turned out to be real

**Finding 9 was not ambiguity, it was an off-by-one.** `a2a.md` read *"the Nth declared dependency binds to … `deps[N]`"* — mixing a 1-based ordinal with a 0-based index. Literally wrong. Now: "the Nth element of the TypeScript `deps` array, which is **zero-based**: `deps[0]` is the first declared dependency."

**Finding 8's wording did conflate the two.** "optional and behaviour-neutral … and fails loudly if someone later reorders one end" is self-contradictory, and the Order bullet below says a correct annotation *silences* the legacy-order warning. Now separates runtime from boot: changes nothing about what binds, but is not inert at boot.

**The nitpick was three more sites, not one.** `service-views_java.md:49` was already fixed in the commit under review, but the same false claim survived at `:7`, `:59` and `:68`, all presenting `@MeshInject` as the per-capability injection mechanism for `@MeshTool` — which is `@Selector` plus position. `man service-views --java` now contains no `@MeshInject` at all.

## The snippet that could not run

`capabilities_typescript.md` declared `weatherData` and referenced `weather_data`. Shown failing verbatim before the fix:

```
a_capabilities_before.ts(12,9): error TS2552: Cannot find name 'weather_data'. Did you mean 'weatherData'?
a_capabilities_before.ts(13,20): error TS2552: Cannot find name 'weather_data'. Did you mean 'weatherData'?
```

That is the fourth non-working doc example this sequence has surfaced, after `.call({})` in two live paths, 25 `addTool` object destructures, and the scaffold template.

## Validation

- **TypeScript**: every touched snippet typechecks under `strict` against the built `dist/`. An **untouched control** (`dependency-injection_typescript.md:43-57`) emits the identical residual `TS2322`, confirming it is the pre-existing `ToolConfig` issue rather than something introduced here. Isolating the handler bodies from `ToolConfig`'s typing: **0 errors**.
- **Golden 1626 → 1633 (+7)**, mechanically verified per-file against `3269209c0`: `upgrading.md` +6, `a2a.md` +1. No `_java`/`_typescript` file contributes, since the tests sample default variants only. The `renderer_test.go` comment is updated in the same commit with the reason. List goldens unmoved.
- `go build ./...` clean, `go test ./src/core/cli/man/...` ok, `mkdocs build --strict` with only the pre-existing git-log warning (verified present at the base commit).
- Rendered from a **locally built** `./cmd/meshctl`; all eight affected topics confirmed. Noted: a ```text fence inside a blockquote renders as literal text — confirmed pre-existing by building at `3269209c0`, unchanged by the tag.
- `RELEASE_NOTES.md` identical to `main`.

## Follow-up worth its own issue

The residual `TS2322` means **no `addTool`-with-dependencies snippet in the docs typechecks under `strict`** — including the SDK's own JSDoc at `types.ts:589-592` and `examples/typescript/greeter`. `ToolConfig.execute`'s rest parameter is `...deps: (McpMeshTool | MeshJob | null)[]`, so `MeshJob` is not assignable to a `McpMeshTool | null` slot. The fix is making `ToolConfig` generic over its dependency tuple, exactly as `route<D>` got in #1438. Say the word and I will file it.

dhyansraj:
All three findings verified and fixed in `c49103127`. **Both substantive claims were confirmed empirically before being rewritten**, and the wrong exemption turned out to exist on three surfaces, not one.

## The one-dependency exemption was wrong in both runtimes

The guide said *"a handler declaring **one** dependency cannot change meaning"* — which tells a reader to skip handlers that will break. Confirmed against the **shipped** components rather than reasoned about:

**Java** — one declared dependency `get_employee`, parameter `@MeshInject("employee_count")`:

```
severity = ERROR
  1 declared dependency: [0] 'get_employee'
    slot 0 = parameter 0 (McpMeshTool stats)
        @MeshInject("employee_count") asserts:  NO MATCH among the declared dependencies
        binds (by position):                    dependency[0] 'get_employee'
>>> boot OUTCOME: IllegalStateException (BOOT FAILS)
```

Unconditional, not gated on `MCP_MESH_STRICT_DI` (`MeshLegacyBindingDetector.java:184`, `case ERROR -> throw`). The 3.3 counterpart returned null with a warning, so the transition is **`null` → boot failure**. Two controls stayed `OK`: a *matching* sole `@MeshInject`, and an unannotated sole dependency whose parameter name disagrees — with one slot, `NAME_CROSSOVER` needs `nameTarget != k`, which one slot cannot produce.

**TypeScript** — one declared dependency, 3.3 callback shape:

```
=== async (req, res, { add }) => ...
>>> OUTCOME: TypeError THROWN
    You accessed `deps.add`; "add" is declared dependency [0].
```

So TypeScript has **no** count-based exemption at all: what changed is the callback shape, not the pairing. The guide now says not to filter by count, with the per-runtime conditions spelled out.

## The `-parameters` blind spot was broader than the truth

*"That blind spot contains no working code"* holds for **unannotated** handlers only. `MeshInjectArgumentResolver` reads `@MeshInject` **before** the parameter-name fallback, so an annotated handler never needed reflection — it is working code compiled without `-parameters`, and it is not in the blind spot: recompiling the probe without `-parameters` still yields `ERROR` and a failed boot, printing `(McpMeshTool <unnamed>)`.

Now narrowed to the unannotated name-bound case, where the argument does hold.

## Fences — the review's count was right, my recount was wrong

I said two; there were three. Line 94 is the *closing* fence of the block opened at line 69, so counting openers in that range gave one block, not two. A whole-file sweep found three — at lines 69, 153 and 239 — all now `text`. Rendered HTML confirms exactly 3 × `language-text highlight`.

## The parallel-maintenance hazard bit again

Checking the siblings was worthwhile: **`man upgrading.md:20` carried the same wrong exemption**, and `docs/upgrading.md` carried a milder form of it ("handlers declaring two or more dependencies may need a source edit"). Both corrected. Two more unlabeled fences in the man page (lines 28 and 93) labelled. Neither sibling carried the `-parameters` claim — grep-verified, it exists only in the migration guide.

Worth recording: `docs/upgrading.md` and `man upgrading.md` are **not** managed by `generate_docs_from_man.py` (no `upgrading` entry), so they are hand-maintained in parallel and have already diverged — the man page carries a full 3.4.0 section, the docs page only an admonition. That is precisely the mechanism that let one safety claim be right in one place and wrong in another.

## Validation

- **Golden 1633 → 1637 (+4)**, all four in `upgrading.md`'s replaced paragraph (`async (req, res, { cap }) => ...`, `@MeshInject` ×2, `null`). Predicted before running; the failure reported exactly +4. List goldens unmoved — the man rewrite was deliberately kept as prose. Fence labels contribute nothing, since the test treats any fence line as a delimiter and continues.
- `go build ./...` clean; `go test ./src/core/cli/...` 5 packages ok.
- `mkdocs build --strict`: the single remaining warning is the environmental `git-revision-date-localized-plugin` one — proven two ways: an unmodified worktree at `b5536ddba` emits **14** of them, and a strict build with only that plugin removed is completely clean.
- Man rendered from a locally built `./cmd/meshctl` and diffed against a binary built at `b5536ddba`: `man --list` identical, `man upgrading` differs in exactly the 3 intended places.
- `RELEASE_NOTES.md` identical to `main`. `git diff --name-only` → exactly 4 files.

dhyansraj:
Added one unrelated line at the repo owner's request, bundled here because GitHub Pages publishes on every merge to `main` — `docs/overrides/main.html:11`:

```diff
-  <script>window.SKY_CONFIG = { apiUrl: "https://chat.mcp-mesh.ai", theme: "auto" };</script>
+  <script>window.SKY_CONFIG = { apiUrl: "https://chat.mcp-mesh.ai", theme: "auto", embeddedAuth: true };</script>
```

This switches on embedded auth for the Sky chatbot widget, coordinated with the skybot side. It is the one remaining line, and it lives in this repo rather than skybot's. No changelog or release-note entry, since it is not part of the positional-DI work.

## Validated beyond "the build passes"

`mkdocs build --strict` cannot catch a malformed inline script — it validates links and nav, not JavaScript — and a stray comma here would break `window.SKY_CONFIG` on *every page of the site* while the build still reported success. So:

- The statement was extracted from built HTML and `node --check`ed, then **`vm`-evaluated on all 119 pages**: every one yields an object with exactly 3 keys and `embeddedAuth === true` as a boolean, not the string `"true"`. 0 failures.
- `embeddedAuth` appears on **119 pages**, matching the baseline's 119 `SKY_CONFIG` pages exactly — so no page lost the widget. The 2 files without it are static copied assets that had no `SKY_CONFIG` before either.
- `mkdocs build --strict`: 1 warning, the environmental `git-revision-date-localized-plugin` one. Confirmed identical by building the unmodified parent commit in a throwaway worktree. (An initial baseline showed 14 — the other 13 were missing `docs/downloads/*.zip`, which is gitignored build output absent from a fresh checkout.)
- `go build ./...`, `go test ./src/core/cli/...` unaffected; no golden or testdata references this file.

`git diff` against the parent is exactly one file, one line.

## One thing worth flagging

The first amend **silently failed** — the SHA did not move, because the `trailing-whitespace` pre-commit hook rewrote the file mid-commit and aborted it. The whitespace it wanted to strip is **pre-existing, on line 6**, unrelated to this change, and stripping it would have made this a two-line diff.

It was committed with `--no-verify` and line 6 restored, so the diff stays at the one line requested. Verified after the fact that nothing else slipped through: the full diff is the single line above, and line 6's trailing whitespace is byte-identical to the parent (`cat -A` shows `  $` on both sides).

The reason that whitespace has survived on `main` at all: `.github/workflows/pre-commit.yml:13` is `if: false # Temporarily disabled - only unit tests required`. **The pre-commit job is disabled repo-wide**, so the local hook and CI disagree about what is committable — the hook blocks a commit that CI would have accepted, and nothing enforces the standard on `main`.

That is the same shape as #1421 (helm checks that only ran on a release tag) and #1399 (`test_bump_version.py` never executed): a check that exists and does not run where it would help. Worth its own look, independent of this PR — either re-enable the job or drop the local hook, but having them disagree means every contributor hits a block that the repo does not actually enforce.

Issue #1443, opened 2026-07-31
Every @MeshRoute handler also registers a phantom route at the controller base path (findAnnotation does not apply @AliasFor)


dhyansraj:
## Summary

Every `@MeshRoute` handler registers an **extra, unintended route at the controller's base path**, because the mapping scan falls through to a meta-annotation lookup that resolves with an empty path.

Measured — three handlers produced four routes:

```
PROBE-C registered route count = 4
PROBE-C   GET:/ -> [beta, alpha]      <-- phantom, arbitrary winner
PROBE-C   GET:/a -> [alpha, beta]
PROBE-C   GET:/b -> [beta, alpha]
PROBE-C   GET:/solo -> [alpha]
```

## Mechanism

`MeshRouteBeanPostProcessor.getMappingInfo` checks `@GetMapping`/`@PostMapping`/etc., then falls through to:

```java
// MeshRouteBeanPostProcessor.java:206
RequestMapping requestMapping = AnnotationUtils.findAnnotation(method, RequestMapping.class);
if (requestMapping != null) {
    String[] paths = requestMapping.value().length > 0 ? requestMapping.value() : requestMapping.path();
    if (paths.length == 0) {
        paths = new String[]{""};          // :210-212
    }
```

`@GetMapping` is itself meta-annotated `@RequestMapping`, so `findAnnotation` **finds it** — but `AnnotationUtils.findAnnotation` does not apply `@AliasFor` overrides, so the `value`/`path` it returns are the meta-annotation's own empty defaults, not the `@GetMapping("/a")` the developer wrote. The empty-path branch then registers at `""`, i.e. the controller's base path.

`AnnotatedElementUtils.findMergedAnnotation` is the API that applies `@AliasFor` and would return the real path.

## Impact today — mild, but the mechanism is wrong

`routesByPath` is not on `MeshRouteHandlerInterceptor`'s lookup path, and its readers dedupe by capability, so this is not currently causing misrouting. What it does cause:

- `getRouteCount()` over-reports, by roughly the number of `@MeshRoute` handlers.
- `getByRoute(method, "/")` returns an arbitrary handler — whichever registered last, and `getDeclaredMethods()` order is unspecified, so that can differ between JVM restarts of the same artifact.
- Any future feature that trusts `routesByPath` inherits a phantom entry per handler.

## Context

[...]

Issue #1442, opened 2026-07-31
Java advertises MCP tools by bare method name, so same-named @MeshTool methods in different classes collide at the protocol layer


dhyansraj:
## Summary

Java advertises an MCP tool under the **bare method name**, with no class qualifier. Two `@MeshTool` methods named the same in different classes therefore collide at the MCP protocol layer — before any mesh registry sees them.

`MeshMcpServerConfiguration.java:156-158`:

```java
private void registerTool(McpStatelessSyncServer server, McpToolHandler handler) {
    // Use method name as tool name to match what registry advertises
    String toolName = handler.getMethodName();
```

This is the deeper form of the defect patched in #1437/#1441. That PR guarded the registry-side collision — a contested bare name now resolves to *neither* wrapper rather than the wrong one — but the MCP tool name is upstream of that guard and still collides.

## Why it is not fixable the same way

`#1441` re-keyed the internal maps by `Method`. That is not available here, because these are **wire identities**:

- `MeshToolBeanPostProcessor.java:322` — `String funcId = targetClass.getName() + "." + method.getName();` — no parameter types
- `MeshLlmRegistry.configsByFunctionId` — `Class.methodName`, same shape
- the MCP tool name above — bare method name, weakest of the three

The Rust core echoes `funcId` back, and the MCP tool name is what a *client* calls. Changing either is a protocol change with cross-runtime consequences, not a refactor. The comment at `:157` says so implicitly — the name is chosen "to match what registry advertises", so the two identities are coupled.

## Scope of the collision

| identity | shape | collides on |
|---|---|---|
| MCP tool name | `methodName` | same method name in **any two classes** |
| `@MeshTool` funcId | `Class.methodName` | **overloads** in one class |
| `MeshLlmRegistry.configsByFunctionId` | `Class.methodName` | overloads in one class |

The MCP tool name is the most exposed: it needs no overloading at all, just two classes that each define `analyze` or `search` or `process`.

## What is unclear and should be established first

- **Do Python and TypeScript have the same exposure?** If they derive the MCP tool name from the capability rather than the function name, Java is the outlier and the fix is to align it. I did not check, and that determines whether this is a Java bug or a cross-runtime contract question.
- **Is a same-named `@MeshTool` in two classes actually rejected somewhere earlier?** If boot already fails on it, the exposure is smaller than it looks — but I found no such check.
[...]

Issue #227, opened 2025-12-14
Fix: @mesh.llm_provider creates conflicting process_chat functions for multiple providers


dhyansraj:
## Problem

When multiple functions in the same agent are decorated with `@mesh.llm_provider`, they all create a function named `process_chat`, causing only the last one to be registered with FastMCP.

```python
from fastmcp import FastMCP
import mesh

app = FastMCP("Multi-LLM Provider")

@mesh.llm_provider(model="anthropic/claude-sonnet-4-5", capability="llm", tags=["claude"])
def claude_provider(): pass  # → becomes process_chat

@mesh.llm_provider(model="openai/gpt-4", capability="llm", tags=["openai"])  
def openai_provider(): pass  # → also becomes process_chat, OVERWRITES the first!

@mesh.agent(name="llm-providers", auto_run=True)
class LLMProviders: pass
```

**Result:** Only `openai_provider` is registered. `claude_provider` is lost.

## Root Cause

In `mesh/helpers.py` (line 145), the decorator hardcodes the function name:

```python
def process_chat(request: MeshLlmRequest) -> dict[str, Any]:
    """Auto-generated LLM handler."""
    ...
```

FastMCP's `@app.tool()` uses function name as the tool name, so all providers get the same tool name and overwrite each other.

## Why This Matters

- **Use case:** Single agent providing multiple LLM models (Claude, GPT-4, Gemini)
- **Mesh resolution works fine** - `@mesh.llm` finds providers by capability + tags, not function name
- **FastMCP registration fails** - requires unique function names within the server

## Proposed Solution: Use Original Function Name

Preserve the user's chosen function name instead of hardcoding `process_chat`:

```python
# mesh/helpers.py - current implementation
def process_chat(request: MeshLlmRequest) -> dict[str, Any]:
    ...

# Proposed change
def llm_handler(request: MeshLlmRequest) -> dict[str, Any]:
    ...

# Set __name__ to match original function
llm_handler.__name__ = func.__name__
llm_handler.__qualname__ = func.__qualname__
```

### Why This Approach?

1. **Users already choose unique names** - `claude_provider`, `openai_provider`, etc.
2. **Descriptive** - Easy to identify which provider when debugging
3. **Consistent** - Aligns with how other decorators preserve function identity
4. **Minimal change** - Just add two lines to set `__name__` and `__qualname__`

### Expected Behavior After Fix

```python
[...]

PR #1446, opened 2026-07-31
fix: boot-fail on duplicate advertised MCP tool names; drop phantom base-path route

dhyansraj:
## Summary

Two issues from the #1441 review round, bundled because both are about a name derived from a method or annotation.

### #1442 — duplicate advertised MCP tool name

Two tool declarations advertising the same MCP tool name made one silently disappear, in **all three runtimes**:

| Runtime | Behaviour today |
|---|---|
| Java | SDK 2.0.0 replaces silently (`"Replace existing Tool with name '{}'"`) |
| Python | `decorator_registry.py:151` bare dict assign — the loser never reaches the heartbeat |
| TypeScript | fastmcp's `addTool` filters out the previous entry, no log |

All three now **fail at startup** on a genuine collision, matching the precedent `MeshToolRegistry` already set for duplicate *capability* ("Capability names must be unique within an agent").

**Nothing is renamed.** The advertised MCP name *is* the wire name — `tools/call` sends `params.name = function_name` in every runtime (`resolver.go:266` → `McpHttpClient.java:525`, `proxy.ts:383`, `unified_mcp_proxy.py:1203`); capability rides along only for logging and cache keys. Renaming would have broken every mesh DI call into the provider, so this is a guard only.

### #1443 — phantom route at the controller base path

Every `@MeshRoute` handler also registered a route at the controller base path: `getMappingInfo` consulted `@RequestMapping` unconditionally, and `AnnotationUtils.findAnnotation` finds a verb shortcut's meta-`@RequestMapping` **without applying `@AliasFor`** — so it returned empty paths and defaulted to `{""}` + GET. Lookups moved to `AnnotatedElementUtils.findMergedAnnotation`, and the merged lookup is now a last resort. `getClassBasePath` had the identical defect one level up and is migrated too.

## Review notes

**The guards discriminate on declaration origin, not on "have I seen this name."** Per the #1445 lesson, each tolerates the *same* declaration registered again (module re-import, repeated bean post-processing) and throws only for two *different* declarations. The discriminators are deliberate and were each validated against a wrong alternative:

- **Python** stamps `_mesh_declaration_origin` in `helpers.py` pointing at the user's function. Without it, every `@mesh.llm_provider` tool shares one code-object identity — `helpers.py` renames a nested `process_chat` for all of them — and the guard would have been **inert across the entire provider surface**, unable to protect the bug (#227) it was built for.
- **TypeScript** derives provider identity from the config, not `execute` — every provider is built by one factory, so their closures share source text and would have compared equal. The same trap.
- **Java** keys on funcId, matching `MeshToolWrapperRegistry.indexBareMethodName`.

**A dual-module filename disagreement broke the guard.** `python main.py` reports `co_filename` as `main.py` while a re-import reports `/agent/main.py`. Identities diverged and the guard raised *over* `dual_module_check`'s much better diagnostic. Fixed with `_is_dual_module_pair`, mirroring `detect_dual_module_registration`'s conservative rule (exactly one side `__main__`, same qualname) — deliberately not a blanket same-qualname pass, pinned by `test_two_ordinary_modules_are_still_a_collision`.

## Behavior changes — agents that boot today and now will not

- **Two `@MeshLlmProvider` classes on one Java agent.** Both advertise `llm_generate` with different funcIds. They already collide today — one is dead on the wire — so failing loudly is the correction, not a regression. `MeshLlmProviderProcessor` builds one wrapper per provider, so multi-provider agents look supported by construction; the error message special-cases the `llm_provider:` prefix rather than saying "rename one of the methods", which is impossible for a framework constant.
- **A user tool named `__mesh_job_status`/`_result`/`_cancel`** now wins; the framework helper is skipped on both surfaces instead of clobbering it on fastmcp while the heartbeat kept advertising the user's.

## Known limit (unchanged by this PR)

Two `@MeshTool` **overloads in one class** are still not caught: they share a funcId (`FQCN.methodName`, no parameter types), so the second evicts the first from the funcId-keyed map before the guard iterates. Widening funcId is off the table — it is the identity the core echoes on `funcId:dep_N`. The javadoc states this limit rather than implying full coverage. This is the tool-side twin of what #1441 fixed for routes and is worth deciding on separately.

`jobs-helper-tools.ts` shows ~155 changed lines but only **21 are insertions** — the rest is reindentation from wrapping each registration in a skip guard.

## Test plan

- [x] Java 1153 passed, 0 failures (full reactor)
- [x] Python 1551 passed, 1 skipped
- [x] TypeScript 1264 passed, 88 files; `typecheck` clean
- [x] Every new test verified red with its own fix reverted, one at a time, in a scratch copy
- [x] Two wrong-fix experiments run for #1443 — deleting the `@RequestMapping` branch, and collapsing the five shortcuts into one merged lookup — each confirmed to break a different test
- [x] Python discriminator validated against two wrong alternatives (`id(func)`, and `__module__`+`__qualname__`); both go red on real paths
- [x] Three tests that could not fail were rewritten or justified, not kept as-is
- [ ] CI green

A zero-context review agent reviewed the first draft and found 1 blocker + 10 warnings; all ten were verified real and fixed. The blocker was TS `addLlmProvider` bypassing the guard entirely, which would have shipped an "all three runtimes" claim with one path still silent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented different tools from registering with the same advertised MCP name, with clearer conflict details and remediation guidance.
  * Preserved safe re-registration of the same tool across reloads, wrappers, and supported runtime modes.
  * Prevented framework job helpers from overwriting user-defined tools; conflicting helpers are skipped.
  * Improved Spring route discovery for composed annotations, aliases, multiple HTTP verbs, and pathless mappings.
* **Tests**
  * Added comprehensive coverage for duplicate tools, helper collisions, and Spring route scanning.

### DIFF
--- a/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshMcpServerConfiguration.java
+++ b/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshMcpServerConfiguration.java
@@ -68,6 +68,9 @@ public class MeshMcpServerConfiguration {
     private static final Map<String, Object> WRAP_RESULT_META =
         Map.of("fastmcp", Map.of("wrap_result", true));
 
+    /** funcId prefix {@code LlmProviderToolWrapper} builds ({@code llm_provider:<capability>}). */
+    private static final String LLM_PROVIDER_FUNC_ID_PREFIX = "llm_provider:";
+
     // MCP SDK 2.0.0 uses Jackson 3 (tools.jackson)
     private final tools.jackson.databind.json.JsonMapper mcpJsonMapper;
 
@@ -216,9 +219,20 @@ public class MeshMcpServerConfiguration {
                 continue;
             }
             String funcId = handler.getFuncId();
-            String previousFuncId = claimedBy.putIfAbsent(toolName, funcId);
-            if (previousFuncId == null || Objects.equals(previousFuncId, funcId)) {
-                // Unclaimed, or the same declaration re-registered — both fine.
+            // containsKey, NOT putIfAbsent: the map permits null values, and
+            // putIfAbsent REPLACES a null-valued mapping and returns null — so a
+            // name first claimed by a handler with a null funcId would read as
+            // unclaimed for every later claimant, and the collision this guard
+            // exists to catch would walk straight through. Nothing enforces a
+            // non-null funcId: McpToolHandler is a public interface and
+            // getFuncId() has no default.
+            if (!claimedBy.containsKey(toolName)) {
+                claimedBy.put(toolName, funcId);
+                continue;
+            }
+            String previousFuncId = claimedBy.get(toolName);
+            if (Objects.equals(previousFuncId, funcId)) {
+                // The same declaration re-registered — fine.
                 continue;
             }
             throw new IllegalStateException(String.format(
@@ -255,9 +269,6 @@ public class MeshMcpServerConfiguration {
         }
         return "Rename one of the methods, or move one of the tools to a separate agent.";
     }
-
-    /** funcId prefix {@code LlmProviderToolWrapper} builds ({@code llm_provider:<capability>}). */
-    private static final String LLM_PROVIDER_FUNC_ID_PREFIX = "llm_provider:";
 
     /**
      * Register a McpToolHandler as an MCP tool specification.
--- a/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshMcpServerConfiguration.java
+++ b/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshMcpServerConfiguration.java
@@ -23,9 +23,11 @@ import org.springframework.boot.web.servlet.ServletRegistrationBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
+import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * MCP Server configuration for MCP Mesh agents.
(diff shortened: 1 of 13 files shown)


dhyansraj:
Review round 2 — all five findings verified against current code and fixed. None skipped.

**One turned out to be a live bug rather than a nitpick.** `registerJobHelperTools` read `toolDeclarations` to find user-claimed helper names but never recorded the helpers it registered, making the guard one-directional. `_autoStart` is `process.nextTick`-scheduled and awaits `resolveStartupBindPort`, so a synchronous `addTool` after construction is safe, but anything after an `await` — `const tools = await loadTools(); tools.forEach(t => agent.addTool(t))`, or any async init — lands *after* the helpers, finds no entry, passes the guard, and clobbers the helper on fastmcp. `addTool` is public and callable at any time.

**The `putIfAbsent` finding is not reachable today, and was still worth fixing.** No in-tree `getFuncId()` returns null. But `HashMap.putIfAbsent` replaces a null-valued mapping and returns null exactly as it does for an absent one, so a name first claimed with a null funcId reads as unclaimed for every later claimant — the guard silently passes. `McpToolHandler` is a public interface with no default and nothing enforcing non-null, so the old code rested on an unenforced invariant. `containsKey` now decides.

**The Python constant fix prevents a repeat of this PR's main bug.** `helpers.py` stamped the literal `"_mesh_declaration_origin"` while `decorator_registry.py` defines `DECLARATION_ORIGIN_ATTR` for it. Demonstrated rather than asserted: with the literal in place, renaming the constant silently took the real `mesh.llm_provider` path back to inert — `test_real_llm_provider_decorator_rejects_two_providers_named_the_same` went red. Post-fix the same rename leaves it green.

The test file keeps its literals deliberately. A test importing the constant would follow a rename silently and stay green, leaving nothing to notice that a cross-module attribute name had changed. The rename simulation confirms the pin fires.

**Test counts:** Java 1153 → 1156, TypeScript 1264 → 1268, Python 1551 (unchanged — #4 is a literal→constant refactor with identical behavior, so no new test could go red; it was validated by the rename simulation above instead).

Each new test verified red without its fix. For the null-funcId case, two plausible *alternative* fixes were also run — skip null funcIds, and reject them outright — each turning a different one of the three tests red, confirming none is dead weight.

**Still open and unchanged by this PR:** two `@MeshTool` overloads in one class share a funcId and evict each other from the funcId-keyed handler map before this guard runs. Widening funcId is off the table — it is the identity the core echoes on `funcId:dep_N`. Noted in the javadoc rather than papered over.

Issue #1437, opened 2026-07-30
MeshRouteRegistry keys handlers by ClassName.methodName, so overloaded @MeshRoute handlers silently share one dependency list


dhyansraj:
## Summary

`MeshRouteRegistry.routesByHandler` is keyed by `"ClassName.methodName"` — **no parameter types** — and written with a bare `put`. Two overloaded `@MeshRoute` handlers in one controller therefore collide, and the second registration silently replaces the first.

`MeshRouteHandlerInterceptor.preHandle` rebuilds that same id from the `HandlerMethod` and uses the resulting metadata to build the injected dependency list. So **both overloads receive the last-registered overload's dependencies**, positionally.

## Evidence

`helm`-style probe against the real registry (run and deleted, not committed):

```
PROBE routesByPath GET:/a -> alpha
PROBE routesByPath GET:/b -> beta
PROBE getByHandlerMethodId(com.example.Ctrl.handle) -> beta
      <-- what MeshRouteHandlerInterceptor.preHandle uses for BOTH overloads
```

Two `@MeshRoute` methods named `handle` on different mappings both resolve `beta`'s dependency list. `/a` receives `beta`'s proxies.

## Where it is

- `MeshRouteRegistry.java:39` — `private final Map<String, RouteMetadata> routesByHandler = new ConcurrentHashMap<>();`
- `MeshRouteRegistry.java:51` — bare `routesByHandler.put(metadata.getHandlerMethodId(), metadata);` — **last-write-wins, no collision guard**
- `MeshRouteRegistry.java:87` — `getByHandlerMethodId`, public
- `MeshRouteHandlerInterceptor.java:112` — rebuilds the id and looks it up
- id construction: `MeshRouteBeanPostProcessor.java:82` — `targetClass.getName() + "." + method.getName()`

Note the asymmetry: `MeshA2ARegistry.register` (`:65`) guards its equivalent with `putIfAbsent` and fails at boot on a collision. The route registry does not.

## Why it is worse on the route path

The A2A instances of this defect (fixed in PR #1436) leaked a cached plan or submitter. Here the interceptor builds the **injected dependency list** from the colliding metadata, so a request to `/a` gets `/b`'s proxies in `/a`'s slots — silent, and now positional, so nothing about the shape of the result signals the swap.

## Discovered by

[...]

PR #1449, opened 2026-07-31
fix(java): reject two @MeshTool overloads in one class instead of evicting one

### LINKED ISSUE #1448: Two @MeshTool overloads in one class silently evict each other and can misroute dependencies
Follow-up to #1442, which closed the cross-class case. The intra-class case is still silent, and it is worse than "one tool goes missing".

## Reproduction

Two `@MeshTool` overloads in one class with **different** capabilities:

```java
@MeshTool(capability = "a") String analyze(String q) { ... }
@MeshTool(capability = "b") String analyze(int n)    { ... }
```

funcId is `targetClass.getName() + "." + method.getName()` with no parameter types (`MeshToolBeanPostProcessor.java:322`), so both compute `Foo.analyze`.

## End state — no error, no warning

| Stage | Result |
|---|---|
| `MethodIntrospector.selectMethods` | both scanned, both registered |
| `MeshToolRegistry.registerTool` duplicate guard | **does not fire** — keyed by capability, and `"a" != "b"` |
| `wrappers` / `handlers` | **second evicts first** (bare `put` on the shared funcId) |
| `wrappersByMethodName` | funcId-equality tolerance at `:189` reads this as idempotent re-registration — plain `put`, no ambiguity flag |
| heartbeat | **two** tool entries, capabilities `a` and `b`, **both with `function_name = "analyze"`** |
| MCP server | **one** tool. Capability `a` is advertised to the registry with no MCP tool behind it |

The only trace is `Registered wrapper: Foo.analyze` logged twice at INFO with identical text.

## Dependencies can be misrouted, not just dropped

Because both advertise `function_name = "analyze"`, the Go registry concatenates **both tools' dependency lists under that one key** (`ent_service.go:886`). `dep_index` then becomes a position in the merged list. Java resolves `"analyze"` to the surviving wrapper only, so out-of-range indices are dropped by the bounds check — but in-range ones are not. **The surviving overload can be handed the other overload's dependency in slot 0.**

## No funcId scheme can fix this

Three independent single-valued namespaces key on the bare method name:

1. the MCP SDK tool table (SDK 2.0.0 silently replaces),
2. the registry's `dependencies_resolved` map key,
3. Java's own `wrappersByMethodName` inbound resolution index.

`@MeshTool` has no `name`/`toolName` attribute — unlike TypeScript's `addTool`, which offers an explicit `name` escape hatch — so there is no way to give two overloads distinct wire names. Widening funcId to include parameter types would make both wrappers locally addressable while both still collapse onto one wire name: the registry would keep advertising capability `a`, and a consumer resolving it would be routed into `analyze(int)`. That is strictly worse than today's eviction.

It would also break `MeshEventProcessor.java:385-388`, which derives the wire-facing short name as `funcId.substring(funcId.lastIndexOf('.') + 1)` — for `Foo.analyze(java.lang.String)` that yields `"String)"`.

**Correcting the record:** `MeshMcpServerConfiguration.java:204-206` currently claims widening is impossible because funcId "is the wire identity the Rust core echoes back on `funcId:dep_N`". That is wrong and should be removed. No funcId crosses the FFI boundary — the core sends the bare method name (`MeshToolRegistry.java:252` → `resolver.go:39` → `runtime.rs:709` → `MeshEventProcessor.java:194`). funcId is process-local. The reason not to widen it is the wire-name collision above, not an FFI constraint.

## Proposed fix — reject at boot

Guard in `MeshToolWrapperRegistry.registerWrapper`, immediately before `wrappers.put(funcId, wrapper)` (`:116`). That is the exact eviction site and the only point that sees both declarations.

Discriminate on `wrapper.getMethod()`, not instance identity (`Method.equals` is value-based and includes parameter types), mirroring the tolerance already at `:189`:

- same funcId + **equal** `Method` → prototype scope / context refresh / repeated post-processing → replace, as today;
- same funcId + **different** `Method` → genuine overload → `IllegalStateException` naming both full signatures.

`getMethod()` can be null for non-`@MeshTool` handlers (`:117`), so guard on that.

## Consistency note

Python and TypeScript already reject this — #1446 put their guards at *declaration* time, before any store (`decorator_registry.py:243`, `agent.ts:490`). Java is the outlier only in guard placement: its check sits at MCP-server-build time, after the eviction has already happened. This moves Java's earliest check to the store site.

## Acceptance

- Two `@MeshTool` overloads in one class fail the boot with a message naming both signatures.
- Re-registering the same `Method` still replaces cleanly (no regression of #1445).
- The incorrect FFI claim is removed from `MeshMcpServerConfiguration`'s javadoc, along with the "known limit" paragraph it justifies.

dhyansraj:
## Summary

Completes #1442. That closed the cross-class case; two `@MeshTool` overloads in **one** class stayed silent — and the failure is worse than a dropped tool.

funcId is `FQCN + "." + methodName` with no parameter types (`MeshToolBeanPostProcessor.java:322`), so both overloads compute `Foo.analyze` and the second evicts the first from `wrappers` and `handlers`. `MeshToolRegistry`'s duplicate guard doesn't fire (it's keyed by capability), and `indexBareMethodName`'s funcId-equality tolerance reads the pair as an idempotent re-registration. Net result, with no error or warning: **two capabilities advertised to the registry, one MCP tool behind them.** Reproduced with a probe rather than argued:

```
handlers(served)=1  capabilities=[analyze_string, analyze_int]
  capability analyze_string served=false     <-- advertised, nothing behind it
  capability analyze_int    served=true
```

**Dependencies can be misrouted, not just dropped.** Both advertise `function_name = "analyze"`, so the registry concatenates *both* tools' dependency lists under that single key (`ent_service.go:886`) and `dep_index` becomes a position in the merged list. Out-of-range indices are dropped by the bounds check; in-range ones are not. The surviving overload can be handed the other one's dependency in slot 0.

## Why not widen funcId

Three single-valued namespaces key on the bare method name — the MCP SDK tool table, the registry's `dependencies_resolved` map, and Java's own `wrappersByMethodName`. `@MeshTool` has no `name`/`toolName` attribute (unlike TypeScript's `addTool`), so there is no way to give two overloads distinct wire names.

Widening funcId would make both wrappers *locally* addressable while both still collapse onto one wire name: capability `a` would stay advertised, and a consumer resolving it would be routed into `analyze(int)`. Strictly worse than eviction. It would also break `MeshEventProcessor.java:385-388`, which derives the short name as `funcId.substring(funcId.lastIndexOf('.') + 1)` — for `Foo.analyze(java.lang.String)` that yields `"String)"`.

## Correcting a wrong claim from #1446

`MeshMcpServerConfiguration`'s javadoc claimed widening funcId was impossible because it "is the wire identity the Rust core echoes back on `funcId:dep_N`". **That is false and is removed.** No funcId crosses the FFI boundary. The core sends the bare method name:

`MeshToolRegistry.java:252` (`setFunctionName(meta.method().getName())` — the only `setFunctionName` call in the runtime) → `resolver.go:39` → `ent_service.go:886` → `runtime.rs:709` → `MeshEventProcessor.java:194`.

funcId is process-local. Each link was re-verified against the source for this PR rather than carried over. The real reason not to widen is the wire-name collision above.

Corroborating detail: `resolveWrapper` tries `wrappers.get(funcId)` and then falls back to the bare-name index — that fallback exists *because* the core sends bare names, so on the Java dependency path the funcId lookup always misses and the fallback is the production path.

## The guard

Sits in `MeshToolWrapperRegistry.registerWrapper` immediately before `wrappers.put(funcId, wrapper)` — the exact eviction site, and the only point that sees both declarations. Discriminates on `Method`, not instance identity (`Method.equals` is value-based and includes parameter types):

- same funcId + **equal** `Method` → prototype scope, context refresh, repeated post-processing → replace as today (#1445 lesson preserved)
- same funcId + **different** `Method` → throw, naming both full signatures
- null `Method` on either side → fall through; non-`@MeshTool` handlers have none

This also brings Java in line with Python and TypeScript, which #1446 guarded at declaration time. Java's only check was at MCP-server-build time — after the eviction had already happened.

## Test plan

- [x] Full Java reactor: **1164 tests, 0 failures** (15 + 117 + 821 + 211)
- [x] 3 of 5 new tests verified red with the guard removed in a scratch copy
- [x] Pre-fix end state reproduced with a probe, not assumed
- [x] All 258 `@MeshTool` annotations in the repo scanned for existing same-name-same-class pairs — **zero found**, nothing existing trips the guard
- [x] `MeshToolInheritanceScanTest` still 11/11: `MethodIntrospector.selectMethods` collapses override pairs to one logical method, so inheritance yields one `registerWrapper` call, not two
- [ ] CI green

**Two tests deliberately pass both with and without the fix.** `sameMethodFreshWrapperReplaces` and `nullMethodDoesNotThrow` assert the guard does *not* fire, so they cannot be red-without-fix by construction. They fence a wrong fix — instance-identity discrimination, or an unguarded null deref — rather than pinning this one. Dropping them would delete the #1445 protection.

## Known, not changed

The guard is check-then-act on a `ConcurrentHashMap` rather than atomic. Registration is Spring bean post-processing, which is serial at boot, and `indexBareMethodName` directly below has the same shape; consistency with the neighbour was preferred over wrapping the put in `compute()`.

Closes #1448

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented conflicting overloaded MCP tools from sharing the same function ID.
  * Added clear error messages identifying duplicate tool registrations.
  * Preserved registry consistency when duplicate registrations are rejected.
  * Allowed safe re-registration of the same method and entries without method metadata.

* **Documentation**
  * Clarified duplicate tool-name handling across registries and MCP tool listings.

### DIFF
--- a/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshMcpServerConfiguration.java
+++ b/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshMcpServerConfiguration.java
@@ -196,17 +196,18 @@ public class MeshMcpServerConfiguration {
      *       so a repeated registrar pass replaces rather than accumulates.</li>
      * </ul>
      *
-     * <p><b>Known limit — overloads in ONE class are not reached.</b> This scans
+     * <p><b>Overloads in ONE class never reach this guard — they are rejected
+     * earlier</b> (issue #1448). This scans
      * {@link MeshToolWrapperRegistry#getAllHandlers()}, which is keyed by funcId,
-     * and a funcId is {@code FQCN.methodName} with no parameter types. Two
-     * overloaded {@code @MeshTool} methods in the same class therefore share a
-     * funcId and the second EVICTS the first from that map before this guard
-     * ever sees it — the collision is real but invisible here. Widening the
-     * funcId is not an option: it is the wire identity the Rust core echoes back
-     * on {@code funcId:dep_N} dependency-resolution events. The narrower
-     * same-class overload case is covered on the registry side by
-     * {@link MeshToolRegistry#registerTool}'s duplicate-capability guard
-     * whenever the overloads also share a capability.
+     * and a funcId is {@code FQCN.methodName} with no parameter types, so two
+     * overloaded {@code @MeshTool} methods in the same class share one and the
+     * second would evict the first from that map before this guard ever sees it.
+     * {@link MeshToolWrapperRegistry#registerWrapper} now refuses that at the
+     * store site. Widening the funcId would not have helped: both overloads
+     * still collapse onto ONE wire name — the bare method name, which keys the
+     * MCP tool table, the registry's {@code dependencies_resolved} map and
+     * {@code wrappersByMethodName} — and {@code @MeshTool} has no name attribute
+     * to separate them.
      *
      * @param handlers every handler about to be registered with the MCP server
      * @throws IllegalStateException naming both colliding declarations
--- a/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshToolWrapperRegistry.java
+++ b/src/runtime/java/mcp-mesh-spring-boot-starter/src/main/java/io/mcpmesh/spring/MeshToolWrapperRegistry.java
@@ -112,6 +112,11 @@ public class MeshToolWrapperRegistry {
         String capability = wrapper.getCapability();
         String methodName = wrapper.getMethodName();
 
+        // Two DIFFERENT declarations must never share a funcId (issue #1448).
+        // Checked HERE, immediately before the put that would evict the first
+        // one — this is the only point that sees both declarations.
+        assertNotAnOverload(funcId, wrapper);
+
         // Store in wrapper maps (for dependency updates)
         wrappers.put(funcId, wrapper);
         if (wrapper.getMethod() != null) {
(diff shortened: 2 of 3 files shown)


dhyansraj:
Review round — one finding, verified valid and fixed. Nothing skipped.

**`handlersByCapability` was the one index that did not honour a replacement.** `registerWrapper` stores into five indices; a same-funcId wrapper with a different capability updated the funcId-keyed maps but left the old capability key pointing at the evicted wrapper. `getHandlerByCapability` (→ `ToolInvoker:162` local invocation, `MeshEventProcessor:303`, `:467`), `containsCapability` and `getAllHandlersByCapability` would all serve or report it.

**Reachability, stated plainly:** this is **not reachable in production today**, and the javadoc says so rather than implying a live bug. `registerWrapper` has one production caller and one construction site (`MeshToolBeanPostProcessor:368`/`:338`), always with a scanned `Method`; `assertNotAnOverload` now rejects a different `Method` under the same funcId; and an equal `Method` carries the same `@MeshTool` annotation, so its capability cannot change. The only opening is a null-`Method` wrapper, which production never builds.

Fixed anyway because "N indices replace, one does not" is precisely the asymmetry behind #1437, #1445 and #1448. It stops being theoretical the moment a second construction site appears.

**The trap in the obvious fix.** A blind `handlersByCapability.remove(previousCapability)` would clobber a legitimate, unrelated wrapper that currently owns that capability. The fix uses two-arg `ConcurrentHashMap.remove(key, value)`; `MeshToolWrapper` does not override `equals`, so the match is identity.

**Verified against two wrong implementations, not one:**

| Scratch variant | Test that goes red |
|---|---|
| release call removed | `changedCapabilityReleasesTheOldKey` — the replaced wrapper still answers under the capability it no longer declares |
| identity condition dropped (`remove(key)`) | `changedCapabilityDoesNotClobberAnUnrelatedOwner` — the capability's current owner was wrongly evicted |

Each discriminating test is red against exactly one variant and green against the other, so neither is a tautology.

**One test cannot be red, and is kept deliberately.** `sameCapabilityReplacementStillReplaces` asserts the release does *not* fire on an ordinary same-capability replacement, so it passes in all three variants. Same category as `sameMethodFreshWrapperReplaces` and `nullMethodDoesNotThrow`: it fences a future over-broad release rather than pinning the current fix.

Full Java reactor: **1167 tests, 0 failures** (was 1164; +3 new). `MeshMcpServerDuplicateToolNameTest` 11/11, `MeshToolWrapperRegistryMethodNameCollisionTest` 5/5, `MeshToolWrapperRegistryOverloadTest` 8/8, `MeshToolInheritanceScanTest` 11/11, `MeshToolRegistryDuplicateCapabilityTest` 2/2.

Issue #1334, opened 2026-07-16
fix(llm): OpenAI gpt-5 reasoning models 400 with function tools (native path is chat.completions-only)


dhyansraj:
## Summary

mesh's native OpenAI handler calls **only** `/v1/chat/completions`. OpenAI gpt-5-family **reasoning** models reason by default, and OpenAI rejects **reasoning + function tools** on `/v1/chat/completions`:

> Function tools with reasoning_effort are not supported for `gpt-5.6-terra` in `/v1/chat/completions`. To use function tools, use `/v1/responses` or set `reasoning_effort` to `'none'`.

So any gpt-5 reasoning model used **with tools** — the common `@mesh.llm` agentic pattern — fails with HTTP 400 on the first LLM turn.

## Reproduction (verified end-to-end, local)

- `examples/toolcalls/` triad: `analyst-py` (consumer, `@mesh.llm` with a weather tool filter) → `openai-provider-py` (provider) → `weather-tool-py`.
- Consumer model override `MESH_LLM_MODEL=openai/gpt-5.6-terra`; `OPENAI_API_KEY` set.
- Calling `analyst-py`'s `analyze` tool → **HTTP 400** on the first turn, before any tool call.

Confirmed during repro:
- The consumer→provider model override propagates correctly (`openai/gpt-5.6-terra` is the model on the wire, not the provider's declared `gpt-4o`).
- The gpt-5-family **sampling-param** handling is fine — no error about `temperature`/`top_p`/`max_tokens`. The *only* failure is reasoning + tools on chat.completions.

## Root cause

`src/runtime/python/_mcp_mesh/engine/native_clients/openai_native.py` uses `client.chat.completions.create` exclusively (~lines 750, 803); there is no `/v1/responses` path. mesh does not send `reasoning_effort` (passthrough-only, ~line 431), so the model's **default** reasoning plus attached tools trips the unsupported combination.

## Fix (proper)

Route gpt-5 **reasoning** models to the OpenAI **Responses API** (`/v1/responses`) so reasoning + tools work, keeping `gpt-4o` and gpt-5 **chat** variants (`gpt-5*-chat*`) on `chat.completions`. Setting `reasoning_effort='none'` is only a fallback — it disables the model's reasoning, which is unacceptable for a reasoning-tier model.

## Cross-runtime

[...]

PR #1453, opened 2026-07-31
feat(llm): native Responses streaming and image input for the OpenAI path

### LINKED ISSUE #1336: feat(llm): OpenAI native Responses path — native streaming + multimodal input
## Summary

Completeness follow-ups to the Python OpenAI **Responses API** path added in #1334 (which routes gpt-5/o-series reasoning models with tools to `client.responses.create`). Two capabilities were intentionally deferred and are marked in code:

### 1. Native Responses streaming
The reasoning+tools **streaming** case currently runs the Responses call **buffered** and emits a single terminal chunk (`# TODO(#1334): native Responses streaming` in `openai_native.py`). Implement native semantic-event streaming — map `response.output_text.delta`, `response.function_call_arguments.delta`, and `response.completed` (usage) to mesh's `_StreamChunk` shape, with the interrupted-stream fallback-usage handling the chat path has.

### 2. Multimodal input translation
Chat-style content-part arrays are translated for **text** parts (`{type:text}` → `input_text`/`output_text`), but **image** parts currently raise a clear `ValueError` on the Responses path (rather than a cryptic OpenAI 400). Add the `{type:image_url}` → Responses `input_image` translation so reasoning + tools + vision works.

## Acceptance

- [ ] Native streaming for reasoning+tools via Responses (no buffered fallback); live smoke with a streaming consumer.
- [ ] Image content translated to `input_image`; live smoke with a vision + tools reasoning request.

Relates to #1334.

dhyansraj:
## Summary

Completes the two capabilities #1334 deferred on the OpenAI **Responses** path (gpt-5/o-series reasoning models with tools).

**1. Native Responses streaming.** The reasoning+tools case ran the Responses call *buffered* and emitted a single terminal chunk. It now maps the semantic event stream onto `_StreamChunk`:

| Event | → `_StreamChunk` |
|---|---|
| `response.output_text.delta` | `delta.content` |
| `response.output_item.added` (`function_call`) | `delta.tool_calls` with `call_id` + `name` |
| `response.function_call_arguments.delta` | argument fragment at the same index |
| `response.completed` / `.incomplete` / `.failed` | usage + `finish_reason` |

**2. Multimodal input.** `{type:image_url}` now translates to Responses `input_image`; previously it raised a deliberate `ValueError`.

## Two shapes the SDK types forced

Both read from the installed `openai` 2.14.0 types, not from memory — and both are places the obvious guess is wrong:

- **`ResponseFunctionCallArgumentsDeltaEvent` carries only `delta` + `item_id`** — no `call_id`, no `name`. Those arrive once, on `response.output_item.added`. Without handling that event the merger drops **every** tool call, because it discards id-less slots. Item ids are re-keyed to contiguous 0-based ordinals so `output_index` (which reasoning items also occupy) can't leak into the merged shape.
- **`response.output_item.done` repeats the FULL arguments string.** Since `_merge_streamed_tool_calls` concatenates, it is used strictly as a gated backfill (`id_emitted`/`args_emitted`), never a re-emit.
- **`ResponseInputImageParam.image_url` is a flat `Optional[str]`** — https URLs and `data:` URIs share one field, with no Anthropic-style source split — and `detail` is `Required`, so one is always emitted (`auto` default; invalid values soft-fail to `auto` with a WARN).

Interrupted-stream handling mirrors the chat path exactly: counters recorded before the yield, `final_usage_emitted` flipped only after it returns, same `finally`, same guard. `stream_options` is deliberately **not** forwarded — Responses' `StreamOptions` has no `include_usage`; usage always rides on `response.completed`.

## Live-verified, and the live run corrected two mistakes

4 gated integration tests pass against **three** reasoning models — `gpt-5.6-terra`, `gpt-5-nano`, and `o4-mini` (o-series included, so the routing isn't gpt-5-specific). Double-gated on `MCP_MESH_LIVE_INTEGRATION=1` **and** `OPENAI_API_KEY`, matching `TestLiveRefusalIntegration`. Verified three ways that neither gate alone spends:

```
no key + no gate       → 13 passed, 5 skipped
gate ON, no key        → 4 skipped, no errors
key present, gate OFF  → 4 skipped        ← no accidental spend
```

Two things a mocked-only pass would have shipped broken:

1. **A colour-word vision assertion that fails on terra.** `gpt-5.6-terra` describes a pure `#FF0000` image as *"Orange"*; `gpt-5-nano` says *"Red"*. The PNG was decoded bytewise to rule out a generation bug — filter bytes all 0, unique pixels `{(255,0,0)}` — so it is genuine model colour perception. Asserting a colour word is exactly the exact-model-text assertion this repo forbids. Replaced with a **structural** discriminator: identical prompt with and without the image, asserting the input-token delta. Colour word is now informational, `skip` on mismatch.
2. **The token threshold was also wrong** — first live run gave `delta=5`. gpt-5-family bills images as **32×32 patches**, so an 8×8 image is ~4 patches. Fixed by generating at 512px: measured `with-image=445, text-only=58, delta=387`. Both facts are documented in the helper docstring.

## The buffered-blob discriminator actually discriminates

Two independent checks, so a short answer can't produce a false pass:
- `len(text_chunks) >= 2`;
- **token-count-independent**: no chunk may carry content *and* usage together. The removed fallback packed content + tool_calls + usage + finish_reason into one chunk, so it fails this regardless of answer length.

Skip-vs-fail is split deliberately and documented: **model behaviour** (won't call the tool, odd colour word) → `skip`; **mesh shape** (blob instead of deltas, missing usage, unmergeable call, image rejected) → `assert`. Both classes assert `_openai_wants_responses_api(...)` up front so they cannot silently degrade into chat-path tests.

## Known defect, deliberately NOT fixed here

`aclose()` on a streaming generator raises `RuntimeError: async generator ignored GeneratorExit`. The `finally` block's fallback `yield` **succeeds**, and the aclose machinery objects afterwards — so the existing `except (GeneratorExit, StopAsyncIteration)` never fires.

**Pre-existing and not introduced here**, confirmed by probing the untouched *chat* path directly, which raises identically. It spans `openai_native`, `anthropic_native` and `gemini_native`. Production impact is contained: `helpers.py` wraps `aclose()` in `except Exception` and debug-logs, so the only casualty is a fallback usage chunk that had nowhere to go.

Pinned by `test_consumer_abort_teardown_matches_chat_path`, which asserts both paths produce the *same* outcome — so a future fix has to move them together rather than diverging. Worth its own issue.

## Test plan

- [x] `python -m pytest tests/unit/ _mcp_mesh/` — **2515 passed, 7 skipped** (was 2515/3; +4 live skips)
- [x] `python -m pytest` — 1551 passed, 1 skipped (unchanged)
- [x] 29 of 31 new unit tests verified red against a scratch copy with `openai_native.py` restored from HEAD
- [x] Live: 4 passed × 3 models
- [x] Gating verified in all three key/gate combinations
- [x] `ruff check` — zero new findings
- [ ] CI green

**Two unit tests cannot fail and are named rather than quietly kept:** `test_unsupported_part_type_raises_value_error` is a preservation test (the old code raised for all non-text parts, so it was already green; it goes red if the raise is removed), and `test_empty_text_delta_yields_no_chunk` is an edge-case guard, not a regression detector.

## Deviation worth noting

Usage counters are harvested from **any** lifecycle event carrying `response.usage`, not only the terminal one. OpenAI fills usage only at the end, so on the real API this changes nothing — it exists so the fallback branch is reachable for OpenAI-compatible servers that report early, and it matches what `anthropic_native` already does. Without it that branch would be decorative and untestable.

Closes #1336

### DIFF
--- a/src/runtime/python/_mcp_mesh/engine/native_clients/openai_native.py
+++ b/src/runtime/python/_mcp_mesh/engine/native_clients/openai_native.py
@@ -639,10 +639,70 @@ def _translate_response_format_to_text(response_format: Any) -> dict[str, Any]:
     return {"format": response_format}
 
 
-# ``ResponseInputImageParam.detail`` is ``Literal["low","high","auto"]`` and is
-# marked Required in the OpenAI SDK's own TypedDict — always emit one.
-_RESPONSES_IMAGE_DETAILS = frozenset({"low", "high", "auto"})
+# ``ResponseInputImageParam.detail`` is marked Required in the OpenAI SDK's own
+# TypedDict — always emit one.
 _RESPONSES_DEFAULT_IMAGE_DETAIL = "auto"
+# Long-stable trio, used only when SDK introspection is unavailable.
+_RESPONSES_IMAGE_DETAILS_FALLBACK = frozenset({"low", "high", "auto"})
+_RESPONSES_IMAGE_DETAILS_CACHE: frozenset[str] | None = None
+
+
+def _literal_strings(annotation: Any) -> frozenset[str]:
+    """Pull the string members out of a (possibly wrapped) ``Literal``.
+
+    Handles ``Required[Literal[...]]`` / ``Optional[Literal[...]]`` by
+    recursing into type args. Returns an empty set when there is no string
+    Literal to find.
+    """
+    import typing
+
+    args = typing.get_args(annotation)
+    if args and all(isinstance(a, str) for a in args):
+        return frozenset(args)
+    for arg in args:
+        found = _literal_strings(arg)
+        if found:
+            return found
+    return frozenset()
+
+
+def _valid_image_details() -> frozenset[str]:
+    """Accepted ``input_image.detail`` values for the INSTALLED openai SDK.
+
+    Read off ``ResponseInputImageParam``'s own ``Literal`` rather than
+    hard-coded. ``openai`` is an unpinned dependency and this enum GROWS:
+    2.14.0 had ``low``/``high``/``auto``; 2.52.0 added ``original``. A frozen
+    allowlist would silently downgrade a caller's valid ``detail="original"``
+    to ``auto`` — a wrong image fidelity, soft-failed with a misleading WARN.
+
+    Soft-fails to the long-stable trio if the SDK is absent or its layout
+    changes; this is a validator, it must never raise. Cached — the installed
+    SDK cannot change within a process.
+    """
+    global _RESPONSES_IMAGE_DETAILS_CACHE
+    if _RESPONSES_IMAGE_DETAILS_CACHE is not None:
+        return _RESPONSES_IMAGE_DETAILS_CACHE
+
+    details = _RESPONSES_IMAGE_DETAILS_FALLBACK
+    try:
+        import typing
+
+        from openai.types.responses import ResponseInputImageParam
+
+        hints = typing.get_type_hints(ResponseInputImageParam, include_extras=True)
+        found = _literal_strings(hints.get("detail"))
+        if found:
+            details = found
+    except Exception as exc:  # noqa: BLE001 — validator must not raise
+        logger.debug(
+            "Could not introspect ResponseInputImageParam.detail (%s); "
+            "falling back to %s",
+            exc,
+            sorted(_RESPONSES_IMAGE_DETAILS_FALLBACK),
+        )
+
+    _RESPONSES_IMAGE_DETAILS_CACHE = details
+    return details
 
 
 def _translate_image_part(part: dict[str, Any]) -> dict[str, Any]:
@@ -682,11 +742,13 @@ def _translate_image_part(part: dict[str, Any]) -> dict[str, Any]:
             f"got image_url={field!r}"
         )
 
-    if detail is not None and detail not in _RESPONSES_IMAGE_DETAILS:
+    valid_details = _valid_image_details()
+    if detail is not None and detail not in valid_details:
         logger.warning(
             "OpenAI Responses path: unsupported image detail %r "
-            "(expected one of low/high/auto); using %s",
+            "(this openai SDK accepts %s); using %s",
             detail,
+            "/".join(sorted(valid_details)),
             _RESPONSES_DEFAULT_IMAGE_DETAIL,
         )
         detail = None
--- a/src/runtime/python/_mcp_mesh/engine/native_clients/openai_native.py
+++ b/src/runtime/python/_mcp_mesh/engine/native_clients/openai_native.py
@@ -639,18 +639,84 @@ def _translate_response_format_to_text(response_format: Any) -> dict[str, Any]:
     return {"format": response_format}
 
 
+# ``ResponseInputImageParam.detail`` is ``Literal["low","high","auto"]`` and is
+# marked Required in the OpenAI SDK's own TypedDict — always emit one.
+_RESPONSES_IMAGE_DETAILS = frozenset({"low", "high", "auto"})
+_RESPONSES_DEFAULT_IMAGE_DETAIL = "auto"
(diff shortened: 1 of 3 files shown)


dhyansraj:
CI failure fixed — and chasing it found a **product** bug, not just a test one.

## Why CI failed and local didn't

Local `.venv` has **openai 2.14.0**; CI resolves **2.52.0** (`openai>=1.60`, unbounded). In 2.52.0 `InputTokensDetails` gained a required `cache_write_tokens`, and the `_sdk_response` fixture hand-wrote that nested dict. Reproduced both directions:

```
PRE-FIX  openai 2.14.0  →  151 passed        ← why it slipped through
PRE-FIX  openai 2.52.0  →  16 failed         ← identical to CI's traceback
```

All 16 funnelled through one fixture. Confirmed first that product code never constructs SDK models — the adapter only *reads* them via `_rget`, and `_responses_usage` reads exactly `input_tokens`/`output_tokens`, so a new nested counter is structurally incapable of affecting it.

The fixture now derives required nested counters from `model_fields` introspection while keeping `input_tokens`/`output_tokens`/`total_tokens` explicit and fully validated — the fields under test retain real required-ness; breakdowns the adapter never reads are absorbed. `model_construct` was rejected because it forfeits exactly the required-ness checking that motivated real SDK models, and would have hidden this.

## The product bug

```
detail ann  2.14.0: Required[Literal['low','high','auto']]
            2.52.0: Required[Literal['low','high','auto','original']]
```

`_RESPONSES_IMAGE_DETAILS` was hard-coded against an enum that grows. On the version users actually resolve, a caller passing the valid `detail="original"` was **silently downgraded to `"auto"`** — wrong image fidelity, plus a WARN claiming it was unsupported.

Mocked-only testing on 2.14.0 could not have caught this, and the live tests wouldn't either since they pass `"high"`. It now reads the allowlist off the installed SDK's Literal (cached, soft-failing to the stable trio if introspection fails — it's a validator, it must never raise), and the test parametrises over that Literal rather than a written-out list.

## Every other premise verified on 2.52.0

| Premise | Result |
|---|---|
| Event type literals (`output_text.delta`, `function_call_arguments.delta`, `output_item.added/done`, `completed`/`incomplete`/`failed`) | identical |
| `ResponseFunctionCallArgumentsDeltaEvent` carries only `delta`+`item_id` | identical — the `output_item.added` design premise holds |
| `ResponseInputImageParam` flat, `detail` Required | identical |
| Responses `StreamOptions` has no `include_usage` | still true, so not forwarding it remains correct |

Two 2.52.0 additions (`ResponseFunctionToolCall.caller`/`.namespace`, `ResponseInputImageParam.prompt_cache_breakpoint`) are optional — no action.

## Verification

Two throwaway venvs, same Python 3.11, only the openai version differing:

```
POST-FIX  openai 2.14.0  →  634 passed, 6 skipped
POST-FIX  openai 2.52.0  →  634 passed, 6 skipped
```

Live on **2.52.0**: 4 passed × `gpt-5.6-terra`, 4 passed × `o4-mini`.

Incidental result: those venvs also pulled **anthropic 0.120.2** and **google-genai 2.16.0** (repo `.venv` has 0.97.0 / 1.75.0) and all 634 adapter tests passed — so the sibling adapters are already clean against CI's resolution. `_sdk_response` was the only version-coupled fixture in the suite.

Suite: `python -m pytest tests/unit/ _mcp_mesh/` → **2516 passed, 7 skipped** (+1 new detail test).

## Related

The underlying drift problem is now tracked as #1454 — Python is the only mcp-mesh runtime without a lockfile, and `python-runtime.Dockerfile` installs by version alone, so two builds of one mesh version can ship different trees. Upper bounds on the vendor SDKs will land separately; they would not have caught this particular break, which came from a *minor*.

PR #1464, opened 2026-08-04
feat(python)!: remove litellm from the base install

### LINKED ISSUE #1383: Python: LiteLLM is not required for big-3 vendors but is unavoidable; make it an extra and fail loudly when missing

dhyansraj:
## Summary

Completes #1383. Items 1, 2, 4 and 5 landed in #1386/#1400; this is **item 3**: `pip install mcp-mesh` no longer pulls LiteLLM.

**110 → 101 dependencies, 311 MB → 189 MB of site-packages.** Long-tail vendors now need `mcp-mesh[litellm]` — an extra that has been valid since 3.3.2, so users have had two releases of runway to pin it ahead of the move.

The issue predicted ~92 packages. The real floor is **101**: that estimate credited litellm for several packages mesh declares directly. Nine removed (`litellm, tokenizers, tiktoken, regex, hf-xet, huggingface-hub, fsspec, filelock, fastuuid`), not eighteen.

## No undeclared dependency needed promoting — verified by construction

litellm's direct requires are `aiohttp, click, fastuuid, httpx, importlib-metadata, jinja2, jsonschema, openai, pydantic, python-dotenv, tiktoken, tokenizers`. Mesh already declares seven directly — the earlier `openai` and `jsonschema` promotions had done that work — and `python-dotenv`/`importlib-metadata` arrive via mcp/fastmcp. The litellm-exclusive residue appears in no `import` statement under `_mcp_mesh` or `mesh`.

Confirmed rather than reasoned: **all 150 modules** under both packages walked with `pkgutil.walk_packages` in a litellm-free venv — 150/150 imported, `litellm` never entering `sys.modules`.

## Two things that would have shipped broken

**49 tests would have turned CI red on merge.** They `patch("litellm.acompletion")`, which needs the real module importable, and CI installs `.[dev]`. Fixed by self-referencing `mcp-mesh[litellm]` from the `dev` extra rather than editing the workflow — keeps exactly one litellm pin per manifest, and `pip install -e '.[dev]'` was verified to resolve it.

**Every big-3 provider agent would have logged a startup WARNING.** `llm_provider` calls `litellm.get_llm_provider` at decoration time, and its `except` branch treated "litellm not installed" as a degradation — which is now the *default* state. Absence now logs at `debug`; present-but-unresolvable stays WARNING. The bare-name "using 'unknown' vendor" warning also moved below the big-3 inference block, where it previously contradicted itself one line later.

That second one is a behaviour change worth a look — it's the one part of this PR that isn't packaging.

## Live verification (litellm absent throughout)

| Model | Result |
|---|---|
| `anthropic/claude-sonnet-4-5` | 200, usage returned |
| `openai/gpt-4o-mini` | 200 |
| `gemini/gemini-2.5-flash` | 200 |
| `vertex_ai/gemini-2.5-flash` | 200 via native google-genai + ADC — confirms the long-tail-*looking* prefix stays native |
| streaming, all three | correct `_mesh_frame` chunk/end frames |
| `cohere/command-r`, `ollama/llama3` | friendly `ImportError` naming model, vendor and `pip install 'mcp-mesh[litellm]==…'` — not a bare `ModuleNotFoundError` |

## Layering verified for real

Installed the slimmed wheel (104 entries), then `pip install 'mcp-mesh[litellm]==3.4.0'` → `Requirement already satisfied: mcp-mesh`, installs only the 9 missing packages, returns to 113. **No base-version upgrade.** That is exactly what a scaffolded agent's `pip install -r requirements.txt` layer over `FROM mcpmesh/python-runtime` does.

`packaging/docker/python-runtime.Dockerfile` needs no change. No Dockerfile, compose file, chart or script anywhere assumes litellm is present.

## Docs and scaffold

`test_litellm_extra_pin_sync.py`'s base-vs-extra comparison became meaningless and now asserts what still matters: litellm is *not* a base dep, the extra declares it exactly once, and the extra agrees across both manifests.

Corrected several places that described the pre-move state as current — the `[litellm]` extra comment ("deleted at the next major"), `man/content/llm.md` and `docs/python/llm/index.md` ("a no-op today", "moving in a future major"), and the scaffolded `requirements.txt.tmpl` comment ("LiteLLM ships in the base install today, so this line currently changes nothing" — now false *and* dangerous). Also dropped a bare `litellm` pin from two big-3 example agents that never needed it.

## Test plan

- [x] Python unit suite (CI invocation): **2537 passed, 6 skipped**
- [x] `pytest scripts/` 106 passed · `go build ./...` · `go test ./src/core/cli/...` green
- [x] 150/150 module import walk in a slim venv
- [x] Live big-3 + long-tail, buffered and streaming
- [x] Extra-layering install verified, no base upgrade
- [ ] **src-tests + full integration** — running now; `tsuite-mesh:local` becomes litellm-free, so the whole suite runs against a slimmed image. Every integration model string is big-3 and nothing sets `MCP_MESH_NATIVE_LLM=0`, so no suite change is expected — but that is a prediction until the run lands.
- [ ] CI green

## For the 3.5.0 release notes

`RELEASE_NOTES.md` lines 95/96/117 describe the 3.4.0 state — "litellm is still installed by default", "additive today". Accurate when shipped; 3.5.0 needs an entry calling this out as the breaking change, and the upgrade action is one line in `requirements.txt`.

Closes #1383

## Summary by CodeRabbit

* **Documentation**
  * Clarified LLM provider installation: Claude, GPT, and Gemini work with base install only; all other providers require the `[litellm]` extra.
  * Added guidance for containerized deployments and troubleshooting missing dependencies.

* **Refactor**
  * Made LiteLLM an optional dependency to reduce base package size.
  * Improved logging to provide clearer guidance when optional dependencies are missing.

### DIFF
--- a/cmd/meshctl/templates/python/llm-provider/requirements.txt.tmpl
+++ b/cmd/meshctl/templates/python/llm-provider/requirements.txt.tmpl
@@ -5,8 +5,9 @@
 
 # LiteLLM provider path (model: {{ .Model }}).
 # mcp-mesh bundles native SDK adapters for Anthropic, OpenAI and Gemini only;
-# every other vendor dispatches through LiteLLM. LiteLLM ships in the base
-# mcp-mesh install today, so this line currently changes nothing. It is
-# declared so this agent keeps working when LiteLLM becomes an opt-in extra.
+# every other vendor dispatches through LiteLLM, which is an opt-in extra and
+# is NOT part of the base install or the python-runtime base image. Without
+# this line the agent starts and registers normally, then fails on its first
+# LLM call.
 mcp-mesh[litellm]==3.4.0
 {{- end }}
--- a/docs/python/llm/index.md
+++ b/docs/python/llm/index.md
@@ -21,9 +21,9 @@ MCP Mesh provides first-class support for LLM-powered agents through the `@mesh.
 The `mcp-mesh` package includes LLM support out of the box:
 
 - **Claude** (Anthropic), **GPT** (OpenAI) and **Gemini** (Google) dispatch through bundled native SDK adapters
-- **100+ other providers** (Bedrock, Cohere, Ollama, …) dispatch through LiteLLM
+- **100+ other providers** (Bedrock, Cohere, Ollama, …) dispatch through LiteLLM, installed via the `[litellm]` extra
 
-No additional packages needed. Just set your API keys:
+For the big-3, no additional packages are needed. Just set your API keys:
 
 ```bash
 export ANTHROPIC_API_KEY=sk-ant-...
@@ -32,16 +32,21 @@ export OPENAI_API_KEY=sk-...
 
 !!! note "LiteLLM"
 
-    The big-3 vendors above need nothing beyond the base install. LiteLLM is
-    also part of the base install today, so long-tail providers work with no
-    extra step either — but it is moving to an opt-in extra in a future major
-    release. Declaring `mcp-mesh[litellm]` now is a no-op today and keeps a
-    long-tail agent working across that change; `meshctl scaffold` already
-    writes that pin into the generated `requirements.txt` when the selected
-    model is not Anthropic, OpenAI, Gemini or Vertex AI. `vertex_ai/*` is
-    native despite the long-tail-looking name: it runs the same Gemini models
-    through the same bundled SDK, only the auth differs (ADC / Workload
-    Identity instead of an AI Studio API key).
+    The big-3 vendors above need nothing beyond the base install. Every other
+    vendor dispatches through LiteLLM, which is an opt-in extra:
+
+    ```bash
+    pip install 'mcp-mesh[litellm]'
+    ```
+
+    For containerized agents, add `mcp-mesh[litellm]==<version>` to the
+    agent's `requirements.txt` so the image build picks it up — `meshctl
+    scaffold` writes that pin for you when the selected model is not
+    Anthropic, OpenAI, Gemini or Vertex AI. Without it the agent starts and
+    registers normally and fails on its first LLM call, with an error naming
+    the extra. `vertex_ai/*` is native despite the long-tail-looking name: it
+    runs the same Gemini models through the same bundled SDK, only the auth
+    differs (ADC / Workload Identity instead of an AI Studio API key).
 
 ## @mesh.llm Decorator
 
--- a/examples/docker-examples/agents/claude-provider/requirements.txt
+++ b/examples/docker-examples/agents/claude-provider/requirements.txt
@@ -4,8 +4,5 @@ mcp-mesh>=3.4.0
 # FastMCP for MCP server
 fastmcp>=3.0.0,<4.0.0
 
-# LiteLLM for LLM provider
-litellm>=1.80.5,!=1.82.7,!=1.82.8
-
 # HTTP client for health checks
 httpx
--- a/examples/docker-examples/agents/openai-provider/requirements.txt
+++ b/examples/docker-examples/agents/openai-provider/requirements.txt
@@ -4,8 +4,5 @@ mcp-mesh>=3.4.0
 # FastMCP for MCP server
 fastmcp>=3.0.0,<4.0.0
 
-# LiteLLM for LLM provider
-litellm>=1.80.5,!=1.82.7,!=1.82.8
-
 # HTTP client for health checks
 httpx
--- a/packaging/pypi/pyproject.toml
+++ b/packaging/pypi/pyproject.toml
@@ -46,9 +46,9 @@ requires-python = ">=3.11"
 # openai SDK is included by default for native OpenAI dispatch (issue #834
 # PR 2). When MCP_MESH_NATIVE_LLM is unset (default ON), OpenAI provider
 # agents dispatch through the native openai SDK; without it they would
-# silently fall back to LiteLLM. (openai is also a transitive dep of
-# litellm today; promoting to a direct base dep prevents breakage if
-# litellm ever drops it as transitive.)
+# silently fall back to LiteLLM. (openai used to arrive transitively via
+# litellm; since litellm left the base install in 3.5.0 — issue #1383 —
+# this direct entry is the ONLY thing that installs it.)
 #
 # google-genai SDK is included by default for native Gemini dispatch
 # (issue #834 PR 3). Supports both AI Studio (gemini/*) and Vertex AI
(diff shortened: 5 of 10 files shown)


dhyansraj:
Integration verification complete. **Green against a genuinely litellm-free image.**

## The image really is slim

Confirmed on the build host (beelink5, where the suite runs — my first check inspected a stale June image in local Docker and reported the opposite):

```
tsuite-mesh:local   created 2 minutes ago
mcp-mesh            3.4.0
import litellm   →  ModuleNotFoundError: No module named 'litellm'
```

So every integration case below dispatched through the native adapters with LiteLLM absent from the image — which has never previously been true.

## Results

| Run | Result |
|---|---|
| src-tests | 12 passed, 0 failed |
| Integration #1 | 557 passed, **2 failed**, 9 skipped |
| Scoped rerun of both | 4 passed, 0 failed |
| Integration #2 | **559 passed, 0 failed**, 9 skipped |

Run 1's two failures:

- `uc10_toolcalls/tc21_consumer_java_provider_claude_py_tool_java` — 502, `context deadline exceeded` reaching the target agent, after 186s
- `uc26_a2a_consumer_typescript/tc01_capability_and_tag_registration` — agent not yet visible in the registry

Both passed in run 2, so **the failure set moved** — contention, not a stable regression. That is the distinction that matters; passing when scoped would not have been sufficient on its own.

The failure *shapes* independently argue against causation: a missing LiteLLM surfaces as an `ImportError` at the point of use — demonstrated in this PR's own live testing, where `cohere/command-r` and `ollama/llama3` produced the friendly error naming the extra. A 502 timeout and a not-yet-registered agent are load symptoms. And `uc26` involves no LLM at all.

Worth noting for its own sake: `tc21` failed at 186s under `--parallel 8` and passed scoped at 111s. It drives real Claude calls through a Java → Python → Java chain while seven other suites also make LLM calls. It is close to its budget the same way `tc11` was before #1459 widened that one — a candidate for the same treatment, separately from this PR.

87 LLM-path cases pass with no LiteLLM present.

Issue #1312, opened 2026-07-07
fix(python): FastMCP transport rejects k8s Service-DNS /mcp calls with 421 (DNS-rebinding host validation) + pin fastmcp/mcp


dhyansraj:
## Problem

Python mesh providers reachable by a **non-localhost `Host`** (e.g. a k8s Service DNS) reject every `/mcp` call with **`421 Misdirected Request`**, breaking all cross-pod tool/LLM calls to Python providers.

Reproduced against a running Python provider:
```
Host: <service-dns>:8080  → 421 Misdirected Request   (body: "Misdirected Request")
Host: localhost           → 200
Host: 127.0.0.1           → 200
```
`/health` returns 200 regardless — only the MCP streamable-HTTP `/mcp` route enforces it.

## Root cause

- `mcp`'s `TransportSecuritySettings` defaults to `enable_dns_rebinding_protection=True` with an **empty `allowed_hosts`** (i.e. localhost/127.0.0.1 only) in recent versions (observed **fastmcp 3.4.3 / mcp 1.28.1**). Any other `Host` → 421.
- Mesh builds the streamable-HTTP app (`src/runtime/python/_mcp_mesh/engine/http_wrapper.py:158`, `src/runtime/python/mesh/decorators.py:1706`) with `stateless_http=True, transport="streamable-http"` and **passes no transport-security config**, so it inherits the localhost-only default. `FastMCP.http_app()` exposes exactly the knobs we omit: `host_origin_protection`, `allowed_hosts`, `allowed_origins`.
- **This is transitive drift, not an intentional bump.** The Python deps are loose unpinned ranges (`fastmcp>=3.0.0,<4.0.0`, `mcp>=1.26.0,<2.0.0`) with **no Python lockfile**, so the resolved version — and this default — is decided at image-build time. The pin strings did not change between releases; a later rebuild silently picked up the newer default-on behavior.

## Impact

Every Python mesh provider addressed by Service DNS in k8s returns 421 on `/mcp` — all cross-pod tool and LLM-provider calls to Python providers fail. Java providers use a different transport and are unaffected. Not reproducible in single-instance/localhost dev; also slipped the k8s integration suite, which does not call a Python provider via a non-localhost `Host`.

## Fix

[...]

PR #1465, opened 2026-08-04
feat(deps): lock the Python dependency set and install from it

### LINKED ISSUE #1454: Python is the only runtime without a lockfile; vendor SDK drift has broken us twice
## Summary

Python is the **only** mcp-mesh runtime without a lockfile. Rust has `Cargo.lock`, TypeScript has `package-lock.json`, Helm has `Chart.lock` — and `scripts/check_release_lockfiles.py` already asserts none of them silently re-resolve during a version bump. Python has nothing, so `pip install mcp-mesh==X.Y.Z` resolves a different dependency tree depending on when you run it.

This is the follow-up `pyproject.toml` already asks for, in the comment added by the #1312 fix:

> Pinned exact (#1312): an upstream FastMCP default-flip enabling DNS-rebinding Host validation drifted in on rebuild and broke every k8s Service-DNS /mcp call with 421. Loose ranges hid it. **Follow-up: add a full transitive lockfile.**

## It has now bitten twice

**#1312** — an unpinned `fastmcp` flipped a default on rebuild and returned 421 for every k8s Python provider. Fixed by pinning `mcp==1.28.1` / `fastmcp==3.4.3` exactly.

**PR #1453** — every Responses streaming test failed in CI while passing locally:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Response
usage.input_tokens_details.cache_write_tokens
  Field required [type=missing, input_value={'cached_tokens': 0}, input_type=dict]
```

Local `.venv` had **openai 2.14.0**; CI resolved **2.52.0**, where `InputTokensDetails` gained a required field. Note this was a *minor* bump introducing a required field — so semver discipline cannot be assumed for the vendor SDKs.

## Current specifiers

The vendor SDKs are not merely unpinned; they span major versions with no upper bound:

```
"anthropic>=0.77",
"openai>=1.60",          # spans 1.60 -> 2.52, two majors
"google-genai>=1.22",
"litellm>=1.80.5,!=1.82.7,!=1.82.8",   # exclusions already record two bad releases
"fastapi>=0.135.0",      # 0.x: every minor can break
```

## Why the Docker images make this worse

`packaging/docker/python-runtime.Dockerfile` does `pip install --no-cache-dir mcp-mesh==${VERSION}`, so **two builds of the same mesh version can ship different dependency trees**. That is exactly the mechanism by which the FastMCP default-flip reached users in #1312 — no mesh change, no release, different behaviour on rebuild.

## Proposed work

**1. Upper bounds in `pyproject.toml`** *(landing separately, ahead of this)* — `openai>=2.14,<3`, `anthropic>=0.77,<1`, `google-genai>=1.22,<2`, and an upper bound on `fastapi`. Cheap and keeps resolution flexible. Would have caught a 1.x→2.x jump; would **not** have caught the 2.14→2.52 minor, which is why it is not sufficient alone.

**2. A full transitive lockfile.** The tested set, generated and reviewed deliberately rather than resolved at install time.

**3. Docker runtime images install from the lock**, so a rebuild of a given mesh version is reproducible.

**4. Extend `check_release_lockfiles.py`** to cover the Python lock exactly as it covers `Cargo.lock` and `Chart.lock` — a version bump may move only mesh versions; anything else moving is a re-resolution and should fail loudly.

**5. A deliberate bump path** — dependency upgrades become their own reviewed PR that moves the lock, with CI proving the suite green against the new set.

## Design note: this is a published library

Hard `==` pins in `pyproject.toml` would conflict for users who depend on `openai` or `anthropic` directly — pinning would export our testing constraint onto their dependency graph. The existing `mcp`/`fastmcp` exact pins are tolerable because almost nobody depends on those directly; the vendor SDKs are different.

So the tight, exact set belongs in the **lockfile** (used by CI and the Docker images), with the library metadata staying permissive-but-bounded. Splitting it this way is what makes "test against a known set" and "remain installable alongside other packages" both true.

## Acceptance

- A rebuild of a given mesh version resolves the same dependency tree.
- CI runs against the same set a user gets.
- A version bump cannot silently re-resolve the graph; changing dependencies requires a reviewed change to the lock.

dhyansraj:
## Summary

Python was the only runtime without a lock — Rust has `Cargo.lock`, TypeScript `package-lock.json`, Helm `Chart.lock`, and `check_release_lockfiles.py` already asserts none of them re-resolve during a bump. Meanwhile `python-runtime.Dockerfile` did `pip install mcp-mesh==${VERSION}`, so **two builds of one mesh version could ship different trees**. That is the mechanism behind #1312, and it broke CI again on #1453.

Completes #1454. Item 1 (upper bounds) landed in #1455; this is the lock, the images installing from it, and the guard.

## Mechanism: a constraints file — chosen on measurement

Resolution was generated four ways:

| environment | vs. linux/amd64 py3.11 |
|---|---|
| linux/arm64 py3.11 | **identical**, 112/112 |
| linux/amd64 py3.12 | −1 (`backports-tarfile`, a `python_version < "3.12"` conditional) |
| macOS arm64 py3.11 | −2 (`jeepney`, `secretstorage` — dbus keyring) |

Linux/3.11 is a strict **superset** of every environment we build, and constraints semantics absorb a superset exactly — an entry for a package that isn't installed is inert. So one marker-free file is correct for both arches, 3.11–3.14 and macOS, with **no new tooling at install time**. That is what defeats the usual cross-platform objection to a pip-compile artifact: it's an objection to using it as a *requirements* file, and it evaporates under `-c`.

`uv lock` would buy universal marker-correct resolution — which that table shows we don't need — at the cost of a new tool in CI, both Dockerfiles, every maintainer's machine, and a TOML format the guard must learn.

**Hashes are out of scope, tested rather than assumed.** pip accepts them in a constraints file but flips the *entire command* into `--require-hashes` mode → `ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==`, which structurally cannot install `pip install -e .` — the install CI most needs constrained. Supply-chain integrity is a real but separate goal needing a separate install path; recorded in the docstring.

**`--strip-extras` is mandatory, not cosmetic.** Without it pip-compile emits `google-auth[requests]==2.56.2` and pip hard-errors `Constraints cannot have extras`. The guard catches that class.

## A finding that changed the design

The first image build **failed**: `mcp-mesh 3.4.0 depends on rich<14.0.0` vs a lock saying `rich==15.0.0`.

The two manifests are not the same document. `packaging/pypi/pyproject.toml` carries upper bounds `src/runtime/python/pyproject.toml` lacks on **ten** packages (`rich`, `uvicorn`, `pydantic`, `httpx`, `fastapi`, `aiohttp`, `click`, `python-dateutil`, `typer`, `urllib3`). One is binding today: **CI has been testing `rich 15.0.0`, a version no user and no image can ever have.** That is this issue's acceptance criterion #2 already violated, one layer below where the issue was looking.

So the lock resolves against `packaging/pypi/pyproject.toml` — the metadata PyPI actually serves. Its bounds are the tighter of the two, so the pins are valid for a user, for both images, *and* for an editable CI install. **The lock now structurally enforces manifest agreement.**

I did not unilaterally reconcile the manifests — that's a judgement about which bounds are right, and #1455 just did deliberate work there. See below.

## Evidence

**Reproducible**, two independent fresh venvs in `python:3.11-slim`:
```
two independent fresh installs:  IDENTICAL (112 packages)
installed vs lock:               EXACT MATCH: 112 locked pins == 112 installed
```

**Load-bearing**, not a coincidence of today's PyPI — same manifest range `openai>=2.14,<3`:
```
unconstrained                     -> openai 2.53.0
-c lock with openai moved to 2.14 -> openai 2.14.0
-c the real lock                  -> openai 2.53.0
```

**Both arches, one file:**
```
BUILD OK (amd64)  112 image packages / 112 lock pins, zero at a non-locked version
BUILD OK (arm64)  EXACT MATCH
CLI BUILD OK
```

**The guard's red and green, against the real 112-pin file:**
```
green: bump 3.4.0 -> 3.5.0, lock untouched  -> exit 0
red:   same bump, openai 2.53.0 -> 2.54.0   -> exit 1, "openai: 2.53.0 -> 2.54.0"
```

All four new assertions mutation-tested — each goes red when its check is defeated. The `_git_show`-returning-`None` history was specifically guarded against: the Python lock has no mesh entry, so its bump signal comes from a *second* file, and an unreadable one resolving to "not a bump" would skip and **pass**. It raises instead, pinned by `test_python_lock_without_a_manifest_fails_instead_of_skipping`.

The agent's own tests caught a real parser bug it had written — `]` inside `anthropic[bedrock]` terminated the dependency array early, silently under-reporting what the lock must cover.

## Test plan

- [x] `pytest scripts/` — **134 passed** (was 124)
- [x] Python unit suite against the locked set — 2531 passed, 6 skipped; unlocked baseline identical
- [x] Docker builds on amd64 and arm64, installed set matches lock exactly
- [x] Generator deterministic: two runs byte-identical; `--upgrade` then plain refresh produces no drift
- [x] hadolint clean both Dockerfiles; shellcheck clean on the generator
- [ ] CI green

## Does this change what CI installs? Yes — one package

```
- rich==15.0.0
+ rich==13.9.4
```

CI moves from a version users can never have to the version they do get. Suite green both ways, same 2531. Everything else absent from the lock is dev tooling (`pytest`, `ruff`, `mypy`, …) — deliberately unlocked, since it's never shipped and would churn the file on every ruff release.

## Three things needing your call

1. **`dependency-update.yml` was broken** — it ran `pip-compile` against `requirements-dev.txt` / `requirements-prod.txt`, neither of which exists anywhere in the repo, so it could only fail. It's repointed at the real lock (`--upgrade` → PR with the pin diff in the body), which implements the issue's item 5. **The original weekly cron was kept**, so this turns on a weekly dependency PR that was previously a no-op. If you'd rather it be `workflow_dispatch`-only, that's a one-line deletion.
2. **The manifest divergence remains.** The lock shields CI and the images from it, but the ten missing upper bounds are still there. Worth its own issue; `tests/unit/test_litellm_extra_pin_sync.py` is the precedent for a cross-manifest sync test.
3. **`mcp-mesh-core` is deliberately unlocked.** Published metadata says `mcp-mesh-core>=3.4.0`, so rebuilding a 3.4.0 image after 3.5.0 ships picks up core 3.5.0 — the #1312 shape, for our own package. Excluded because a stale pin would make the *next* release's image unresolvable. Cheapest fix is `~=` in the published metadata, not here.

Closes #1454

## Summary by CodeRabbit

* **New Features**
  * Added a complete, reproducible lock for Python dependencies, ensuring consistent installations across supported environments.
  * Docker-based installations now use the locked dependency set.

* **Bug Fixes**
  * Improved release validation to detect missing, unpinned, or unexpectedly changed Python dependencies.

* **Documentation**
  * Added contributor guidance for managing, refreshing, validating, and troubleshooting dependency locks.

* **Tests**
  * Expanded coverage for dependency locking, release checks, and version management.


dhyansraj:
Review round — **six fixed, three skipped.**

## The one that mattered: a test that could not fail

`test_bump_version.py`'s constraints.txt guard did:

```python
if not lock.exists():
    return
for handler in bv.HANDLERS:
    assert lock not in bv._glob_files(handler.globs)
```

Two independent ways to pass while asserting nothing: an early return if the file ever moves, and a raw path comparison that silently never matches if `_glob_files` returns paths in a different form.

**Demonstrated**, not argued — simulating a walker refactor to `f.resolve()` with `PROJECT_ROOT` reached through a symlink:

```
OLD raw comparison: control found by NOTHING
OLD raw comparison: lock assertion passes: True (vacuously)
NEW comparison:     control found by ['Documentation (--version OLD)', ...]
NEW comparison:     lock found by NOTHING
```

Now hard-fails on a missing lock, normalises both sides, and carries two positive controls covering **both** `_glob_files` code paths — `packaging/pypi/pyproject.toml` (literal glob) and `docs/contributing.md` (`**` walk).

## The rest

**`peter-evans/create-pull-request` v5 → v8.** Three majors behind, running deprecated `node16`. Both `action.yml` files were fetched and diffed rather than assumed: v8 is a strict superset (`branch-token`, `sign-commits`, `maintainer-can-modify`, `pull-request-branch` added), **nothing renamed or removed**, runtime `node16` → `node24`. All six inputs we pass are unchanged. Floating `@v8` to match the repo's convention.

**Quoted `[dev]` in install commands.** It's a shell glob — bash passes it through on no-match, but zsh errors, and zsh is the macOS default. Reproduced: unquoted → `zsh: no matches found`, exit 1; quoted → exit 0. Fixed in `docs/contributing.md`, `Makefile`, `packaging/README.md`.

**The lock's comments overclaimed.** They presented `pip install "mcp-mesh==X.Y.Z" -c constraints.txt` as a supported flow, but the file isn't shipped in the wheel — a PyPI user has no way to obtain it. Scope is now stated as repo/CI/images, with an explicit "NOT shipped inside the published wheel", and `packaging/pypi/pyproject.toml` records that its ranges are therefore the only bound a user actually gets. Same overclaim corrected in `ci.yml`.

**`_strip_toml_comment` documents its real support.** It toggles on `"` only, so TOML literal strings and escaped quotes are out of scope. Impact is low, but mis-parsing silently drops a requirement from the coverage check — i.e. the guard under-reporting — so the limit is now written down, with a cross-note in `_toml_array_strings` which shares the assumption.

**Second `RefError` branch covered end to end.** A `dynamic = ["version"]` manifest now has a test asserting exit 1 and its own distinct message. Confirmed reachable *and* non-vacuous: defeating the branch makes the run exit 0 and the test red. The fixture's `[tool.poetry] version = "0.0.0-decoy"` doubles as a check that the section-scoped parser doesn't answer from the wrong table.

## Skipped

- **SHA-pinning the action.** The repo tag-pins every action — 38 × `actions/checkout@v4`, zero SHA pins anywhere. Pinning one is inconsistent and doesn't change the posture; that's a repo-wide policy decision, not this PR's.
- **Digest-pinning `python:3.11-slim` in the generator.** The generator's *output* is the reviewed artifact — if the base shifts and resolution changes, it lands in the `constraints.txt` diff and CI tests it. A digest adds staleness burden for a guarantee review already provides, and pip-tools is version-pinned already.
- **Rewording "pin it out in `packaging/pypi/pyproject.toml`".** The suggested rewrite puts exact pins in `constraints.txt`, which is generated and must not be hand-edited. The current text is correct.

## Noted, not fixed

`ci.yml` lines 34, 67, 272 still carry unquoted `pip install -e .[dev]`. Runners are bash, so it's not reachable today, and two of the three are outside this PR's diff — quoting one would leave the file inconsistent. Worth a repo-wide sweep separately.

`pytest scripts/` — **135 passed** (was 134). `check_release_lockfiles.py` exit 0: 112 exact pins, 24/24 declared deps covered. Both workflows parse; the rendered step shows `@v8` with inputs intact.

PR #1468, opened 2026-08-04
fix: probe liveness and readiness separately, so an upstream outage cannot restart a pod

### LINKED ISSUE #1467: helm: liveness and readiness both probe /health, so an upstream outage restart-loops the provider
## Summary

The agent chart points **all three probes at `/health`**, and on Python `/health` consults the user's `health_check`. So an upstream vendor outage fails the **liveness** probe and Kubernetes restarts the pod — which cannot fix the vendor, and destroys the only state that knew the provider was failing.

```yaml
# helm/mcp-mesh-agent/values.yaml
startupProbe:    path: /health
livenessProbe:   path: /health     # failureThreshold: 3
readinessProbe:  path: /health
```

## Observed failure

Two LLM providers (claude, gpt), consumer prefers `+claude`:

1. The Anthropic API goes down. The scaffolded `health_check` detects it — `meshctl scaffold llm-provider` generates a check that validates *"Anthropic API is reachable"*.
2. `/health` returns 503 → readiness fails (correct) **and liveness fails** (wrong).
3. After 3 failures Kubernetes kills the pod.
4. It restarts, re-registers as healthy, and resolution switches back to claude.
5. Anthropic is still down. Repeat indefinitely.

The consumer flaps between providers for the entire outage instead of settling on gpt.

## Root cause

Mesh cannot say *"don't route to me"* without also saying *"restart me"*, because liveness and readiness are the same URL. Restarting is the wrong remedy for an upstream outage, and it launders a known-bad provider back into the pool by erasing the evidence.

Note this is **not** a missing resolver feature. Resolution is already health-aware — `ReasonUnhealthy`, `ReasonUnavailable`, `ReasonUnreachable` and `ReasonDeregistering` are live eviction reasons in `src/core/registry/resolver.go`. The registry correctly evicts the unhealthy provider. The problem is that the pod restart makes it healthy again.

## The Python runtime is already correct

`src/runtime/python/mesh/decorators.py:460-480` registers all three endpoints with the right semantics:

| endpoint | semantics |
|---|---|
| `/health` | consults the user's `health_check` |
| `/ready` | readiness, consults the health-check result |
| `/livez` | *"always returns 200 if app is running"* — pure liveness |

**No Python runtime change is needed.** The chart simply never used `/livez`.

## Java and TypeScript lack the endpoint

| runtime | `/health` | `/ready` | `/livez` |
|---|---|---|---|
| Python | user health check | ✅ | ✅ |
| TypeScript | static `{status:"healthy"}` | `handle !== null` | **missing** |
| Java | `runtime.isRunning()` | **missing** | **missing** |

So repointing the chart *before* those runtimes have `/livez` would 404 the liveness probe and **restart-loop every Java and TypeScript agent** — agents that never had this bug. The endpoint must land first.

Note Java and TypeScript also have no user-health-check concept at all, so they cannot currently surface a vendor outage in either direction. That is a **separate feature**, tracked separately — this issue is only the endpoint and the probe wiring.

## Scope

1. **Java**: add `/livez` (always 200 while the runtime is up) and `/ready`. `MeshHealthController` already has the shape.
2. **TypeScript**: add `/livez`. `express.ts:196` already has `setupHealthEndpoints()`.
3. **Chart** (`helm/mcp-mesh-agent/values.yaml`): `livenessProbe` → `/livez`, `readinessProbe` → `/ready`.
4. **Check `helm/mcp-mesh-registry/values.yaml`** — it points both probes at `/health` too, but the registry is a Go binary with no user health check, so its `/health` may already be pure liveness. Verify rather than assume.

## Upgrade constraint

**Chart and runtime images must move together for Java and TypeScript agents.** A chart pointing liveness at `/livez` against a 3.5.0 Java image gets a 404 and restarts a healthy agent. Python agents are safe either way — `/livez` has existed there all along.

## Acceptance

- An upstream vendor outage makes a Python provider **unready**, never **restarted**.
- Java and TypeScript agents serve `/livez` and are unaffected by the chart change.
- The registry chart is either confirmed correct or fixed alongside.

dhyansraj:
## Summary

The agent chart pointed **all three probes at `/health`**, and on Python `/health` consults the user's `health_check`. So an LLM vendor outage failed **liveness** and Kubernetes killed the pod — which cannot fix the vendor, and destroys the only state that knew the provider was failing. It restarts, re-registers healthy, resolution switches back, and the consumer flaps for the entire outage.

**Not a resolver gap.** `ReasonUnhealthy` / `ReasonUnavailable` / `ReasonUnreachable` are already live eviction reasons, and the registry evicted correctly. The *restart* is what made the provider healthy again.

| runtime | before | after |
|---|---|---|
| Python | `/health`, `/ready`, `/livez` all correct | **unchanged** |
| TypeScript | `/health`, `/ready` | + `/livez` |
| Java | `/health` only | + `/livez`, `/ready` |
| chart | 3 probes → `/health` | startup → `/health`, liveness → `/livez`, readiness → `/ready` |

## The finding that would have caused a worse bug

**TypeScript has two health surfaces, not one.** MCP agents (`MeshAgent`) get `/health` and `/ready` from **fastmcp's own request handling**, not from `express.ts`. Patching `express.ts` alone would have left every `@mesh.tool` TypeScript agent 404ing `/livez` — producing exactly this restart loop in the runtime that never had it. `/livez` is now registered on FastMCP's Hono app too (fastmcp consults Hono first), mirroring `jobs-cancel-route.ts` conventions.

## `startupProbe` moved as well — the reasoning

A startup probe's failure action is **also a container restart**. Left on `/health`, a Python provider that boots *during* the outage never passes startup (`failureThreshold: 30 × periodSeconds: 10` = 300s) and gets killed, then restarts into the same outage — the same bug on a five-minute cadence, re-entered on every restart.

Counter-argument weighed: `/health`-based startup is the only probe that can restart a process which binds its port but wedges during init. Judged the worse trade — it can't distinguish a wedged boot from an external outage, and a wedged agent on `/livez` still shows `0/1 READY` (diagnosable, no traffic, no flapping), which fits mesh's soft-fail posture better than CrashLoopBackOff. One-line revert if you disagree.

## Two details in the Java implementation

`/livez` deliberately does **not** consult `isRunning()` — a runtime that hasn't started yet would otherwise restart-loop through a slow boot. The agent-name lookup is in a try/catch because `runtime` is a `@Lazy` proxy that can raise mid-boot, and liveness must still answer 200.

`/ready` is `isRunning()`, with a javadoc saying plainly that Java has no user health check and this does **not** reflect dependency health. Better an honest signal than an implied one.

## Registry chart — verified, not assumed

`GetHealth` (`ent_handlers.go:156`) returns a hardcoded `healthy` with uptime and touches no DB, dependency or user check. Pure liveness already, so both probes sharing it cannot loop on anything external. No change; a comment records why, so the next reader doesn't re-derive it.

## Proof all three runtimes answer `/livez` — before the chart default moved

Real agents, real curls, killed by PID:

| runtime | `GET /livez` | `HEAD /livez` | `/ready` | `/health` |
|---|---|---|---|---|
| Python (`hello_world.py`) | 200 | 200 | 200 | 200 |
| TypeScript (`greeter`, fastmcp path) | 200 | 200 | 200 | 200 |
| Java (`basic-tool-agent`) | 200 | 200 | 200 `{"ready":true}` | 200 |

All four surfaces answered with no registry reachable. MeshExpress is covered by a test binding a real socket.

## Test plan

- [x] TypeScript **1277 passed** (89 files), `typecheck` clean; 9 new tests
- [x] Java **1175 passed**, 0 failures; 11 new tests incl. HEAD variants
- [x] Python **1565 passed, 1 skipped** — unchanged, as expected
- [x] `check_helm_render_matrix.py` **39/39** (was 38) — now asserts the probe paths, since a chart default that silently reverts *is* this bug returning
- [x] `helm lint` 0 failed · `check_helm_pss.py` · `check_helm_volume_keys.py` green
- [x] Fail-without-fix proven for each: removing the express `/livez` → 3 failures; making Java's `/livez` consult `isRunning()` → 5 failures; flipping the chart's liveness path → matrix FAIL
- [ ] CI green

## ⚠️ Upgrade constraint — needs a release note

**Chart and runtime images must move together for Java and TypeScript agents.** The chart now probes `/livez`; a 3.5.0 Java or TS image doesn't serve it, so the probe 404s and Kubernetes restarts a healthy agent. Python agents are safe either way — `/livez` has existed there all along. Nothing in the chart can detect the image version, so this is documentation-only.

## Noted, not fixed

The new `probe_paths` matrix case asserts the agent chart's shipped defaults only; a user overriding paths in their own values file can still collapse them, and the values comment is the only defence there.

Java and TypeScript still have no user-health-check concept, so their `/ready` cannot surface a vendor outage in either direction. That is the separate feature #1467 calls out and this PR deliberately leaves alone.

Closes #1467

## Summary by CodeRabbit

- **New Features**
  - Added dedicated `/livez` liveness and `/ready` readiness endpoints across supported runtimes.
  - Liveness checks now return promptly and independently of dependency or runtime readiness.
  - Readiness checks report whether the mesh is available, including appropriate success or failure status codes.
  - Added support for both `GET` and `HEAD` requests on probe endpoints.
  - Preserved the existing `/health` endpoint.

- **Documentation**
  - Improved deployment guidance describing probe behavior, endpoint semantics, and restart handling.

- **Tests**
  - Added coverage for probe routing, responses, status codes, and Helm configuration validation.

### DIFF
--- a/helm/mcp-mesh-agent/values.yaml
+++ b/helm/mcp-mesh-agent/values.yaml
@@ -309,10 +309,37 @@ agentCode:
   # Mount path
   mountPath: "/app/agent"
 
-# Health probe configurations (matches working examples)
+# Health probe configurations.
+#
+# Liveness and readiness MUST NOT share a URL (issue #1467). Every mesh
+# runtime serves three endpoints with three different meanings:
+#
+#   /livez   always 200 while the process is serving. Nothing else.
+#   /ready   should traffic be routed here? On Python this consults the
+#            user's health_check, so an upstream vendor outage makes the
+#            agent unready and the registry evicts it from resolution.
+#   /health  the diagnostic view for humans and `kubectl exec` curls.
+#            No probe points at it: it carries no consequence of its own.
+#
+# When liveness also pointed at /health, an upstream outage failed the
+# LIVENESS probe and Kubernetes restarted the pod. A restart cannot fix
+# someone else's API, and it erases the only state that knew this
+# provider was failing — the agent re-registers as healthy, consumers
+# route back to it, it fails again, forever. Readiness is the probe
+# whose remedy (stop sending traffic) matches that cause.
+#
+# The same reasoning applies to the startup probe: its failure action is
+# also a container restart, so gating startup on /health meant an agent
+# that booted DURING an outage never became "started" and was killed
+# after failureThreshold × periodSeconds — reproducing the same loop on
+# a slower cadence. It gates on /livez instead; readiness still governs
+# traffic, and an agent stuck mid-boot stays visible as 0/1 READY rather
+# than flapping through CrashLoopBackOff.
+#
+# Do not collapse these back onto one path.
 startupProbe:
   httpGet:
-    path: /health
+    path: /livez
     port: http
   initialDelaySeconds: 5
   periodSeconds: 10
@@ -321,7 +348,7 @@ startupProbe:
 
 livenessProbe:
   httpGet:
-    path: /health
+    path: /livez
     port: http
   initialDelaySeconds: 15
   periodSeconds: 10
@@ -330,7 +357,7 @@ livenessProbe:
 
 readinessProbe:
   httpGet:
-    path: /health
+    path: /ready
     port: http
   initialDelaySeconds: 10
   periodSeconds: 5
--- a/helm/mcp-mesh-registry/values.yaml
+++ b/helm/mcp-mesh-registry/values.yaml
@@ -366,7 +366,14 @@ envFrom: []
   # - secretRef:
   #     name: special-secret
 
-# Liveness and readiness probes
+# Liveness and readiness probes.
+#
+# Unlike the agent chart (issue #1467), both may share /health here: the
+# registry's GET /health (src/core/registry/ent_handlers.go) returns a
+# static 200 with uptime and never consults the database, a dependency,
+# or a user-supplied check. It is already a pure liveness signal, so it
+# cannot turn an outage elsewhere into a registry restart. Should it ever
+# grow a dependency check, split the probes as the agent chart does.
 livenessProbe:
   httpGet:
     path: /health
@@ -378,7 +385,7 @@ livenessProbe:
 
 readinessProbe:
   httpGet:
-    path: /health # Matches working examples
+    path: /health
     port: http
   initialDelaySeconds: 10
   periodSeconds: 5
--- a/scripts/check_helm_render_matrix.py
+++ b/scripts/check_helm_render_matrix.py
@@ -27,6 +27,18 @@ That is for values whose whole purpose is to add or remove a resource — an
 opt-in object silently becoming opt-out (or the reverse) is a behaviour change
 a non-empty render cannot detect.
 
+It may also pin the container probe paths:
+
+  probe_paths={"livenessProbe": "/livez", "readinessProbe": "/ready"}
+
+Issue #1467: liveness and readiness pointing at the SAME url is the bug —
+a dependency outage that should only make an agent unready instead restarts
+the pod, which cannot fix the dependency and launders the failing agent back
+into resolution. A values default is one careless edit away from collapsing
+back, and every render stays green when it does. Declaring the paths makes
+that edit fail here. Two probes sharing a path is additionally rejected
+outright, whatever the declared paths were.
+
 Usage: python3 scripts/check_helm_render_matrix.py  (run from anywhere)
 Exit code 0 = every case behaved as declared.
 """
@@ -63,6 +75,7 @@ class Case:
     extra_args: tuple[str, ...] = field(default_factory=tuple)
     requires_kind: str | None = None
     forbids_kind: str | None = None
+    probe_paths: dict[str, str] | None = None
 
 
 CASES: list[Case] = [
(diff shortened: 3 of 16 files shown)


dhyansraj:
Review round — one finding, valid, fixed. Nothing skipped.

## The fix

`registerLivezRoute` returns `false` on three paths and `_autoStart()` dropped it, so an agent could start with no `/livez` while the chart probes it. It now throws; `scheduleAutoStart()`'s existing handler turns that into a clear `console.error` + `process.exit(1)`.

**Why fail-fast, against mesh's usual soft-fail posture:** there is no degraded mode available here. Without the route the kubelet 404s liveness and Kubernetes restarts the pod every probe interval — so not throwing produces *the same* CrashLoopBackOff, just with a misleading cause. Throwing produces it with an accurate one.

Direct precedent: #1387/#1389, where route rebuilding depended on private FastAPI internals, one was renamed, the `ImportError` was swallowed and success reported. Made fail-fast in v3.3.2 as a documented behaviour change. `registerLivezRoute` reaches into FastMCP internals for its Hono app and carries the same fragility — if fastmcp's shape changes, this silently returns false and every TypeScript agent restart-loops. The reasoning is in the code comment, with an explicit "do not soften this back to a warning."

`registerCancelRoute`'s call site is deliberately unchanged: same code shape, but a genuinely soft failure mode — cancellation stops working while the agent keeps serving.

Cause and consequence are now stated once each rather than twice: the `console.error` branches carry only the cause (`FastMCP.getApp() unavailable: <reason>`), the thrown error carries the consequence and the agent id.

## What the throw exposed — worth more than the fix

**Seven spec files have been silently running real startup code in unit tests.** They restore their `_autoStart` prototype stub in a *synchronous* `afterEach`, so a `process.nextTick` auto-start tick still pending at that moment ran the **real** `_autoStart` against a stub FastMCP — real napi startups, real registry heartbeat attempts — invisibly, for as long as those tests have existed. The throw converted that silence into `process.exit(1)` inside the vitest worker, which is how it surfaced.

Fixed uniformly: `await` a `nextTick` before the restore, so the pending tick runs while the stub is still installed. One hunk per file, no test logic changed.

Affected: `agent-add-tool`, `agent-duplicate-tool-name`, `direct-invoke-required`, `service-view`, `unresolved-dep-no-shift`, `unwired-slot-warn`, `a2a/agent-a2a-config`.

## Verification

- `npx vitest run` → **89 files, 1280 passed, 0 unhandled errors, exit 0** (+3 new). Re-run twice more, exit 0 both times.
- `npm run typecheck` clean.
- **Fails without the fix**: reverting the call site to the bare call → `2 failed | 10 passed`, *"promise resolved 'undefined' instead of rejecting"*.
- The happy-path test asserts the real `_autoStart` still resolves and `on(["GET","HEAD"], "/livez", fn)` was called — so this can't pass by simply making startup always fail.
- **Live re-verified after the change**: rebuilt the SDK, started `examples/typescript/greeter` on :19104 → `GET /livez` 200 `{"alive":true,"agent":"greeter"}`, no `cannot start` lines. Killed by PID.

Java (1175) and Python (1565 passed, 1 skipped) untouched this round.

Issue #1430, opened 2026-07-30
Docs contradict each other on load balancing: the registry selects one deterministic winner, but two concept docs claim round-robin / DDDI load-balancing


dhyansraj:
## Summary

Three concept docs make three different claims about how mesh distributes calls across replicas. One is correct, two describe behaviour the registry does not have.

The registry does **not** load balance. It selects exactly one winner, deterministically. Distribution across replicas — where it happens at all — is Kubernetes balancing across the pods behind a Service, not mesh doing anything.

## Ground truth

`src/core/registry/resolver.go:480-495`:

```go
// Stage 6: tiebreaker — sort by tag score DESC, then highest semver version
// DESC, then agent ID ASC for determinism.
sort.SliceStable(afterSchema, func(i, j int) bool {
    if afterSchema[i].Score != afterSchema[j].Score { return afterSchema[i].Score > afterSchema[j].Score }
    if vc := s.matcher.CompareVersion(...); vc != 0 { return vc > 0 }
    return afterSchema[i].AgentID < afterSchema[j].AgentID
})
winner := afterSchema[0]
...
Reason: TiebreakerScoreThenVersion
```

`src/core/registry/llm_provider_resolver.go:104-118` does the same for LLM providers. When tag score, version and schema all tie, the ordering is stable on agent ID — the same candidate every time, not a rotating one.

The endpoint handed back is built from the winner's registered host/port. In Kubernetes that is Service DNS, so any spreading across pods is kube-proxy's, and it is invisible to and uncontrolled by mesh.

## The three claims

| file:line | claim | verdict |
|---|---|---|
| `docs/concepts/audit.md:144` and `src/core/cli/man/content/audit.md:141` | "Configurable tiebreakers (round-robin, latency-aware, etc.) are out of scope for v1" | ✅ **correct** — matches the code exactly |
| `docs/concepts/in-process-state.md:239` | "tool calls from any given consumer will **round-robin** across them" | ❌ mesh does not round-robin |
| `docs/concepts/dddi.md:210` | "Second calculator starts — **DDDI load-balances** between both instances" | ❌ DDDI does not load balance |
| `docs/concepts/dddi.md:119` | "Best match selected **(or load balanced)**" | ⚠️ misleading — the parenthetical implies an alternative mode that does not exist |

A reader comparing `audit.md` with `dddi.md` gets flatly contradictory statements about the same subsystem.

## Careful: `in-process-state.md`'s *conclusion* is right

[...]

PR #1471, opened 2026-08-05
docs: publish release notes on the site and document probe endpoints

### LINKED ISSUE #1470: docs: release notes are not on the site, changelog links drifted, and the v3.5.1 probe split is undocumented
Two documentation gaps, both surfaced while releasing v3.5.1.

## 1. Release notes are not on the docs site

`RELEASE_NOTES.md` lives at the repo root and is reachable only by browsing GitHub. Nothing on mcp-mesh.ai links to it, so a user on the upgrading page — exactly where a 3.5.1 reader lands, given the chart/image ordering constraint — has no path to "what changed and must I act."

Related: `.github/workflows/docs.yml` triggers only on `docs/**` and `mkdocs.yml`, so a release commit touching only `RELEASE_NOTES.md` would not redeploy the site even once the page exists.

## 2. `[Full Changelog]` links drifted away from their headings

The file's convention — verified mechanically across all 77 links — is that each link sits immediately above the `## vX.Y.Z` heading it describes:

```markdown
[Full Changelog](.../compare/v3.2.0...v3.2.1)

## v3.2.1 (2026-07-17)
```

70 links follow it. From **v3.3.0 onward** each release appended its link to the *top of the file* instead, so seven stacked between the H1 and the first version heading — and the link physically above `## v3.5.1` described a release seven sections away.

Nobody noticed because a 2,300-line markdown file is never read top-down. It was obvious within seconds of rendering it.

## 3. The v3.5.1 probe split is undocumented

#1467/#1468 changed which endpoint each Kubernetes probe should use, but touched neither `docs/` nor `src/core/cli/man/content/`. `/livez` appears in two man pages only as an aside about loop topology.

So anyone writing their own manifests rather than using the chart has no guidance — and the obvious choice, pointing every probe at `/health`, is precisely the restart-loop bug #1467 fixed. `meshctl man` cannot answer it offline either.

## Scope

- `docs/release-notes.md` sourcing the root file with **zero duplication**, plus a nav entry and a pointer from the home page.
- Move the six stranded changelog links to their own headings; keep only the unreleased pointer at the top.
- A few lines in the deployment man pages and their docs counterpart listing the three endpoints and which probe each serves. Short — `meshctl man` documents current state, not rationale.

## Acceptance

- Release notes readable at mcp-mesh.ai, with no second copy of the content that can drift.
- A release commit touching only `RELEASE_NOTES.md` redeploys the site.
- Every `## vX.Y.Z` heading has its own changelog link directly above it.
- Someone hand-writing probe config can find which endpoint to use, on both the web and offline surfaces.

dhyansraj:
Three related documentation gaps, all surfaced while releasing v3.5.1.

## 1. Release notes on the site — with no second copy

`docs/release-notes.md` is an 11-line stub whose only content is `--8<-- "RELEASE_NOTES.md"`. **Zero duplication** — the root file is read at build time.

`pymdownx.snippets` was chosen because it is *already* enabled, `base_path` already includes the repo root, and the tutorial already uses it 54 times to pull from `examples/`. No new plugin, no copy step, no symlink (a symlink would make `git-revision-date-localized` report the wrong date, and degrades to a text file on Windows checkouts — rendering as garbage rather than failing).

**Fail-loud, verified.** `check_paths: true` means a missing source kills the build by name:

```
ERROR - Error reading page 'release-notes.md': Snippet at path 'RELEASE_NOTES.md' could not be found
BUILD EXIT: 1
```

That flag immediately failed on a **pre-existing dead directive** — `auto_append: includes/abbreviations.md`, a file that has *never existed in this repo's history*, with no page defining a single abbreviation. Silently no-op'ing forever. Removed rather than stubbed. All 54 existing snippet references were then confirmed to resolve, so a renamed tutorial example will now break the build instead of rendering an empty code block.

**The docs workflow wouldn't have redeployed.** `docs.yml` triggered only on `docs/**` and `mkdocs.yml`, so a release commit touching only `RELEASE_NOTES.md` left the site stale. Added to the trigger paths.

Nav: after Security, before FAQ — Release Notes is something users actively look up, so it leads the meta tail rather than sitting behind it. Plus a one-sentence pointer on the home page after Project Status.

## 2. The changelog links had drifted

Rendering the file made this obvious in seconds. The convention — verified mechanically across **all 77 links**, not sampled — is that each sits immediately above the `## vX.Y.Z` heading it describes:

```markdown
[Full Changelog](.../compare/v3.2.0...v3.2.1)

## v3.2.1 (2026-07-17)
```

70 follow it. From **v3.3.0** onward each release appended its link to the *top of the file* instead, so seven stacked under the H1 — and the link physically above `## v3.5.1` described a release seven sections away. Two of those six were added by me while writing the 3.5.0 and 3.5.1 notes, by pattern-matching the top of the file rather than checking the older sections.

The six are moved to their headings. The unreleased pointer stays at the top and is relabelled **`[Unreleased changes]`** — two adjacent links both reading "Full Changelog" scan as a copy-paste slip. Link count is 77 before and after, so nothing was lost or doubled.

Also backticked a `+/-` at line 2271 that `pymdownx.smartsymbols` was rendering as `±`.

## 3. The v3.5.1 probe split was undocumented

#1467/#1468 changed which endpoint each probe should use and touched **neither** `docs/` nor the man corpus. `/livez` appeared only as an aside about loop topology. So anyone hand-writing manifests would point everything at `/health` — exactly the restart loop that change fixed — and `meshctl man` couldn't answer it offline.

Three bullets, after the existing "Always configure health checks" example:

```
- `/livez`  - livenessProbe and startupProbe. 200 while the process is serving; consults nothing else.
- `/ready`  - readinessProbe. Whether traffic should be routed here; reflects your health_check.
- `/health` - no probe. The diagnostic view: the /ready signal plus checks and errors.
```

Semantics were verified per runtime rather than assumed: on Python `/health` and `/ready` read the **same** stored value and both 503 together — `/health` merely carries `checks`/`errors` as well. Stated plainly rather than inventing a distinction.

**Bullets, not a table** — `renderer_test.go`'s #1430 provenance note records that the man renderer passes `|` rows through raw, so table markup would print literally.

**Added to `_java` and `_typescript` too.** A Java user never reads the Python page, and their `/ready` genuinely differs — neither runtime has a user health-check concept, so it reports only that the runtime is running. A cross-reference would have been wrong, not merely terse.

## Test plan

- [x] `mkdocs build` exit 0; fail-loud path verified by moving the source aside
- [x] All 54 pre-existing snippet references resolve under `check_paths`
- [x] Page renders 86 versions newest-first; every heading has its own link directly above it
- [x] `go test ./src/core/cli/man/...` green; goldens **+12 / +10 / +3**, reconciled exactly (10 spans across 3 bullets, 2 in the closing sentence, 3 new markup list lines)
- [x] `go build ./...` clean
- [ ] CI green

## Known and not fixed

**The TOC is long.** `toc.integrate` puts it in the left sidebar, and at `toc_depth: 3` that is **360 entries** — 86 versions plus 274 subsections, all expanded. It *is* a version list, but reaching v2.0.0 means scrolling past ~200 entries. Search rescues it in practice (362 indexed anchors — typing `3.2.0` jumps straight there). `toc_depth: 2` would give a clean 86-entry list but is global and strips `###` from every other page. A per-version split would need a generator, a nav plugin, 86 files churning per release, **and would break every existing anchor URL**.

**Four pre-existing changelog oddities**, each with two defensible repairs, so left for a decision: `## v3.2.3` has no link (v3.2.2 was never fully published); `## v1.1.0` is missing one; line 1494 duplicates v0.8.0's link above `## v0.7.21`; lines 2164/2174 are an interlocked off-by-one caused by v0.5.5 being tagged without a section. Everything at v0.5.1 and older predates the convention.

Closes #1470

## Summary by CodeRabbit

- **Documentation**
  - Added a Release Notes page to the documentation site, including version-specific changes, upgrade notes, and fixes.
  - Reorganized release-note links for easier navigation between current and historical releases.
  - Added Kubernetes health-probe guidance covering separate liveness, readiness, and diagnostic endpoints across supported deployment types.
  - Documented the risks of using readiness or diagnostic endpoints for liveness and startup checks.
  - Updated project status information to link directly to release notes.

- **Chores**
  - Documentation publishing now updates when release notes change.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,6 +7,7 @@ on:
     paths:
       - 'docs/**'
       - 'mkdocs.yml'
+      - 'RELEASE_NOTES.md'  # included verbatim by docs/release-notes.md
       - '.github/workflows/docs.yml'
       - 'examples/tutorial/**'
       - 'scripts/generate_tutorial_artifacts.sh'
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,18 +1,8 @@
 # MCP Mesh Release Notes
 
-[Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.5.1...HEAD)
+[Unreleased changes](https://github.com/dhyansraj/mcp-mesh/compare/v3.5.1...HEAD)
 
 [Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.5.0...v3.5.1)
-
-[Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.4.0...v3.5.0)
-
-[Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.3.2...v3.4.0)
-
-[Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.3.1...v3.3.2)
-
-[Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.3.0...v3.3.1)
-
-[Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.2.3...v3.3.0)
 
 ## v3.5.1 (2026-08-04)
 
@@ -29,6 +19,8 @@ A one-fix patch. The agent chart pointed liveness, readiness and startup at the 
 
 - **⚠️ Upgrade the agent chart and your Java or TypeScript images together.** The chart probes `/livez` for liveness and startup, and a 3.5.0 or older image of either runtime does not serve it — the probe 404s and Kubernetes restarts an agent that is perfectly healthy. This inverts the usual assumption that a chart tolerates older images, and nothing in the chart can detect the image version. Python agents are safe in either order; `/livez` has been served there since well before this release.
 - **Java and TypeScript `/ready` reports only whether the runtime is running.** Neither has a user health-check concept, so their readiness cannot yet reflect a dependency outage in either direction. Python's does.
+
+[Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.4.0...v3.5.0)
 
 ## v3.5.0 (2026-08-04)
 
@@ -52,6 +44,8 @@ A small release with two breaking changes, both in deployment and packaging rath
 - **Python agents on any vendor other than Anthropic, OpenAI or Gemini need `mcp-mesh[litellm]`** in `requirements.txt`. The extra has been installable since 3.3.2 and layers over an existing install without upgrading the base. The big three need no change.
 - **⚠️ If a chart is already wedged in the dual-type volume state, delete the Deployment and let it be recreated before upgrading.** The rename is a clean remove-and-add only when the applying field manager owns the existing volume; otherwise the rename itself can stick.
 - **No wire-protocol, registry-schema, dependency-resolution or declaration-syntax changes.** The only runtime code change is the Python provider's vendor extraction, which now treats an absent `litellm` as the normal case instead of logging a startup warning.
+
+[Full Changelog](https://github.com/dhyansraj/mcp-mesh/compare/v3.3.2...v3.4.0)
 
 ## v3.4.0 (2026-08-01)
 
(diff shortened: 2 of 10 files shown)


dhyansraj:
Review round — two findings, both valid, both fixed. Nothing skipped.

Both were the same error: bullets that stated a **Python** truth as universal.

## Verified returns, per runtime

| Runtime | `/health` actually returns |
|---|---|
| Python | the stored `health_check` result with `checks`/`errors`, 200/503 |
| Java | `{status, agent}`, 200/503 on the **same `running` flag as `/ready`**, no diagnostic fields |
| TypeScript | `express.ts:199` — a fixed `{status:"healthy"}`. **No status code, no condition.** |

So "the diagnostic view of the same signal" was wrong twice over for TypeScript — neither diagnostic nor the same signal — and overstated for Java, which shares the signal but has nothing extra to show.

## The warning: guidance kept, reason scoped

"Pointing liveness or startup at `/health` or `/ready` turns an upstream outage into a pod restart" is true only where a user health check exists. On Java and TypeScript, `/ready` reports runtime state and cannot observe an upstream outage.

Deleting it for those runtimes would have been wrong — the guidance still holds, the *reason* differs. Each variant now carries a reason that is true for it: on Java and TypeScript, pointing liveness at `/ready` restarts pods that are merely still booting. TypeScript gets its own clause, because "both endpoints" is false there — its `/health` cannot fail, so `/ready` is the only harmful target.

`deployment.md` (Python) is untouched: its bullet and its warning were both already correct, which is the finding's own premise.

## Nothing describes the future state

The Java/TypeScript user-health-check API lands in the next PR. All three files state present tense — "has no user health check" — and that PR flips them as part of its own diff. `meshctl man` documents what exists; someone configuring Java probes this week must not act on a promise that isn't kept yet.

## The goldens did not move, and that is the finding

`wantInlineCodeSpans` / `wantListCodeSpans` / `wantMarkupListLines` stay at **1698 / 516 / 445**.

They sample **only the default variant**, so edits confined to a `_java` or `_typescript` file are invisible to them — meaning **neither of these two bugs could ever have been caught by the corpus tests**, and a review of those variant files cannot lean on that suite.

The provenance comment now says so explicitly rather than leaving it to imply coverage it doesn't give.

## Verification

`go test ./src/core/cli/man/...` ok · `go build ./...` clean · `mkdocs build` exit 0. Both variants rendered via `meshctl man deployment --java` / `--typescript` to confirm the new bullets style correctly.

PR #1599, opened 2026-09-06
fix(registry): heartbeat error classification, guarded job terminal writes, audit query cost

### LINKED ISSUE #1580: fix(registry): a transient DB error on HEAD heartbeat is answered 410 Gone, forcing fleet re-registration
## Summary
`ent_handlers.go:590-595` (`FastHeartbeatCheck`): `if err != nil || agentEntity == nil { c.Status(410) }`. Any `GetAgent` error, including a Postgres blip or pool exhaustion, is reported as "unknown agent". The Rust core treats 410 as `Unregistered` and immediately POSTs a full registration (`heartbeat.rs:197-203`). During a 5-second DB hiccup every agent in the mesh re-registers at once, which is the load spike most likely to prolong the outage. POST maps DB errors to 400 (`:226-238`), which the core treats as a registry error and backs off from, so the two paths disagree.

Related on the same path (`ent_handlers.go:590,611,619,631`, `ent_service.go:2399,1479`, `status_hooks.go:72,87`): HEAD runs seven queries and loads the agent row three times; every full POST fires the status hook (two more queries) even when status is unchanged.

## Fix
Distinguish `ent.NotFoundError` (410) from other errors (503, which the core already handles with backoff, `heartbeat.rs:204-218`). Load the row once per HEAD; short-circuit the status hook when status did not change.

CONFIRMED (source, two readers).

---

### LINKED ISSUE #1581: fix(registry): MeshJob terminal writes in Expire/Cancel lack a status predicate; claim guard omits working
## Summary
Every sibling sweep phase guards its terminal transition with `StatusIn(working, input_required)` or `StatusNotIn(completed, failed, cancelled)` (`ent_service_jobs.go:168,496,741,960,1218,...`). Two do not:

- `ExpireDeadlinedJobs` (`:1840-1851`) selects non-terminal candidates and then issues an unconditional `UpdateOneID(...).SetStatus(failed)`. A `complete()` landing between select and update is overwritten with `deadline_exceeded`.
- `CancelJob` (`:855-878`) checks `isTerminalStatus(current.Status)` inside the transaction but the UPDATE carries no predicate; under read-committed a completion that commits between the read and the write is overwritten with `cancelled`.
- `ClaimNextJob`'s guard (`:794-798`) omits `status=working`, so a claim racing an unowned-job cancel claims the cancelled row.

The windows are narrow, but the outcome (a completed job reported failed/cancelled, result discarded) is silent.

## Fix
Add the same status predicate to both UPDATEs and treat 0 rows affected as "lost the race"; add `working` to the claim guard.

CONFIRMED (source; narrowed by the coordinator: CancelJob does have the in-transaction read check).

---

### LINKED ISSUE #1582: perf(registry): audit prior-trace lookups run two unindexed JSON scans per dependency on every full heartbeat
## Summary
`audit_emit.go:61,75,233-238,260-268`: before the emission gating decides whether an audit event is even needed, the emitter fetches the last 64 `registry_events` rows and scans their JSON payloads twice per dependency to find the prior chosen provider. `src/core/ent/schema/registryevent.go` has no index supporting that query. On a full POST heartbeat from an agent with N dependencies this is 2N table scans over a table capped at 100k rows (`MCP_MESH_EVENT_MAX_ROWS`), on the hot path that every 202 triggers fleet-wide.

## Fix
Gate first (single-candidate clean resolutions are silent by design), then look up the prior trace; add a composite index on `(agent_id, event_type, created_at)` or store `prior_chosen` on the agent/dependency row so no scan is needed.

CONFIRMED (source).

---

dhyansraj:
Three independent registry defects from the September 2026 audit, grouped because they live in the same package and share a test harness.

## #1580 — a transient DB error told the whole fleet to re-register

`FastHeartbeatCheck` did `if err != nil || agentEntity == nil { c.Status(410) }`, so any `GetAgent` error — a Postgres blip, pool exhaustion — was reported as "unknown agent". The Rust core treats 410 as `Unregistered` and immediately POSTs a full registration, so a five-second DB hiccup made every agent in the mesh re-register simultaneously: the worst possible load during a DB problem.

Now `ent.IsNotFound(err)` → 410, everything else → 503, which the core already handles with backoff (`heartbeat.rs:204-218`) and Python maps identically (`fast_heartbeat_status.py:49-59`).

**This was the published contract already** — `api/mcp-mesh-registry.openapi.yaml:189-191` declares both 410 ("Unknown agent, please register") and 503 ("Registry error, back off") for this operation. The handler was violating its own spec; no spec change needed.

Same path, also fixed: HEAD loaded the agent row 4 times (the issue said 3; the 4th is ent's post-update re-select inside the tx `UpdateOneID` opens) — now 1. And every full POST woke the status hook even when status was unchanged; the mutation now only carries `status` when it actually changes.

## #1581 — terminal job writes could overwrite a completion

`ExpireDeadlinedJobs` and `CancelJob` selected a non-terminal candidate and then issued an unconditional `UpdateOneID(...).SetStatus(...)`. A `complete()` committing between the read and the write was overwritten with `deadline_exceeded`/`cancelled` and its result discarded — silently. Both now carry the `StatusNotIn(terminal)` predicate the sibling sweeps already use, and treat `Affected == 0` as losing the race rather than an error. `ClaimNextJob`'s guard gains `status = working`.

`ReleaseJob` had the identical unguarded read-then-write shape 200 lines below and is fixed alongside — including the destructive exhausted branch, which stamped `failed` over a completed row.

## #1582 — audit lookups ran two unindexed JSON scans per dependency

On every full heartbeat, before the gate decided whether an event was even needed. Now two memoised queries per *function* rather than 2N per *dependency* — a 32-dep function goes from 64 queries to 2 — plus a supporting index.

Measured on PostgreSQL 16 with 100k events: `Index Scan Backward`, no sort, **0.18 ms / 74 buffers**, versus `Seq Scan` + top-N heapsort at **8.1 ms / 1433 buffers**.

## Two places the issue's prescription didn't survive contact with the code

**"Gate first, then look up the prior trace" is not achievable.** The gate *consumes* the prior trace: both the chosen-producer flip and the unresolved→resolved transition are detected from it, and the skip condition literally reads `priorChosen == ""`. Gating first would silently drop the flip detection the code comments say operators want. The cost is reduced instead, which is a larger win than the reordering would have given.

**Merging the two lookups into one window was tried and reverted.** The window is keyed `(consumer, function)`, not `(consumer, function, dep_index)`, so on a function with flapping unresolved siblings a dep slot's own last *resolved* event gets evicted — and then neither `flipped` nor `unresolvedToResolved` fires, so a producer flip goes unrecorded. Two typed queries keep the exact prior semantics while preserving essentially all of the reduction. `TestAudit_SiblingUnresolvedFloodDoesNotHideFlip` covers it.

**The suggested index doesn't match the schema.** #1582 proposed `(agent_id, event_type, created_at)`; `registry_events` has no `agent_id` (the agent is an edge, FK column `agent_events`) and no `created_at` (it's `timestamp`). Actual index is `("function_name", "timestamp")` + the agent edge.

## Tests

`go test ./... -race -count=1`: **929 PASS / 0 FAIL / 10 SKIP** (was 924; skips are pre-existing env gates). Every new test was verified to fail against the pre-fix code by reverting each hunk individually and restoring — including the audit one, which fails under the merged-window design with `"[]" should have 1 item(s), but has 0`.

The race tests inject the competing transition via an Ent mutation hook so the row is terminal at the instant the guarded UPDATE evaluates its WHERE, which is what a READ COMMITTED re-read sees. What that does **not** cover: two genuinely concurrent Postgres transactions (lock waits, commit ordering). The test backend is SQLite, where an out-of-band commit is invisible to an open transaction, so same-transaction injection is the only faithful model available. Documented in the test file.

## Known residual

`UpdateHeartbeat`'s status decision now happens inside the transaction, which closes the window where a health-monitor flip to `unhealthy` would go uncorrected by the heartbeat that proves the agent is alive — in this system that means transient de-wiring of a live provider, not just a stale field. It shrinks the window rather than eliminating it: under READ COMMITTED the monitor can still commit between the in-tx read and the in-tx write. That is the same residual gap `RegisterAgent` has always had; closing it fully needs `SELECT ... FOR UPDATE`, which is not in this PR.

Also note `UpdateHeartbeat` is query-*neutral* in the steady state (our read replaces the hook's). The POST-path saving is in `RegisterAgent`, where the row was already read inside the transaction.

## Also

Removes the dead `develop` entry from the CI triggers — this repo has `main` and `dev`, and `origin/develop` does not exist. Dead-config removal, no behavior change.

Closes #1580
Closes #1581
Closes #1582

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved heartbeat error handling: temporary database failures now return 503, while unknown agents continue to return 410.
  - Prevented unnecessary agent status updates and ensured unhealthy agents recover reliably.
  - Strengthened job state transitions to prevent concurrent operations from overwriting terminal states.
  - Improved audit history lookups for functions with multiple dependencies.

- **Chores**
  - CI workflows now run only for pushes and pull requests targeting `main`.

- **Tests**
  - Added regression coverage for heartbeat handling, audit lookups, and concurrent job-state races.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI/CD Pipeline
 
 on:
   push:
-    branches: [main, develop]
+    branches: [main]
   pull_request:
-    branches: [main, develop]
+    branches: [main]
   workflow_dispatch:
 
 env:
--- a/src/core/ent/migrate/schema.go
+++ b/src/core/ent/migrate/schema.go
@@ -424,6 +424,11 @@ var (
 				Unique:  false,
 				Columns: []*schema.Column{RegistryEventsColumns[1]},
 			},
+			{
+				Name:    "registryevent_function_name_timestamp_agent_events",
+				Unique:  false,
+				Columns: []*schema.Column{RegistryEventsColumns[2], RegistryEventsColumns[3], RegistryEventsColumns[5]},
+			},
 		},
 	}
 	// SchemaEntriesColumns holds the columns for the "schema_entries" table.
--- a/src/core/ent/schema/registryevent.go
+++ b/src/core/ent/schema/registryevent.go
@@ -48,5 +48,26 @@ func (RegistryEvent) Indexes() []ent.Index {
 	return []ent.Index{
 		index.Fields("timestamp"),
 		index.Fields("event_type"),
+		// Supports the dependency-audit prior-trace lookup, which runs on every
+		// full heartbeat that re-resolves dependencies (issue #1582):
+		//   WHERE function_name = ? AND event_type IN (...) AND <consumer agent>
+		//   ORDER BY timestamp DESC LIMIT 64
+		// Emitted column order is (function_name, timestamp, agent_events) —
+		// ent always places Fields() columns before Edges() columns, so the FK
+		// column behind the `agent` edge cannot lead.
+		//
+		// What the shape guarantees: function_name is an equality prefix and
+		// timestamp the ordered suffix, so the ORDER BY ... LIMIT is served by
+		// an index scan instead of a full scan plus sort. What it does NOT
+		// guarantee: event_type is not in the index (it is a recheck), and the
+		// consumer filter reaches the query as an EXISTS subquery rather than
+		// `agent_events = ?`, so the trailing FK column is not inherently an
+		// access predicate — how many index entries the LIMIT touches depends
+		// on the planner. (Postgres 16 was observed folding the semi-join
+		// equality into the index scan, making both function_name and
+		// agent_events access predicates; that is a planner behavior, not a
+		// property of the index.) Also narrows `meshctl audit` / GET /events
+		// when filtered by function name.
+		index.Fields("function_name", "timestamp").Edges("agent"),
 	}
 }
--- a/src/core/registry/audit_emit.go
+++ b/src/core/registry/audit_emit.go
@@ -17,10 +17,17 @@ import (
 )
 
 // priorTraceLookupLimit caps how many recent registry events the prior-trace
-// lookups scan when searching for a matching dep_index. Headroom for functions
+// lookup scans when searching for a matching dep_index. Headroom for functions
 // with many deps + concurrent writes; raise if you have functions with >64
 // deps. Future: push the dep_index filter into SQL via a JSON path expression
 // so we don't need to overscan in Go.
+//
+// The limit applies per event-type set, exactly as it did when each lookup
+// issued its own query (#1582 changed WHERE the queries run, not what they
+// select). Merging the two sets into one LIMIT-64 window was tried and
+// rejected: the window is scoped to (consumer, function), NOT to dep_index, so
+// unresolved events from sibling dep slots would evict a slot's own last
+// resolved event and silently swallow a chosen-producer flip.
 const priorTraceLookupLimit = 64
 
 // emitAuditEventIfInteresting writes a dependency_resolved or dependency_unresolved
@@ -47,6 +54,7 @@ func (s *EntService) emitAuditEventIfInteresting(
 	depIndex int,
 	trace *AuditTrace,
 	resolved *DependencyResolution,
+	history *auditHistoryCache,
 ) error {
 	if trace == nil {
 		return nil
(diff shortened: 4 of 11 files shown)


dhyansraj:
Both findings applied in 8e18d54dc, with one deviation and one correction each. Regression tests for both were verified to fail against the unfixed code.

**Audit memo (`audit_emit.go`)** — valid, fixed in the cache rather than upstream. The duplicate-slot hazard is real: `ResolveAllDependenciesIndexed` (`resolver.go:33-63`) takes `function_name` from each entry of the agent-supplied `tools` list and does not deduplicate, so two tools sharing a name yield duplicate `(function, dep_index)` pairs and the second reads a pre-emission snapshot the pre-#1582 code would have re-queried past.

The suggestion was to enforce uniqueness in `ResolveAllDependenciesIndexed`. That function is the single source of truth for dependency resolution *and* storage, so changing what it returns would alter resolution semantics well beyond this audit path — a much larger blast radius than the defect warrants. Instead the cache now tracks which slots have emitted during the call and bypasses the memo only for those. The ordinary distinct-`dep_index` case stays at 2 queries per function; only a genuinely repeated slot re-queries.

`TestAudit_DuplicateFunctionNameSlotSeesOwnEmission` fails without it with `should have 2 item(s), but has 1` — the second slot's flip silently dropped.

**`ReleaseJob` claim epoch** — predicate added, rationale corrected. The comment does not say what the review suggested, because that rationale does not hold: `ReleaseJob(ctx, jobID, instanceID, reason)` takes no epoch (`ent_service_jobs.go:1072`), and its only caller passes just the instance id and reason (`ent_handlers_jobs.go:521`). `current.ClaimEpoch` is read fresh inside the transaction, so a genuinely *delayed* release re-reads and matches whatever epoch is current. Preventing that would require the client to send the epoch it holds and the server to reject a mismatch — a wire/protocol change, not a `WHERE` clause.

What the predicate does fix is the same read→write window the rest of this PR addresses: the read takes no row lock, `ClaimNextJob` does `AddClaimEpoch(1)`, and a reclaim plus re-claim by the *same* instance leaves `owner_instance_id` identical — so the existing owner predicate cannot tell the two claim generations apart. That is the framing in the code comment.

The existing `Affected == 0` classification already handles it correctly: on an epoch mismatch the row is non-terminal and still owned by the caller, so it falls through to `ErrJobNotOwner` → 403. No logic change, only an annotation of why that branch is right.

`TestReleaseJob_LosesToConcurrentReclaimBySameInstance` fails without the predicate with `An error is expected but got nil` — the release succeeded and cleared a freshly granted claim.

**Out of scope, flagged for a separate issue**: the same duplicate `function_name` payload also makes `StoreDependencyResolutions` write two `DependencyResolution` rows for one `(consumer, function, dep_index)`. Pre-existing, untouched here.

`go test ./... -race -count=1` green.

Issue #1606, opened 2026-09-06
perf(registry): long-polls do not observe shutdown, so graceful stop always waits out its full budget


dhyansraj:
## Summary

`listJobEventsCore` parks a long-poll for up to 60s (`listJobEventsMaxWait`) without selecting on any drain signal. `http.Server.Shutdown` counts a parked long-poll as an in-flight request and does not cancel its request context, so a registry with any jobs consumer attached cannot drain until either the consumer's wait elapses or the shutdown deadline expires.

The practical effect: graceful stop burns its entire HTTP budget on every restart where a consumer is connected, even though nothing is actually in progress. The same applies to `/proxy/*` SSE relays, which are unbounded and can never be waited out.

## Why it matters now

#1583's PR implements `Shutdown(ctx)` for the registry, which previously had `// TODO: Implement graceful shutdown for HTTP server` and never shut the listener at all. Sizing that budget ran straight into this: the window had to be tuned *down* rather than up, because a longer one would make every `meshctl stop` sit out the full duration and then get SIGKILLed at meshctl's 10s default.

So the shutdown budget in that PR is a workaround for this issue, not an independent choice. Fixing this would let the registry exit as soon as real work finishes, and would let the budget be sized for actual in-flight requests rather than for parked ones.

## Fix

Give `listJobEventsCore` a drain signal to select on alongside its existing wait and cancellation, so a shutdown wakes parked callers and returns them an empty result promptly. The `Server` already carries `shutdownCtx` for handler-spawned goroutines (`server.go`, added for the cancel-forward path); the same context is the natural source, but the jobs service does not currently receive it.

**Care needed**: waking a parked long-poll must not look like a lease event or disturb claim epochs. This touches the #1252 lease semantics — a consumer that gets an early empty return must re-poll cleanly without losing its cursor or having its claim treated as abandoned. That coupling is why it was deferred rather than folded into #1583.

Consider whether `/proxy/*` SSE relays should also be terminated on drain, or deliberately left to be cut by process exit.

## Test

Park a long-poll, call `Stop`, and assert shutdown completes well inside the budget rather than at the deadline — and that the parked caller returns a clean empty result rather than an error, and can resume from its cursor.

Found during the #1583 review.

Issue #1604, opened 2026-09-06
fix(registry): IPv6 agent hosts produce unparseable dependency endpoints, so consumers cannot reach them


dhyansraj:
## Summary

`src/core/registry/resolver.go:503` builds the endpoint handed to consumers with:

```go
endpoint = fmt.Sprintf("%s://%s:%d", scheme, winner.HttpHost, winner.HttpPort)
```

When `HttpHost` is an IPv6 literal this yields `http://::1:8080`, which is not a parseable URL. That endpoint is what dependency resolution returns to every consumer, so an IPv6-addressed provider cannot be wired at all — the consumer receives a target it cannot dial.

## Scope beyond the proxy fix

#1583 identified the same class of defect on the registry's proxy path and it is fixed in that PR via `net.SplitHostPort` / `net.JoinHostPort`. That fix is proxy-local. The resolver path is separate, is on the main DI hot path, and is arguably the more consequential of the two: the proxy is an operator convenience, while dependency resolution is how the mesh wires itself.

The correct construction is `net.JoinHostPort(host, strconv.Itoa(port))`, which brackets IPv6 literals.

## What to check while fixing

- Audit every place an endpoint is assembled from a host and port, not just this one. `HttpHost` originates from agent registration, so the same literal flows through the registry, the meshui trace views, and anything that persists or displays an endpoint.
- Confirm what agents actually send as `http_host` when bound to an IPv6 address or dual-stack — it is possible no runtime currently emits a bare IPv6 literal, in which case this is latent rather than live. Establish that before sizing the fix; the audit issues in this batch have repeatedly overstated reach.
- `src/core/cli/utils.go:114` has the same `"%s:%d"` → `net.Dial` shape on the meshctl side and is currently the only `go vet` finding left in the repo.

## Test

Register an agent with an IPv6 `http_host`, resolve a dependency on it, and assert the returned endpoint parses and dials. An end-to-end case through the resolver rather than a unit test on the formatting, since the point is that consumers get something usable.

Found during the #1583 review.

PR #1607, opened 2026-09-06
fix(registry): harden HTTP listeners, bound request bodies, implement graceful shutdown

### LINKED ISSUE #1583: security(registry): HTTP listeners have no timeouts or body limits; admin port serves /admin/* plaintext and unauthenticated
## Summary
- `server.go:203,424-430,522-528` construct `http.Server` without `ReadHeaderTimeout`/`IdleTimeout`, and no handler caps request bodies (`ent_handlers.go:201,796` bind JSON directly). A slow-header client or a large POST holds a connection or memory indefinitely.
- `MCP_MESH_ADMIN_PORT` (`server.go:298-304,512-528`) moves `/admin/*` (rotate, entities, drain) off the mTLS-verified engine onto a second listener that is plaintext with no auth. `man security.md:194-201` describes this as "isolation"; in a cluster it is the opposite unless a NetworkPolicy fronts it.
- `Server.Stop` never shuts the HTTP listener (`server.go:257`, TODO), so `main.go:153-158` `os.Exit(0)` cuts in-flight proxy streams and long-polls after draining cancel-forwards.
- `ent_handlers.go:1177-1185` splits proxy targets on `:`; IPv6-literal hosts can never be proxied. `src/core/ui/proxy.go:55-68` truncates registry bodies at 10 MB and returns the truncated JSON with the upstream status.

## Fix
Set `ReadHeaderTimeout`, `IdleTimeout`, `MaxHeaderBytes` and wrap bodies in `http.MaxBytesReader`; either apply the same TLS/trust engine to the admin listener or document that it must be network-isolated; implement `Shutdown(ctx)`; use `net.SplitHostPort`; return 502 on truncation.

CONFIRMED for the timeouts/body/IPv6/truncation; PLAUSIBLE for the admin-port exposure (depends on deployment).

---

dhyansraj:
Issue #1583 — registry HTTP hardening.

## Timeouts on every listener

`server.go:424` was the only explicit `http.Server`; the plaintext and admin paths used gin's `Run`, which builds a server internally with **no timeouts at all**. All three now go through one helper.

`ReadHeaderTimeout` 10s, `IdleTimeout` 120s (must exceed the heartbeat cadence or every heartbeat pays a fresh TCP+TLS handshake), `MaxHeaderBytes` 64KB (Go's default is 1MB).

**`ReadTimeout` and `WriteTimeout` are deliberately unset, and a test asserts they stay zero.** Both are absolute deadlines. The registry serves a 60s job-event long-poll (`listJobEventsMaxWait`) and relays unbounded `/proxy/*` SSE — any value severs one of them. This is the kind of thing a future reader will "fix", hence the test.

## Request body cap

`MCP_MESH_MAX_REQUEST_BODY_BYTES`, default 10MB, `0` disables.

The default is measured rather than picked. A `MeshToolRegistration` carrying all four schema blobs (`InputSchema`, `InputSchemaCanonical`, `OutputSchema`, `OutputSchemaCanonical`) plus kwargs, dependencies, tags and warnings marshals to **~9.8KB/tool**, so 1000 fully-schema'd tools ≈ 9.8MB. A test pins a 100-tool payload as acceptable so the default cannot be tightened silently later.

413 is produced two ways: a declared `Content-Length` over the cap is refused before a byte is read; an undeclared/chunked body trips `http.MaxBytesReader` and surfaces as `*http.MaxBytesError`. Malformed JSON still yields a byte-identical 400. All **7** unguarded `ShouldBindJSON` sites were converted — the issue named 2, and 6 of the 7 are in a file it never mentions.

## Graceful shutdown

`Server.Stop` carried `// TODO: Implement graceful shutdown for HTTP server` and never shut the listener, so `os.Exit(0)` cut in-flight proxy streams and long-polls. `Shutdown(ctx)` now drains every registered listener, and `registerServer` refuses to start one after `Stop` has run, closing the signal-before-listen race.

`main.go` waits on a **bounded** select rather than exiting immediately. An unbounded wait would have been a regression: `Stop`'s tail (`tracingManager.Stop()`) has no deadline of its own, and `signal.Notify` stays installed, so a wedged Redis consumer would leave a process that ignores further SIGTERM.

A deadline expiry with long-polls still attached is the *expected* outcome on a busy registry — it is logged as such rather than returned, so a routine restart no longer prints `Error during server shutdown`. See #1606 for why the budget had to be tuned down rather than up.

## Admin listener — opt-in, default unchanged

Timeouts and the body cap apply unconditionally. TLS and `TLSVerifyMiddleware` are **opt-in** via `MCP_MESH_ADMIN_TLS`, default off.

Applying them unconditionally was implemented and then reverted, because it breaks two things:

- It flips the port from `http://` to `https://`, breaking every existing admin client and `tc11_admin_port_isolation`, which asserts `curl http://localhost:8001/admin/rotate` → 200 under `MCP_MESH_TLS_MODE=auto`.
- In `strict` it returns 403 to `meshctl`, which has **no client-certificate support** anywhere in `src/core/cli` (the only TLS knob is `--insecure`). That would remove `meshctl registry drain|resume|status` and `meshctl entity rotate` — the documented drain-before-upgrade path — in exactly the configuration `security.md`'s production checklist recommends.

Both traps are now stated in the man page, so anyone enabling the flag learns them before they trip over them. A Go test asserts plaintext-by-default *on a TLS-configured registry*, so tc11 cannot be broken silently.

Note this leaves the exposure open by default. That is deliberate — it matches the project's strictness-opt-in stance — but the admin port genuinely is unauthenticated, which the docs now say plainly instead of calling it "isolation".

## Also

- `net.SplitHostPort` / `net.JoinHostPort` on the proxy target path. The issue named the split; the same line also built the dial host with `fmt.Sprintf("%s:%d")`, which yields `::1:8080` and an unparseable URL even after the split is fixed. Both halves are needed. (The resolver has the same defect on the DI path — filed as #1604, not fixed here.)
- `src/core/ui/proxy.go` returns 502 on truncation instead of forwarding a body cut mid-token with the upstream status.
- `detect_endpoints.py` now skips test files. It walked `*_test.go` and demanded that fixture routes on throwaway engines appear in the OpenAPI contract, which blocked this change. The detector governs production handlers; verified it still fails on a real unspecified production route.

## Corrections to #1583

Four of its claims did not survive verification: only one of the three cited `http.Server` sites exists; `ent_handlers.go:796` is a 403 block, not a bind; there are 7 bind sites, not 2; and the IPv6 line reference is off by ~18 lines and misses half the bug. The conclusions all hold — the locations and counts did not.

## Tests

`go test ./... -race -count=1` green across 11 packages. New: `http_limits_test.go`, `server_listener_test.go`, `proxy_ipv6_test.go`, `src/core/ui/proxy_test.go`. Man goldens 1796 → 1809 with the arithmetic annotated; `go test ./src/core/cli/man/` passes and `check_doc_claims.py` is clean.

Closes #1583

## Summary by CodeRabbit

- **New Features**
  - Added configurable registry request-body limits, with clear errors for oversized requests.
  - Added optional TLS and client-certificate enforcement for the admin listener.
  - Added support for IPv6 registry targets.
  - Added response-size protection for UI-to-registry proxy requests.

- **Bug Fixes**
  - Registry shutdown now drains active requests and completes cleanup before exiting.
  - Improved handling of invalid configuration values and malformed JSON requests.

- **Documentation**
  - Updated deployment, security, and environment-variable guidance for admin TLS and request limits.

### DIFF
--- a/cmd/mcp-mesh-registry/main.go
+++ b/cmd/mcp-mesh-registry/main.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	flag "github.com/spf13/pflag"
 
@@ -19,6 +20,14 @@ import (
 
 // version is injected at build time via ldflags
 var version = "dev"
+
+// shutdownWaitTimeout bounds how long main waits for Server.Stop after
+// the listeners have closed. Stop's own bounded phases are 5s of HTTP
+// drain plus 10s of handler-goroutine drain, so 20s leaves both room to
+// finish and still guarantees the process exits: Stop's tail (tracing
+// shutdown) has no deadline of its own, and by this point the SIGTERM
+// has already been consumed.
+const shutdownWaitTimeout = 20 * time.Second
 
 func main() {
 	// Command line flags
@@ -123,12 +132,13 @@ func main() {
 		TlsKeyFile:              os.Getenv("MCP_MESH_TLS_KEY"),
 		TrustDir:                os.Getenv("MCP_MESH_TRUST_DIR"),
 		AdminPort:               getEnvIntDefault("MCP_MESH_ADMIN_PORT", 0),
+		AdminTLS:                getEnvBoolDefault("MCP_MESH_ADMIN_TLS", false),
 	}
 
 	if registryConfig.TlsMode != "off" {
 		appLogger.Info("🔒 TLS configuration: mode=%s, backend=%s", registryConfig.TlsMode, registryConfig.TrustBackend)
 		if registryConfig.AdminPort > 0 {
-			appLogger.Info("🔒 Admin API port: %d", registryConfig.AdminPort)
+			appLogger.Info("🔒 Admin API port: %d (TLS: %t)", registryConfig.AdminPort, registryConfig.AdminTLS)
 		}
 	}
 
@@ -141,7 +151,15 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Setup graceful shutdown
+	// Setup graceful shutdown.
+	//
+	// shutdownDone is what makes the drain observable to the main
+	// goroutine: Server.Stop now closes the HTTP listeners and waits for
+	// in-flight requests, and Run returns as soon as they are closed. The
+	// old os.Exit(0) in here raced that wait — the process could exit
+	// while requests were still draining. Main blocks on shutdownDone
+	// instead, so the deferred db.Close also runs (issue #1583).
+	shutdownDone := make(chan struct{})
 	go func() {
 		sigChan := make(chan os.Signal, 1)
 		signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
@@ -154,8 +172,7 @@ func main() {
 			appLogger.Error("Error during server shutdown: %v", err)
 		}
 
-		appLogger.Info("✅ Registry service stopped")
-		os.Exit(0)
+		close(shutdownDone)
 	}()
 
 	// Start server
@@ -165,6 +182,23 @@ func main() {
 		appLogger.Error("❌ Failed to start server: %v", err)
 		os.Exit(1)
 	}
+
+	// Run returned without error, which only happens once the listeners
+	// have been shut down. Wait for the rest of Stop (background-goroutine
+	// drain, tracing flush) before returning — but bounded. Stop's own
+	// phases are capped, yet its tail is not: tracingManager.Stop() waits
+	// on the trace-trim WaitGroup and the Redis consumer with no deadline,
+	// so a wedged Redis would otherwise leave a process that has already
+	// consumed its SIGTERM and needs a SIGKILL to die. The bound sits
+	// above Stop's bounded phases (HTTP drain + goroutine drain) so it
+	// never cuts a drain that is still making progress.
+	select {
+	case <-shutdownDone:
+		appLogger.Info("✅ Registry service stopped")
+	case <-time.After(shutdownWaitTimeout):
+		appLogger.Warning("Shutdown did not finish within %s; exiting anyway", shutdownWaitTimeout)
+		os.Exit(0)
+	}
 }
 
 func getEnvDefault(key, defaultValue string) string {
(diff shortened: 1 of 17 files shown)


dhyansraj:
Both applied in 016bc8dc0.

**Stale admin-port comment** (`docs/environment-variables.md:242`) — valid, and it was a leftover from the unconditional admin hardening that was implemented and then reverted in this PR. The line claimed the admin port had "the same TLS + client-cert policy as the main port", which is false now that the default is plaintext and unauthenticated.

Worth noting the failure mode was slightly worse than "wrong about TLS": it implied the admin port tracks `MCP_MESH_TLS_MODE`, so an operator running `strict` would reasonably conclude the port was already cert-gated and skip the NetworkPolicy. The replacement says "plaintext and unauthenticated by default, whatever `MCP_MESH_TLS_MODE` is set to" and points at `MCP_MESH_ADMIN_TLS`.

I audited every other surface describing the admin port rather than patching only the line the review named, since this wording had already survived one rework:

| Surface | Verdict |
|---|---|
| `docs/environment-variables.md:242` | stale — fixed |
| `docs/environment-variables.md:219` (`MCP_MESH_ADMIN_TLS`) | correct |
| `man environment.md:502`, `:615` | correct — `:502` never claimed TLS inheritance, so it survived the rework right rather than stale |
| `man security.md:194-217, 365` | correct — re-verified against `newAdminEngine`/`adminTLSEnabled` (`server.go:686-736`) rather than assumed, including the auto-passes/strict-403 claim and the https scheme warning |
| `docs/deployment.md:156` | correct |
| `docs/upgrading.md:155`, `man upgrading.md:187`, `man registry.md:176`, `cli/registry.go:31,261` | correct — `http://` matches the shipped default |
| `helm/.../values.yaml:257`, `docs/concepts/registry.md:200` | no transport claim |

No man content changed, so the goldens did not move.

**`MinVersion` on `serverTLSConfig`** — added, `ClientAuth` and `Certificates` untouched. Both listeners pick it up (`serverTLSConfig` has exactly two callers: `runWithTLS` and `startAdminServer`).

To be accurate about what this does: Go 1.25 already defaults servers to TLS 1.2, so **behaviour is unchanged today**. The reasons to pin it anyway are that the server default is reversible at runtime via `GODEBUG=tls10server=1` and an explicit `MinVersion` is not, and that `tlsutil.go:32` and `cli/utils.go:71` already pin it — this was the inconsistent one. The code comment says that rather than implying a vulnerability was closed.

`go test ./... -race -count=1` green; pre-commit clean including `Detect Unauthorized Endpoints`.

**Out of scope, flagged**: `MCP_MESH_ADMIN_TLS` has no Helm surface. `helm/mcp-mesh-registry/templates/configmap.yaml:91-93` emits only `MCP_MESH_ADMIN_PORT`, and `values.yaml` has `adminPort` but no `adminTLS`, so chart users cannot enable the hardening without a raw env override.

PR #1612, opened 2026-09-07
fix(core): preserve nullability through $ref promotion, bound $defs inlining, compare the whole tool spec

### LINKED ISSUE #1587: fix(core): schema normalizer edge cases (nullable $ref promotion, camelCase collisions, inline_refs growth)
## Summary
`src/runtime/core/src/schema_normalize.rs` is the single canonicalizer all three SDKs rely on for `expected_type` matching, so divergence here silently mis-resolves dependencies.

- `:685-719` nullable promotion drops `null` when the single non-null branch is a kept `$ref` (cyclic) or a nested `anyOf`: `Optional[RecursiveNode]` hashes as `RecursiveNode`, and `Optional[Union[A,B]]` canonicalizes differently from Pydantic vs Zod input. PLAUSIBLE.
- `:504-513` camelCase renaming collides (`value`/`Value`, `user_id`/`userID`) and silently overwrites one property with no warning. CONFIRMED.
- `:367-432` `inline_refs` re-inlines shared defs per use site with no size/depth guard; a diamond `$defs` graph expands exponentially. PLAUSIBLE.

## Fix
Preserve nullability when the non-null branch is a ref or composite; emit a WARN verdict on rename collision (the verdict tier exists for exactly this); memoize inlined defs or cap depth with a BLOCK verdict.

---

### LINKED ISSUE #1588: fix(core): tools_are_different ignores tags, schemas, kwargs and filters, so those updates never reach the registry
## Summary
`runtime.rs:517-550` decides whether an `UpdateTools` command changes the registration by comparing a subset of fields. Tags, description, input/output schemas, `kwargs`, `llm_filter` and `match_mode` are not compared. An update that changes only those is dropped as "unchanged" and the registry keeps the stale registration until the process restarts.

Related: `HandleState.dependencies` (`runtime.rs:748-761`, `handle.rs:21-22`) is keyed by capability while the topology is keyed by `(function, index)`; removing one of two same-capability dependencies deletes the key and `get_dependencies()` misreports.

## Fix
Compare a full canonical hash of the tool spec; key handle state the way the topology is keyed.

PLAUSIBLE (source read; no runtime today calls UpdateTools with a tags-only change, but the surface is public).

---

dhyansraj:
Issues #1587 and #1588. Both were rated PLAUSIBLE, so every item here has a test that was verified to fail against the unfixed code.

`schema_normalize.rs` is the single canonicalizer all three SDKs use for `expected_type` matching, so a shape that canonicalizes wrongly is a *silent mis-resolution*, not an error.

## ⚠️ Canonical hashes change — read before merging

Four shapes now hash differently, all of which were previously canonicalized **wrongly**:

| shape | before | after |
|---|---|---|
| `Optional[RecursiveNode]` | hashed identically to `RecursiveNode` | distinct |
| nested `Optional[Union[A,B]]` (Zod-shaped) | dropped the null | keeps it |
| Pydantic `Optional[Literal[...]]` | dropped the null | keeps it |
| any nested union spelled `oneOf` | Java's flat form never matched py/ts | all three converge |

During a mixed-version rollout an old-core consumer and a new-core producer will not match on these; they resolve once both sides upgrade. Since they previously matched *incorrectly* (`Optional[X]` ≡ `X`), the fix is right either way — but the window is real.

**Two new verdicts can refuse startup**: a camelCase property collision emits WARN (a refusal under `MCP_MESH_SCHEMA_STRICT=true`), and a `$defs` graph exceeding `MCP_MESH_SCHEMA_MAX_INLINED_NODES` is a BLOCK. Both are documented in `docs/concepts/schema-matching.md` and the man page.

## #1587a — nullability dropped through `$ref` promotion

Promotion only appended `"null"` when the promoted branch carried a `type`. A `$ref` branch has none. A fourth trigger not in the issue: Pydantic omits `type` on enum branches, so `Optional[Literal[...]]` loses it the same way and diverges from Zod.

Those branches now keep an explicit null wrapper rather than getting a `type` sibling hung next to a `$ref`, which pre-2019-09 drafts ignore.

### The flattening had to move — it broke a pair that matched before

Nested unions are flattened so Zod and Pydantic spellings converge, but doing it at the top of `normalize_nullable` (before the metadata strip at `:543` and the `oneOf`→`anyOf` rewrite at `:560`) **split a pair that matched at HEAD**. Pydantic's `{discriminator, oneOf}` branch has two keys and would not splice; Zod's sole-key `anyOf` would:

```
Optional[Pet]   py      ts
HEAD            0984245e  0984245e   ← matched (both wrongly dropping null)
first attempt   bad443fd  7a217923   ← split
now             7a217923  7a217923   ← match, and Java's flat form joins them
```

Flattening now runs after both spellings converge, asking whether a branch's *sole remaining* key — ignoring exactly the keys the normalizer strips anyway, via a shared `STRIPPED_METADATA_KEYS` — is `anyOf` or `oneOf`. Sharing that const is deliberate: "is this a bare nested union?" must ask the same question the strip rule answers, or they drift apart again.

**Fixtures added** for both shapes × three spellings. The corpus contained no nested union, which is exactly why 36 unchanged hashes looked reassuring.

## #1587c — `$defs` inlining had no bound

Measured before choosing anything (diamond graph, each level referencing the one below twice):

```
levels= 4  input=  514 B  output=    1,902 B  (1.3ms)
levels= 8  input=  874 B  output=   30,462 B  (25ms)
levels=12  input=1,242 B  output=  487,422 B  (491ms)
levels=16  input=1,614 B  output=7,798,782 B  (9.5s)
```

**The issue's suggestion to memoize would not have fixed this.** Memoization removes recomputation, but each use site still gets its own cloned subtree, so the 7.8MB output stands. Only a bound helps.

The budget counts **materialized nodes, not `$ref` sites** — a site count says nothing about output size, since each site inlines a body of unbounded size. It is overridable via `MCP_MESH_SCHEMA_MAX_INLINED_NODES`, because a consumer-side BLOCK on `expected_type` has **no per-tool escape hatch** (`rust_heartbeat.py:214` hardcodes `tool_strict=True`, with a comment saying so). An unescapable refusal calibrated against 12 toy fixtures would have failed realistic Pydantic models: a measured 40-sub-model × 50-shared-enum domain model is 2,040 `$ref` sites but only 27,205 nodes / 198KB, and the default admits it at ~18× headroom.

## #1588 — `tools_are_different` compared a hand-picked subset

Instrumenting the comparison found **15 of 22 fields uncompared**, not the 7 the issue names — including every schema field and `match_mode`, which a comment conceded was "intentionally left out".

It now compares the serialized spec, derived from the struct so a newly added field cannot silently go uncompared again. That recurrence is the defect, not the current misses.

**Reach is narrow and the PR is sized accordingly.** Python has no `update_tools` caller; Java's `MeshRuntime.updateTools` is a logging no-op ("this would require adding a mesh_update_tools FFI function"); the one live TS path (`api-runtime.ts:726`) hardcodes `capability:""`, `version:"1.0.0"`, `tags:[]`, `description:""`, `inputSchema:undefined`, so every uncompared field is a constant there. This is public-API hardening, not a live bug.

Also: `HandleState.dependencies` was keyed by capability while the topology is keyed by `(function, index)`, so dropping one of two same-capability dependencies deleted the key and `get_dependencies()` misreported. Re-keyed; the documented capability→endpoint shape is preserved as a projection with deterministic tie-breaking (the old behaviour was nondeterministic).

## Tests

All four Rust legs, uniformly **+33**: 541 / 569 / 556 / 545 (HEAD: 508/536/523/512). `cargo check --all-targets` and clippy clean on all five legs.

Grafted onto a pristine `d8dc0ec17` worktree, **18 of the new tests fail** — including both cross-runtime parity cases, the budget override, the warning dedup and the diamond BLOCK. The ones that pass at HEAD are deliberate negative controls, plus the two `java_hash_matches_fixture` cases, since Java's flat spelling was already producing the correct form.

Man goldens moved +9/+9/+2, matching the added spans exactly, annotated in the same commit.

## Not fixed, deliberately

`api-runtime.ts:59-82` drops the per-dependency `expectedSchema*`/`matchMode` fields that `express.ts:87-125` populates, so `@mesh.route` schema matching is inert on that path. Unrelated to both issues; worth its own issue.

Closes #1587
Closes #1588

## Summary by CodeRabbit

- **New Features**
  - Added bounded schema normalization for large or deeply reused schemas, returning a clear BLOCK verdict instead of exhausting resources.
  - Improved handling of nested unions, nullable schemas, cyclic references, and camelCase property collisions.
  - Dependency reporting now preserves duplicate capability resolutions consistently.

- **Bug Fixes**
  - Smart-diff detection now recognizes changes across all tool metadata and schema fields.
  - Fixed dependency updates where removing one duplicate capability could remove another.

- **Documentation**
  - Documented schema verdict limits, consumer-side BLOCK resolution, and known schema-matching limitations.


dhyansraj:
Applied in `e6c863473`, with one addition.

`parse_node_budget(Option<&str>)` is extracted; `resolved_node_budget()` is now a thin wrapper that reads the env var and delegates. `node_budget_env_parsing` tests the pure function directly — no `set_var`, no restore block.

**Added beyond the suggestion**: extracting the parser alone would have silently dropped coverage of the env plumbing. Nothing would then verify the variable is actually *read* — deleting the `std::env::var` call in `resolved_node_budget` would leave every test green. So there is one small `node_budget_env_is_wired_up` test that keeps that wired, using a `Drop`-based guard so the restore survives an unwinding assertion. Verified it bites: with the env read replaced by `parse_node_budget(None)` it fails with `left: 500000, right: 4321`.

That is now the only place in this file touching global state, down from eight assertion points ahead of a trailing restore.

**On the rationale** — worth being accurate, since the original test was not actually racy: the crate is edition 2021 so `set_var` is not `unsafe` here, CI runs `--test-threads=1`, and `config.rs:550-627` already mutates env across a dozen of its own tests. The real hazard was narrower: any of the eight assertions failing would skip the restore and leak `MCP_MESH_SCHEMA_MAX_INLINED_NODES` into every subsequent test in the process, turning one failure into confusing downstream ones.

`config.rs` and the other env-mutating tests are deliberately untouched — same pattern, but unrelated scope for a schema-normalizer PR.

All four legs +1 (542 / 570 / 557 / 546); `cargo check --all-targets` and clippy clean, with the only `schema_normalize.rs` clippy hit being a pre-existing collapsible `if let` also present on `469d069e9`.

dhyansraj:
Applied in `1099e296f`, taking the second option rather than the first.

**The race is real, and worse than described.** `resolved_node_budget()` is called at `schema_normalize.rs:236` inside `normalize_schema`, so *every* schema test reads that env var. `default_budget_admits_a_realistic_pydantic_model` normalizes a ~27,205-node model and asserts `OK`; if it observes the wiring test's `4321` it BLOCKs and fails — surfacing the failure in an innocent test rather than the one that caused it. CI is safe (all four legs run `--test-threads=1`), but plain local `cargo test` is parallel by default. No "safe" value exists either: below 27,205 breaks the realistic-model test, above 500,000 masks `diamond_defs_graph_blocks_instead_of_exploding`.

**I did not delete the test.** It is the only thing verifying the env var is actually read, and `MCP_MESH_SCHEMA_MAX_INLINED_NODES` is the operator's sole escape from a consumer-side BLOCK — `rust_heartbeat.py:214` hardcodes `tool_strict=True`, so there is no per-tool override. An unescapable startup refusal was one of the two blockers this PR was sent back for; if a refactor drops the env read, that regression returns silently. Removing the safety net to fix a test-isolation problem is the wrong trade.

Instead the test re-execs its own binary with `--exact`, passing the marker and the budget var through `Command::env`. The parent never calls `set_var`, so nothing is ever visible to a sibling test; `EnvGuard` is gone. The child's stdout/stderr are embedded in the parent's panic message, and the comment names the concrete race and its victim so nobody simplifies it back.

Mutation-checked — stubbing the env read to `parse_node_budget(None)` fails through the parent with the child's assertion intact (`left: 500000, right: 4321`).

Verified under the configuration that actually matters, plain parallel `cargo test`: `schema_normalize::` is 77/77 across repeated runs. Four legs unchanged at 542 / 570 / 557 / 546; `cargo check` and clippy clean.

**Unrelated pre-existing issue found while verifying**: `tracing_publish::tests::*` fails under default parallelism (3–4 failures, varying run to run — `publish_span_caps_stream_with_maxlen`, `publish_span_resolves_maxlen_once_at_init`, `reprober_recovers_from_cold_start`, `publisher_reuses_connection_across_publishes`). Same class of problem — a shared global singleton — and it keeps plain local `cargo test` red for this crate. Confirmed present on `e6c863473` before this change. Filing separately; not touched here.

PR #1613, opened 2026-09-08
fix: X-Mesh-Job-Id is inbound-only in every runtime

### LINKED ISSUE #1570: fix: X-Mesh-Job-Id allowlisting and push-mode dispatch disagree across runtimes
## Summary
The push-mode dispatch protocol header `x-mesh-job-id` is handled three different ways.

- **Java allowlists it unconditionally.** `TraceContext.java:66-68` adds `x-mesh-job-id` to the propagate list, with a comment claiming this matches Python/TS. Python deliberately excludes it (`_mcp_mesh/tracing/context.py:15-31`: "x-mesh-job-id doubles as the dispatch discriminator, so it CANNOT also carry calling identity"). Consequence: a Java `task=true` handler calling a downstream Java `task=true` tool forwards its own job id, and `MeshToolWrapper.java:1085-1090` dispatches the downstream tool **as the caller's job**. This is the self-dispatch hazard Java's own comment warns about.
- **Java never seeds the calling claim epoch.** `TraceContext.java:45-87` omits `x-mesh-calling-claim-epoch`, which Python (`context.py`) and TS (`tracing.ts:22-45`) seed, so a Java executor's writes get owner-only fencing downstream, contradicting `MeshToolWrapper.java:1070-1078`.
- **Python/TS push-mode over HTTP is unreachable.** Python `engine/job_dispatch.py:87-92` reads the job id from the propagated map, but `http_wrapper.py:483-486, 424-430` filters it out before that point; TS `agent.ts:995` checks the raw map while `:1131` dispatches from the filtered one. `man jobs.md:460-462` (and the TS/Java variants) claim the header is on the default allowlist and name a non-existent `X-Mesh-Trace-Id`.
- **TS misclassifies a nested required-dep failure as a job dispatch.** `agent.ts:990-1026` + `claim-dispatcher.ts:551-562` + `proxy.ts:351-352`: the claim dispatcher seeds `x-mesh-job-id` into the propagated store the proxy forwards unfiltered; a nested call into a TS `task:true` tool whose required dependency is down releases the **caller's** lease and returns `""` instead of the `dependency_unavailable` refusal.

## Fix
Decide the contract once (Python's: job-id is never propagated; calling identity rides the dedicated `x-mesh-calling-*` pair), then align Java and TS, seed the epoch in Java, make the Py/TS dispatch gate read the same map, and correct the three jobs pages.

CONFIRMED for the Java allowlist and TS nested-call cases; PLAUSIBLE effect for the Py/TS unreachability.

---

dhyansraj:
Issue #1570. **Breaking for job chains in all three runtimes** — see the upgrading entry added in this PR.

`X-Mesh-Job-Id` is the push-mode dispatch discriminator: a producer reads it off the **incoming** request to bind its handler to a job row. Forwarding it makes a nested same-instance `task=true` call self-dispatch as the *caller's* job — owner and epoch match — and complete the caller's row with the callee's result, before the caller finishes.

## The issue's framing was wrong in both halves, and verifying it changed the fix

#1570 presents this as Java-only, with Python's contract as the model to adopt. Neither holds:

- **Python propagated it too.** `unified_mcp_proxy.py:1332` stamped `X-Mesh-Job-Id` on every outbound mesh→mesh call under an active job context. Python looked safe only because its inbound gate discarded the header — the same filter that made push-mode dispatch over HTTP unreachable there. **Implementing the issue as written — fix Python's inbound gate — would have activated the hazard in Python.** A third leak (`trace_context_helper` copying the propagated store) surfaced only because the first regression test still failed after the obvious one was fixed.
- **Java already seeds `x-mesh-calling-claim-epoch`**, and has since the audit base `4b4cb00f1`. The real epoch defect was inbound: the dispatch gate parsed it off the allowlist-filtered map where it was never allowlisted, so it was unconditionally `null` and every Java executor silently fell back to owner-only fencing.

The allowlist conflated two concerns — capture-for-dispatch and propagate-outbound. Those are now separate stores in each runtime: read raw inbound, scrub on the way out. Java's own source contained both halves of this bug twelve lines apart: `TraceContext.java:66-68` added the header claiming parity with Python/TS, while `:69-81` warned in detail about the self-dispatch hazard that addition creates.

## Rolling upgrades needed a guard, not a paragraph

A 3.7 caller still forwards the header, and a 3.8 callee now reads raw — so **old→new would reproduce the corruption this PR removes**, silently. Previously-safe pairs (old-Java → Python, old-TS → TS) become unsafe for the length of the skew window. Doc guidance can't carry that: "upgrade callers first" is unactionable in a mesh where every agent is both, and the failure mode is data corruption rather than an error.

Every pre-3.8 leak path emits the job id from inside a bound job context, and that context stamps `x-mesh-calling-job-id` on the same request (`McpHttpClient:466`, Python `_inject_trace_into_args`, `proxy.ts:370-374`). A genuine push dispatch cannot carry calling identity — the submitter isn't executing that job, and per-call headers are allowlist-filtered so user code can't set either name. The signal also survives a relay hop, since a 3.7 intermediary allowlisted both names together.

**So job id + calling job id ⇒ leak, not dispatch.** The callee ignores the job id, runs the call normally, and warns naming the fix.

Verified by deleting the guard and replaying a 3.7-shaped request through the real Python middleware:
```
assert seen["job"] == (None, None, None, None)
E  AssertionError: assert ('job-caller', 30.0, None, None) == (None, None, None, None)
```

**Residual, documented rather than hidden**: a pre-3.8 *non-`task`* Java tool that received a proxied dispatch relays the job id with no calling identity, since it never bound a job context. That is indistinguishable from a genuine proxied dispatch by any wire signal, predates this change, and needs a client-originated push into a non-task Java relay.

## Also

- **Registry proxy forwards `X-Mesh-Claim-Epoch`** alongside the job id (`ent_handlers.go:851-864`). It forwarded the id but not the epoch, so the one HTTP path carrying a dispatch job id split the pair and the newly-fixed raw epoch read still degraded to owner-only fencing.
- **All three runtimes now agree on the gate.** Python's originally kept the propagated store as a fallback, which meant widening `MCP_MESH_PROPAGATE_HEADERS` could re-arm self-dispatch there — Java's own test asserts the gate "does not consult this map at all".
- **Python's outbound scrub moved after the header merges**, matching Java and TS; it previously ran before, so a `custom_headers` or per-call `headers=` kwarg could put the trio back on the wire.
- Six stale comments asserting outbound propagation that no longer exists, and four "See Also" lines advertising it, are corrected. `docs/concepts/jobs.md` contradicted itself 900 lines apart.
- `X-Mesh-Trace-Id`, named in the old jobs pages, exists in no runtime. The real header is `X-Trace-ID`.

## Tests

Python full unit suite + `ruff check`/`format --check` clean; Go 10 packages `-race`; TypeScript 1458 across 97 files + `typecheck:all`; Java `mvn test` 924. Man goldens 1824→1843 inline, 532→531 list, 1662→1663 variant — derived and annotated, including why the delta shrank.

Regression tests (raw read, outbound scrub, e2e caller→callee replay through the real `ClaimDispatcher`, registry proxy) all fail against unfixed code. The skew-guard tests pin *new* behaviour and can't fail against the old tree, so that vector was proven the other way — by removing the guard from the fixed tree and watching a 3.7-shaped request dispatch.

Closes #1570

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented inbound job-dispatch headers from being forwarded on outbound calls, avoiding incorrect nested job dispatch.
  - Ensured timeout and caller identity continue to propagate through their dedicated headers.
  - Improved mixed-version compatibility by treating legacy header combinations as ordinary tool calls.
  - Preserved required dispatch headers across registry push requests while blocking receive cursors from crossing proxy boundaries.

- **Documentation**
  - Updated job cancellation, timeout propagation, audit references, and 3.8.0 upgrade guidance across runtime documentation.

### DIFF
--- a/docs/concepts/jobs.md
+++ b/docs/concepts/jobs.md
@@ -406,11 +406,20 @@ requests `max_duration=30`, the child gets 5 seconds. This is enforced
 at submission time so the child's runtime sees a single coherent
 deadline regardless of how deep the chain runs.
 
-The header is on the default `MCP_MESH_PROPAGATE_HEADERS` allowlist
-alongside `X-Mesh-Job-Id`, `X-Mesh-Trace-Id`, etc. — no per-agent
-configuration is needed for propagation to work. Tracing this in
+`X-Mesh-Timeout` is a baked-in default — it propagates with no
+`MCP_MESH_PROPAGATE_HEADERS` configuration, alongside the trace headers
+(`X-Trace-ID` / `X-Parent-Span`). Tracing this in
 [Audit Trail](audit.md) shows the deadline shrinking as it walks the
 chain.
+
+`X-Mesh-Job-Id` is deliberately **not** propagated. It is the inbound
+dispatch discriminator — the producer runtime reads it off the incoming
+request to bind the handler to that job row — so attaching it to an
+outbound call would make a nested `task=True` call dispatch as, and
+auto-complete, its caller's job. Which job invoked the current handler
+is a separate question, answered by the dedicated
+`X-Mesh-Calling-Job-Id` / `X-Mesh-Calling-Claim-Epoch` pair that
+`calling_job()` reads.
 
 ## meshctl
 
@@ -1300,8 +1309,8 @@ MeshJob event-channel variables in
 
 - [Streaming](streaming.md) — token-by-token progress for the request-
   response case where the work fits in a single SSE
-- [Audit Trail](audit.md) — `progressToken`, `X-Mesh-Job-Id`, and
-  `X-Mesh-Timeout` propagation through the audit pipeline
+- [Audit Trail](audit.md) — why the registry wired a job's `task=True`
+  provider to the consumer it did
 - [DDDI](dddi.md) — Distributed Dynamic Dependency Injection overview;
   explains how `MeshJob` slots are resolved
 - [Stateful Agents](stateful-agents.md) — the broader decomposition
--- a/docs/concepts/jobs.md
+++ b/docs/concepts/jobs.md
@@ -414,12 +414,14 @@ chain.
 
 `X-Mesh-Job-Id` is deliberately **not** propagated. It is the inbound
 dispatch discriminator — the producer runtime reads it off the incoming
-request to bind the handler to that job row — so attaching it to an
-outbound call would make a nested `task=True` call dispatch as, and
-auto-complete, its caller's job. Which job invoked the current handler
-is a separate question, answered by the dedicated
-`X-Mesh-Calling-Job-Id` / `X-Mesh-Calling-Claim-Epoch` pair that
-`calling_job()` reads.
+request to bind the handler to that job row — so a call a runtime
+originates never carries it, and a nested `task=True` call cannot
+dispatch as, or auto-complete, its caller's job. The registry proxy is
+the one hop that does forward it: there it is still the *inbound*
+header of a push dispatch on its way to the producer that has to bind
+that row. Which job invoked the current handler is a separate question,
+answered by the dedicated `X-Mesh-Calling-Job-Id` /
+`X-Mesh-Calling-Claim-Epoch` pair that `calling_job()` reads.
 
 ## meshctl
 
--- a/src/core/cli/man/content/jobs.md
+++ b/src/core/cli/man/content/jobs.md
@@ -227,7 +227,9 @@ app). On the producer side, the in-process cancel token fires:
 - `asyncio.CancelledError` is raised at the next `await` point in the
   handler.
 - Outbound `McpMeshTool` proxy calls abort their underlying HTTP
-  request (cancel propagates through `X-Mesh-Job-Id` header binding).
+  request. This is the in-process cancel registry firing on the
+  handler's own job context, not something the outbound request
+  carries — the callee is never bound to the caller's job.
 
 The registry treats cancel as terminal (idempotent — already-terminal
 jobs return ok without re-firing).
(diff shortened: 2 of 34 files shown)


dhyansraj:
All six applied in `3ca8ee1ae`. Two were more serious than filed, and I was wrong about the scope of one.

**`options.customHeaders` was a bypass, not a casing nit.** `proxy.ts:557-560` spread it straight into the final headers map, downstream of both the allowlist filter and the dispatch scrub (which only cleaned `mergedHeaders`). User code setting `customHeaders: {"X-Mesh-Job-Id": ...}` put the dispatch discriminator on the wire, and a post-#1570 callee reads it raw and dispatches on it — the exact corruption this PR removes, reachable from application code. Now scrubbed case-insensitively from both maps.

**I told the implementer to leave `mergedHeaders` alone; that was wrong.** My reasoning covered only the in-runtime writers, which all lowercase. It missed two things: `runWithPropagatedHeaders` is exported from `index.ts:168`, so user code can write arbitrary-cased keys into the propagated store; and `mergedHeaders` becomes `_mesh_headers` in the request **body**, whose keys the callee lowercases at `agent.ts:955` into `rawMeshHeaders` — the very map the dispatch gate reads. So an exact-name scrub is defeated by the receiving side re-normalising. Demonstrated by reverting just that scrub: the wire is clean and the body path still carries it.

**`_mesh_headers` leaked into `execute`.** The strip was gated on trace ids or a non-empty `propagatedHeaders`; a request carrying only `x-mesh-job-id` — a genuine dispatch, newly meaningful in TS because of this PR — satisfied neither, so the tool received an internal field in its arguments. Now gated on `rawMeshHeaders` being present.

**The registry recv-cursor guarantee was prose only.** The comment said the header "never travels between agents at all" while the allowlist loop below would forward it under `x-mesh-*` or an exact entry. Now denied before the allowlist is consulted, with both forms tested. Same shape as several other claims this batch has corrected — a guarantee asserted in a comment and unenforced in code.

**The timeout duplicate was worse than a duplicate.** Reading `cur_timeout` from one casing only meant a lowercase inbound value read as `0` — "unset" — so a *tighter* inbound cap got widened rather than merely duplicated.

`upgrading.md` no longer calls a pre-3.8 relay "safe to leave running": that holds for the calls the guard catches, but not for the documented residual, where no warning fires and the callee binds a row it does not own.

The `test_1277` assertion was repointed at the dispatch store and mutation-checked — relaxing the source guard leaves the old assertion passing and fails the new one.

Goldens 1843→1846, all three spans from the one new `upgrading.md` paragraph; the `jobs` rewording is span-neutral, which the delta confirms.

Suites: Python unit + ruff `check`/`format --check` clean; Go 10 packages under `-race`; TypeScript 1461 across 97 files + `typecheck:all`; Java `mvn test` across 7 modules.

**Noted, not fixed**: `llm-provider.ts:632-644` builds `propagatedHeaders` with no strip at all. Out of scope here and it was not established whether that path hands args to user code.

Issue #1556, opened 2026-08-31
python: a non-bool in HealthStatus.checks makes the agent vanish instead of reporting unhealthy


dhyansraj:
Split out of #1551. That issue was closed by #1555, which fixed only its LiteLLM half — the two turned out to be unrelated, so this is the remainder rather than a reopen.

**Reported by a downstream session running on 3.7.0. Not reproduced here yet** — filed so the detail is not lost, and should be reproduced before anything is changed.

## Reported behaviour

`HealthStatus.checks` is typed `dict[str, bool]`. A string value there reportedly causes:

- `fastapi-server-setup` to fail,
- auto-run to exit with *"no FastAPI app prepared"*,
- the pod to serve `/livez` and `/startupz` while **never registering**.

The result is an **absent** agent rather than an unhealthy one: readiness at 503, nothing in the registry, and no explanation on any surface. An unhealthy agent is visible and diagnosable; an absent one looks like it was never deployed.

## Why this shape is worth fixing rather than documenting

The type is reasonable. The failure handling is not: a health check that returns the wrong value *type* is a bad health check, and the runtime already has a verdict for exactly that.

#1539 established the principle — the check throwing, an unusable return type, and TypeScript's deadline all land on an indeterminate verdict that keeps the agent serving, precisely so a buggy probe cannot take down a working agent. A type error inside `checks` is the same class of authoring mistake, and it currently takes down server setup instead.

## Before fixing

Reproduce it. The report suggests returning `{"db": "ok"}` instead of `{"db": True}` from a `health_check`. If it behaves differently than described, that difference is itself worth understanding — the report is from a real deployment on 3.7.0.

Worth establishing while reproducing:

- Where the type actually bites. `HealthStatus` is a typed structure, so is this Pydantic validation, an explicit assertion, or something downstream in `fastapi-server-setup`?
- Whether the same happens for other wrong types in `checks` (int, None, nested dict), or only for `str`.
- Whether `errors` has the equivalent exposure.
- Whether TypeScript and Java fail the same way. #1539 found Java has no dict path at all — it rejects any return type other than `MeshHealth`/boolean at boot — so it may be immune by construction, which would be worth stating.

## Related

The general shape — a small authoring mistake converting into a state that gives the operator nothing to go on — is the same one #1539 and #1555 both addressed on the provider path. Worth checking whether other type or import failures during server setup fail equally silently, rather than fixing this one symptom.

dhyansraj:
[...]

PR #1614, opened 2026-09-08
fix(python): gateway dependency selectors reach the wire; auto_run gates the server, not mesh membership

### LINKED ISSUE #1571: fix(python): @mesh.route and A2A dependency edges lose tags, version, required and match_mode on the wire
## Summary
`mesh/decorators.py:1908-1944` validates and stores the full selector for route dependencies, but `pipeline/api_startup/route_integration.py:501` reduces it to capability names and `pipeline/api_heartbeat/rust_api_heartbeat.py:120-127` registers `DependencySpec(capability, tags="[]", version=None)` with no `required` or `match_mode`. `rust_a2a_heartbeat.py:128-132` does the same for A2A-surface edges. TypeScript (`express.ts:104-128`) and Java (`MeshRouteRegistry.java:693-713`) send all fields.

Related drift in the same file: dependency validation is copy-pasted three times (`decorators.py:1102-1216` tool, `:1883-1949` route, `:2296-2346` a2a) and the a2a copy silently drops `required=True`; the route and a2a copies drop `expected_type`/`match_mode` without a warning.

## Failure scenario
A Python gateway declares `{"capability": "llm", "tags": ["+claude"], "required": True}`. The registry resolves it with no tag preference, `meshctl list`/audit show no required edge, and transitive availability (#1249) never sees the gateway. The local 503 perimeter still works because it reads the decorator, so the gap is invisible until a tag-routed gateway picks the wrong provider.

## Fix
Serialize the full selector on the route and A2A heartbeat paths exactly as the tool path does (`rust_heartbeat.py:243-254`); extract one `validate_dependency_selector` used by all three decorators.

CONFIRMED (source, two readers).

---

### LINKED ISSUE #1589: fix(python): MCP_MESH_AUTO_RUN truthy variants start a serving-but-unregistered agent; auto_run=False is ignored
## Summary
Two lifecycle bugs, both reproduced with the scripts in the audit scratchpad.

- `_mcp_mesh/__init__.py:63-64` accepts `MCP_MESH_AUTO_RUN` in `{"1","yes","on"}` for the immediate uvicorn start, but `mesh/decorators.py:1667-1672` only calls `start_runtime()` when the value is exactly `"true"`. With `MCP_MESH_AUTO_RUN=1` the agent serves `/health` and `/ready` forever, never runs the pipeline, and never registers. Kubernetes sees a healthy pod that is absent from the mesh.
- `@mesh.agent(auto_run=False)` (`decorators.py:1536-1537`) is ignored by `startup_orchestrator.py:743-757`, which checks only the env var, so a blocking uvicorn is still started on the debounce timer thread. The documented use (embed mesh in your own server loop) does not work.

## Fix
One `parse_bool` for the env var shared by both sites; honor the decorator argument in the orchestrator with env taking precedence per the documented hierarchy.

CONFIRMED (reproduced).

---

dhyansraj:
Issues #1571 and #1589. **Two breaking changes** — see release notes below.

## #1571 — route and A2A dependency selectors were dropped before the wire

`@mesh.route` edges were reduced to a bare capability name, with `tags`, `version`, `required`, `match_mode` and consumer schemas hardcoded away. `@mesh.a2a` dropped `required`/`match_mode`/schemas while keeping tags and version. Only `@mesh.tool` sent the full selector. All three now build through one shared builder.

**#1571 says the A2A path "does the same" as the route path. It does not** — A2A carried tags and version. Fixed what was actually missing rather than rewriting working code.

### The builder is strictly index-aligned, and that matters more than it looks

The registry resolves by position (`IndexedResolution.DepIndex`) and the injector applies by position in `metadata["dependencies"]`. So dropping a degenerate entry does not merely lose an edge — **it silently rewires every later dependency onto the wrong slot**. Empty capabilities are reachable (the validator accepts `""`), so entries are never dropped or reordered; a placeholder is emitted and a warning logged. Three alignment tests, including `["first", "", "third"]` holding indices 0/1/2.

Dependency validation was copy-pasted three times, each copy dropping different fields. Now one validator.

## #1589 — `MCP_MESH_AUTO_RUN` parsed four different ways

An exact `== "true"` match, a truthy allowlist, a falsy denylist, and one correct site. With `1`/`yes`/`on` the gates disagreed and the agent served `/ready` as 503 forever, never registering.

**The issue describes this backwards.** It says `__init__.py` is permissive and `decorators.py` strict; the reverse is true. It names two sites; there are four. And its "Kubernetes sees a healthy pod that is absent from the mesh" is wrong in the safer direction — the pod never becomes Ready at all.

### `auto_run` now means what it says

No server, no keep-alive. It does **not** mean "stay out of the mesh": the pipeline runs, the agent registers, and the heartbeat keeps it registered on a daemon thread with its own event loop. The decorator argument is honoured; env still wins.

E2E, against a fake registry:

| scenario | registers | heartbeats | mesh serves | blocks caller |
|---|---|---|---|---|
| `auto_run=False` | yes | 4 | no | no |
| `auto_run=True` + `MCP_MESH_AUTO_RUN=false` | yes | 4 | no | no |
| default | yes | 5 | yes | yes |
| `MCP_MESH_ENABLED=false` | no | 0 | no | no |

That required decoupling the import-time gate from auto-run, since `start_runtime()` is the prerequisite for the pipeline that registers. `MCP_MESH_ENABLED` is now the only inert switch, and it **fails closed**: unset enables, truthy enables, anything else — including empty, a routine empty-Helm-value outcome — disables. The old strict comparison had that property; routing it through `get_config_value`'s soft-fail-to-default would have made a typo *enable* the master off switch.

### This repo's own test suite was relying on the old behaviour

`conftest.py` set `MCP_MESH_AUTO_RUN=false` to keep mesh quiet. Under the new semantics that armed a real pipeline and heartbeat during collection — reproduced before fixing:

```
ORCHESTRATOR_SET=True   PENDING_TIMER=True
Background heartbeat thread started for MCP agent 'contamination-probe-agent-...'
mcp_mesh_core: Starting agent with Rust core
NEW_THREADS=['Thread-4 (run_heartbeat)']
mcp_mesh_core::registry: error sending request for url (http://localhost:8000/agents/...)
```

**From a green run** — `_execute_processing` runs on a `threading.Timer`, so its failures never reach pytest. conftest now uses `MCP_MESH_ENABLED=false`, with inertness asserted positively rather than inferred from a passing suite. `PYTEST_RUNNING` was deleted: nothing read it, so it implied a guard that did not exist.

### Embedded startup reports through `mesh.startup_status()`

The first attempt raised on pipeline failure, with a comment claiming it mirrored the auto-run branch. It did not — that branch deliberately calls `abort_agent_process()` because *"a raise would only kill the timer thread and leave the immediate uvicorn serving probes for an agent that never registered (issue #1556)"*. In embedded mode the caller's server is by definition still running, so a raise reproduces #1556 exactly; and `abort_agent_process()` is also wrong, because mesh does not own the process. Failure is now recorded in a pollable status record, tested through the real Timer path.

## Release notes

> **Breaking: `MCP_MESH_AUTO_RUN=false` and `@mesh.agent(auto_run=False)` no longer keep an agent out of the mesh (Python).** `auto_run` gates the HTTP server and process lifetime only. An agent with auto-run disabled now runs its pipeline, registers and heartbeats; mesh starts no server and does not block. **If you used `MCP_MESH_AUTO_RUN=false` to keep mesh quiet in CI, switch to `MCP_MESH_ENABLED=false`** — otherwise those processes will now register into whatever `MCP_MESH_REGISTRY_URL` points at. `MCP_MESH_ENABLED` fails closed: unset or truthy enables, anything else disables.

> **`@mesh.route` and `@mesh.a2a` dependency selectors now reach the registry (Python).** A deployed gateway declaring `tags: ["+claude"]` on a route dependency has been resolving against *any* provider of that capability and will now resolve only against matching ones — intended, but a routing change on upgrade. The shared validator is also stricter: a non-boolean `required`, or a `match_mode` outside `{"subset","strict"}`, now fails at decoration on route/a2a instead of being ignored.

## Known limits, documented not hidden

Embedded MCP mode has no public accessor for the pipeline-built FastAPI app and installs no SIGTERM handler, so removal is by registry timeout rather than prompt deregistration. Practical today for `@mesh.route`/`@mesh.a2a`, where you own the app. Exposing a handle and a shutdown path needs a design decision and is not in this PR.

## Tests

Full Python suite green (deterministic + three randomized orders, 1 pre-existing skip); `ruff check` and `ruff format --check` clean across 270 files; `go build ./...` clean; man goldens 1852 → 1870 inline and 450 → 452 markup list lines, each annotated with its cause.

Closes #1571
Closes #1589

## Summary by CodeRabbit

- **New Features**
  - Added `MCP_MESH_ENABLED` as a runtime-wide switch to disable mesh activity.
  - Added embedded-mode startup status reporting through `mesh.startup_status()`.
  - `auto_run=False` now supports non-blocking operation while preserving registration and heartbeat behavior.
  - Dependency selectors now retain tags, versions, required flags, match modes, and expected schemas across supported integrations.

- **Documentation**
  - Clarified the differences between `MCP_MESH_ENABLED`, `MCP_MESH_AUTO_RUN`, and embedded mode.
  - Updated testing guidance and examples for the new runtime controls.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -187,6 +187,24 @@ export MCP_MESH_HEALTH_INTERVAL=30
 export MCP_MESH_ENABLED=true
 ```
 
+### MCP_MESH_ENABLED — the off switch
+
+`MCP_MESH_ENABLED=false` makes the runtime inert: no startup pipeline, no
+registration, no heartbeat, no dependency injection. Decorators still import and
+still record their metadata, so your module loads normally — mesh simply never
+runs. This is the flag to use in unit tests and in any process that imports an
+agent module without wanting to join a mesh.
+
+It fails **closed**. Unset means enabled; `true`, `1`, `yes` and `on` enable it;
+anything else — including `false`, `0`, `off`, an empty value and a typo —
+disables it and logs why. An empty value is a routine outcome of an unset Helm
+key, and for the one flag whose job is to turn mesh off, "I could not parse
+this" must not mean "on".
+
+Do not use `MCP_MESH_AUTO_RUN=false` for this. It gates the HTTP server and the
+process lifetime only: an agent with auto-run disabled still registers and still
+heartbeats, by design. See `meshctl man decorators` for the full split.
+
 ## Registry Server Configuration
 
 > These variables configure the **Go registry server** (`mcp-mesh-registry`)
@@ -935,7 +953,7 @@ MCP_MESH_HTTP_HOST=api-service.company.com
 # .env.testing
 MCP_MESH_LOG_LEVEL=WARNING
 MCP_MESH_DEBUG_MODE=false
-MCP_MESH_AUTO_RUN=false
+MCP_MESH_ENABLED=false          # Inert: no pipeline, no registration, no heartbeat
 MCP_MESH_REGISTRY_URL=http://test-registry:8000
 MCP_MESH_NAMESPACE=testing
 ```
@@ -1013,7 +1031,7 @@ class MyAgent:
 # Override decorator settings
 export MCP_MESH_AGENT_NAME=overridden-service
 export MCP_MESH_HTTP_PORT=9090
-export MCP_MESH_AUTO_RUN=false
+export MCP_MESH_AUTO_RUN=false   # No server, no blocking — still registers
 export MCP_MESH_NAMESPACE=custom
 
 # Runs with overridden values
@@ -1171,7 +1189,7 @@ python my_agent.py
 
 ```bash
 # Test environment variables
-export MCP_MESH_AUTO_RUN=false          # Don't auto-start in tests
+export MCP_MESH_AUTO_RUN=false          # No server, no blocking (still registers)
 export MCP_MESH_LOG_LEVEL=ERROR         # Minimal logging
 export MCP_MESH_REGISTRY_URL=http://test-registry:8000
 export MCP_MESH_NAMESPACE=ci-${BUILD_ID}
--- a/docs/python/decorators.md
+++ b/docs/python/decorators.md
@@ -57,6 +57,58 @@ class MyAgent:
     pass
 ```
 
+### auto_run: server and process lifetime, not mesh membership
+
+`auto_run` controls who starts the HTTP server and who owns the process. It does
+not control whether the agent joins the mesh.
+
+| | `auto_run=True` (default) | `auto_run=False` |
+|---|---|---|
+| Startup pipeline runs | yes | yes |
+| Registers with the registry | yes | yes |
+| Heartbeats, dependencies resolve | yes | yes |
+| Starts the HTTP server | yes | no — you do |
+| Keeps the process alive | yes | no — your loop does |
+
+Use `auto_run=False` to embed a mesh agent inside a server loop you already own.
+Serve on the configured `http_port`: that is the address the agent registers, so
+a mismatch registers an endpoint nothing answers on.
+
+`MCP_MESH_AUTO_RUN` overrides the decorator argument, so
+`MCP_MESH_AUTO_RUN=false` behaves exactly like `auto_run=False` — including
+still registering. To make mesh inert instead, set `MCP_MESH_ENABLED=false`.
+
+Startup runs on a background timer, after your decorators import and after
+`@mesh.agent` returns, so mesh can neither raise into your code nor exit the
+process it does not own. Poll the outcome:
+
+```python
+import mesh
+
+status = mesh.startup_status()
+# {"state": "pending"|"ready"|"failed", "pipeline_type": ..., "error": ...,
+#  "heartbeat": bool, "warnings": [...]}
+if status["state"] == "failed":
+    raise SystemExit(f"mesh failed to start: {status['error']}")
+```
+
+Check `heartbeat` and `warnings` even on `ready`: a run can succeed and still
+leave the agent unable to stay registered (standalone mode, or heartbeat setup
+failing), and those are reported rather than announced as success.
+
+**Current limits of embedded mode.** Two things auto-run does for you have no
+public equivalent yet:
+
+- **No app accessor.** The MCP pipeline builds a FastAPI app with FastMCP
+  mounted, but it is only reachable inside the pipeline context — there is no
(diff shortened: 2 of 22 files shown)


dhyansraj:
Applied in `9e24a2caf`. Valid finding — and it is the other half of a blocker this PR already addressed once.

The embedded failure contract was half-wired. The failed-**result** path recorded status; a pipeline that **raised** recorded nothing, so `mesh.startup_status()` reported `pending` forever. That is worse than `failed`: the documented remedy — poll and shut your server down if `state == 'failed'` — never fires, so a caller waits indefinitely instead of reacting. Same #1556 shape, different route in.

Pre-fix failure, driven through the real `threading.Timer` path:

```
AssertionError: the pipeline raised on the timer thread and left no trace the
caller can see — startup_status() is still 'pending' forever
assert 'pending' == 'failed'
```

Two things beyond the suggestion, both of which would have quietly weakened the fix:

**Double-recording.** The failed-result path already calls `set_failed` and *then* raises, and `set_failed` is unconditional last-write-wins. A plain try/except would catch that raise and overwrite the useful `"boom"` with `"RuntimeError: MCP pipeline failed: boom"`. A `status_recorded` flag prevents it, and a test pins that the detail survives its own re-raise.

**Error detail keeps the exception class.** A bare `str(e)` on a `KeyError` produces `'registry_url'` — which is exactly what the pre-fix outer handler logs, and it reads as a stray string rather than a missing key. Detail is now `f"{type(e).__name__}: {e}"`.

On the `Unsupported pipeline type` raise inside that region: **recorded rather than carved out.** It is a mesh bug, not a runtime failure, but the caller's situation is identical — nothing registered, nothing ever will be — and a poll that only terminates for anticipated failures is one the caller still has to time out, which defeats the contract. The class-qualified detail keeps it legible as a bug.

The diff reads as 163/117 but `git diff -w` is 50/4 — nearly all of it is reindenting the existing block into the `try`.

Full Python suite green; `ruff check` and `ruff format --check` clean across 270 files.

**Noted, not fixed** (out of scope): the shared outer handler still logs `f"...: {e}"` without the exception class, and the auto-run path has no status surface at all — `abort_agent_process` is correct there, but that terse log is its only crash diagnostic.

